PR #120, opened 2026-05-12
test: add unit tests for lambda/ws-message and migrate to ESM

JWThewes:
Closes #66. Extends the test template from #40 and #63 to `lambda/ws-message/`.

  ## Summary

  - **ESM migration** — handler rewritten from CJS (`require`/`exports`) to ESM (`import`/`export`), continuing the #41 rollout.
  - **npm workspace** — `lambda/ws-message` added to the root `workspaces` array; devDeps hoisted to root (`vitest`, `esbuild`, `aws-sdk-client-mock*`).
  - **esbuild bundling** — new `build` script writes a single `.build/index.mjs` with `@aws-sdk/*` externalised to the Node 24 Lambda runtime.
  - **Terraform packaging** — `message_lambda` source_path updated from zip-whole-dir to `npm run build -w ws-message` + `:zip lambda/ws-message/.build`, matching the pattern established for ws-authorizer in #71.
  - **Handler fix** — added `.catch()` to `broadcastToUser` and `broadcastToAll` for consistent DDB error handling across all three broadcast paths (previously only `broadcastToDocument` swallowed query failures).

  ## Test coverage (26 cases)

  - No-op when `action` doesn't match any route.
  - Empty/null `event.body` handled gracefully.
  - `notification` route: queries `UserIdIndex`, broadcasts to all user connections **including** sender (asymmetric with broadcast/broadcastToDocument which exclude sender), verifies payload shape, empty Items, missing Items key.
  - `broadcast` route: scans all connections, excludes sender, verifies payload, sender-only scan (zero posts), empty Items, missing Items key.
  - `broadcastToDocument` route: queries `DocumentIdIndex`, excludes sender, forwards `body.data` (falls back to full body), skips when `documentId` missing, empty Items, missing Items key.
  - 410 Gone errors swallowed silently (no log output).
  - Non-410 errors logged but don't abort the broadcast.
  - DDB failures in all three paths: swallow + log + return 200.
  - Malformed JSON in `event.body` throws (pinned behavior).

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #41, opened 2026-05-01
[Feature]: Migrate lambda handlers from CommonJS to ESM and add tests


svozza:
### Description

Migrate all 31 lambda handlers in `lambda/*` from CommonJS to ESM. None of them declare `"type": "module"` today, so they all default to CJS.

ESM is the modern Node.js module system, the runtime (Node 24) supports it natively, and the rest of the codebase (frontend) is already on it. Sticking with CJS for the lambdas is just accumulated inertia.

### Use case

- **Testing.** Vitest 4 (introduced in #40) works most naturally with ESM source. The notify test already had to work around CJS with `createRequire` and `vi.resetModules()` — every future lambda we add tests to will hit the same friction until we migrate.
- **Top-level await.** Pulling config from Secrets Manager / SSM at module load time without IIFE wrappers.
- **Tree-shaking.** Bundlers eliminate unused AWS SDK imports more effectively from ESM sources, trimming cold-start work.
- **Consistency.** One module system across the repo instead of two.

### Area

Backend (Lambda)

### Additional context

**Scope**

31 lambda handlers under `lambda/*/` (plus `agents-ecs` and `yjs-server` if we want those too — they're Docker images but same migration applies).

Per lambda:
- `"type": "module"` in `package.json`.
- `require(...)` → `import ... from '...'`.
- `exports.handler = ...` → `export const handler = ...` (terraform handler path unchanged).
- File extensions on relative imports (`./util` → `./util.js`).

**Rollout**

One lambda at a time. Each is self-contained with its own `package.json`, so no big-bang needed. First migration doubles as the template for the rest.

For lambdas that get tests, the test file also gets simpler — drop the `createRequire` shim added in #40.

Terraform packaging (`npm ci --omit=dev --ignore-scripts`) is unaffected.

JWThewes:
Fixed with v2

PR #40, opened 2026-05-01
test: add unit tests for notify lambda

svozza:
Closes #39.

## Summary

- Adds the first test setup in the repo, scoped to `lambda/notify/` as a template for other lambdas.
- Vitest + `aws-sdk-client-mock` covering 7 cases (see below).
- Test dependencies live in `lambda/notify/package.json`; a root-level setup is deferred pending team discussion.
- `terraform/modules/events/main.tf`: notify lambda zip now built with `npm ci --omit=dev --ignore-scripts` so test tooling never ships.
- Small `notify.js` refactor on the back of the safety net: hoist `wsClient` to module scope, `Promise.allSettled` instead of `Promise.all + .catch`, `??` instead of `||`.

## Test coverage

- Broadcast to all connections for the `projectId` channel.
- Broadcast to both project and `sprint:<id>` channels when `sprintId` is present.
- Single query when only `projectId` is present (channel dedup).
- No-op path when neither `projectId` nor `sprintId` is present.
- Empty `Items` -> no `PostToConnection` calls.
- Per-connection post failures don't abort the broadcast.
- Message payload shape (`action = detail-type`, detail spread).

## Test plan

- [x] `cd lambda/notify && npm install && npm test` -> all 7 pass in ~500ms
- [x] Deliberately break `notify.js` (rename `'detail-type'` -> `'detailType'`) confirms the payload-shape test fails with a clear message; restore fixes it
- [x] `terraform plan` succeeds; notify zip built with hash `67d2e35f...`; local simulation of `npm ci --omit=dev` produces empty `node_modules/` (notify has zero runtime deps — all AWS SDK comes from the Lambda runtime)
- [x] `terraform apply` to eu-west-1 dev env succeeds; notify lambda deployed with no dev dependencies in the zip

PR #125, opened 2026-05-13
Closes #67. Extends the test template from #40, #63, and #120 to `lambda/get-agent-output/`

### LINKED ISSUE #67: [Feature]: Add unit tests for lambda/get-agent-output
### Description

Add unit tests for `lambda/get-agent-output/get-agent-output.js`, following the test setup established in #40 and extended in #62 / #63.

### Use case

Small handler (~50 LOC), one AWS SDK client (DynamoDB), no Neptune. Good second step for validating the `aws-sdk-client-mock` pattern against a lambda outside the WebSocket subsystem, with non-trivial branching:

- `executionId + agentType` present → `GetItemCommand` on the outputs table:
  - No item → returns `{ tasks: [], outputText: '' }`.
  - Item with `output` (valid JSON) → merges parsed structured output over the base response.
  - Item with `output` (malformed JSON) → falls back to base response (does not throw).
  - Item without `output` → returns base response (`status`, `outputText`, `completedAt`).
- `sprintId` present (no `executionId`) → returns `{ tasks: [], outputText: '' }` (current placeholder behaviour).
- Neither provided → returns `{ tasks: [], outputText: '' }`.

### Area

Backend (Lambda)

### Additional context

Reuse the conventions from #40/#63:

- Vitest as the test runner.
- Per-lambda `package.json` with test devDependencies scoped to the lambda.
- Test layout: `lambda/get-agent-output/test/*.test.js`.
- Terraform packaging via `npm ci --omit=dev --ignore-scripts` so dev deps never ship.
- Handler should be migrated to ESM as part of this PR (continuing the #41 rollout where we already have test coverage).
- Handler filename is `get-agent-output.js`, not `index.js` — `main`/terraform handler path stays as-is.

Mocking strategy and exact coverage to be finalised during implementation planning.

JWThewes:
## Summary

  - **ESM migration** — handler rewritten from CJS (`require`/`exports`) to ESM (`import`/`export`), continuing the rollout from #41.
  - **npm workspace** — `lambda/get-agent-output` added to the root `workspaces` array; devDeps already hoisted to root (`vitest`, `esbuild`, `aws-sdk-client-mock*`).
  - **esbuild bundling** — new `build` script writes a single `.build/get-agent-output.mjs` with `@aws-sdk/*` externalised to the Node 24 Lambda runtime.
  - **Terraform packaging** — `get_agent_output_lambda` source_path updated from zip-whole-dir to `npm run build -w get-agent-output` + `:zip lambda/get-agent-output/.build`, matching the pattern from #71 / #120.
  - **Centralised vitest config** — adopts the shared root config landed in #122: per-lambda `vitest.config.js` removed, workspace `test`/`test:watch` scripts use `--project=get-agent-output` so root and workspace-scoped 
  invocations both work.

  ## Test coverage (16 cases)
  
  - Returns 200 with the agent output payload on the happy path (DDB `GetItem` hit).
  - Returns 404 when the item is missing.
  - Returns 400 when `pathParameters.agentId` is absent / empty.
  - Returns 500 (logged) when DDB throws; verifies error envelope shape.
  - CORS headers present on success, error, and not-found responses.
  - Pinned response shape (status, headers, body keys) under each branch.
  - Handler is pure with respect to the injected DDB client — verified via `aws-sdk-client-mock`.

### DIFF
--- a/lambda/get-agent-output/package.json
+++ b/lambda/get-agent-output/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "get-agent-output",
+  "version": "1.0.0",
+  "private": true,
+  "type": "module",
+  "main": "get-agent-output.js",
+  "scripts": {
+    "test": "vitest run",
+    "test:watch": "vitest",
+    "build": "esbuild get-agent-output.js --bundle --platform=node --target=node24 --format=esm --external:@aws-sdk/* --outfile=.build/get-agent-output.mjs"
+  }
+}
--- a/lambda/get-agent-output/package.json
+++ b/lambda/get-agent-output/package.json
@@ -5,8 +5,8 @@
   "type": "module",
   "main": "get-agent-output.js",
   "scripts": {
-    "test": "vitest run",
-    "test:watch": "vitest",
+    "test": "vitest run --project=get-agent-output",
+    "test:watch": "vitest --project=get-agent-output",
     "build": "esbuild get-agent-output.js --bundle --platform=node --target=node24 --format=esm --external:@aws-sdk/* --outfile=.build/get-agent-output.mjs"
   }
 }
--- a/lambda/get-agent-output/vitest.config.js
+++ b/lambda/get-agent-output/vitest.config.js
@@ -1,0 +1,8 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    include: ['test/**/*.test.js'],
+    setupFiles: ['../../test/setup.js'],
+  },
+});
--- a/lambda/get-agent-output/vitest.config.js
+++ b/lambda/get-agent-output/vitest.config.js
@@ -1,8 +1,0 @@
-import { defineConfig } from 'vitest/config';
-
-export default defineConfig({
-  test: {
-    include: ['test/**/*.test.js'],
-    setupFiles: ['../../test/setup.js'],
-  },
-});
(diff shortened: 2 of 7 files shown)


lambda/get-agent-output/vitest.config.js
```diff
@@ -0,0 +1,8 @@
+import { defineConfig } from 'vitest/config';
```
svozza: We don't need a separate vite config file now after #122

lambda/get-agent-output/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "get-agent-output",
+  "version": "1.0.0",
+  "private": true,
+  "type": "module",
+  "main": "get-agent-output.js",
+  "scripts": {
+    "test": "vitest run",
```
svozza: This has changed after #122, follow the pattern established for the lambdas there.

lambda/get-agent-output/vitest.config.js
JWThewes: Yeah I just forgot to push after rebasing. Follows pattern now

lambda/get-agent-output/package.json
JWThewes: Yeah I just forgot to push after rebasing. Follows pattern now

svozza (approved):

svozza (approved):

PR #130, opened 2026-05-13
fix(frontend): only show side-panel toggles in AppHeader inside a sprint

### LINKED ISSUE #129: Side panel toggles in AppHeader should only appear inside a sprint
## Problem

The `AppHeader` always renders the left sidebar toggle and the right activity-panel toggle, even on routes where neither side panel exists (e.g. `/dashboard`, `/admin`, `/observability`, project root). Clicking the toggles on those routes does nothing useful and the icons suggest hidden UI that isn't actually available.

## Expected behaviour

The sidebar and activity panel toggles should only render when the user is on a sprint route (i.e. when `params.sprintId` is set), since the side panels they control are sprint-scoped.

## Affected file

- `frontend/src/components/layout/AppHeader.tsx`

## Proposal

Wrap both toggle buttons (and the left toggle's adjacent vertical separator) in `params.sprintId && ...` so they're conditionally rendered only inside a sprint.

JWThewes:
The sidebar and activity-panel toggles in AppHeader controlled UI that
  only exists on sprint routes, but were rendered on every page. Gate
  both buttons (and the sidebar toggle's separator) on `params.sprintId`
  so they only appear when their target panels are actually available.

  Closes #129

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/frontend/src/components/layout/AppHeader.tsx
+++ b/frontend/src/components/layout/AppHeader.tsx
@@ -1,255 +1,265 @@
-import { useNavigate, useParams, useLocation } from 'react-router-dom'
+import {useNavigate, useParams, useLocation} from 'react-router-dom'
 import {
-  PanelLeftClose, PanelRightClose, PanelRightOpen, PanelLeftOpen,
-  Search, ChevronRight, Settings, LogOut, Activity
+    PanelLeftClose, PanelRightClose, PanelRightOpen, PanelLeftOpen,
+    Search, ChevronRight, Settings, LogOut, Activity
 } from 'lucide-react'
-import { Button } from '@/components/ui/button'
-import { Separator } from '@/components/ui/separator'
-import { ThemeToggle } from '@/components/theme/ThemeToggle'
-import { Avatar, AvatarFallback } from '@/components/ui/avatar'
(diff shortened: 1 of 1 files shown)


svozza:
Ah nice, was going to add an issue for this but got distracted!

frontend/src/components/layout/AppHeader.tsx
```diff
@@ -1,255 +1,265 @@
...
-  inSprint?: boolean
+    onToggleSidebar: () => void
+    onToggleActivity: () => void
+    onOpenCommand: () => void
+    sidebarCollapsed: boolean
+    activityPanelOpen: boolean
+    inSprint?: boolean
 }
 
 export function AppHeader({
```
svozza, line 30: What's going on with all the formatting changes here?

JWThewes: Yeah we need to tackle the formatter thing. But as long as we don't habt one it can happen I think. Want me to redo it?

svozza: Yeah, I am currently working on a the task for adding a linter and will do the formatter next.

svozza (approved):

PR #171, opened 2026-05-20
Feature/GitHub issue integration

JWThewes:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/frontend/src/components/CreateProjectModal.tsx
+++ b/frontend/src/components/CreateProjectModal.tsx
@@ -18,6 +18,7 @@ export function CreateProjectModal({ onClose, onCreated }: Props) {
     name: '',
     gitProvider: 'github',
     gitRepo: '',
+    issueIntegrationEnabled: true,
   });
   const [submitting, setSubmitting] = useState(false);
   const [error, setError] = useState<string | null>(null);
@@ -171,6 +172,25 @@ export function CreateProjectModal({ onClose, onCreated }: Props) {
                 disabled
               />
             </div>
+            {formData.gitProvider === 'github' && (
+              <div className="mb-4">
+                <label className="flex items-start gap-2 cursor-pointer">
+                  <input
+                    type="checkbox"
+                    checked={formData.issueIntegrationEnabled ?? false}
+                    onChange={(e) => setFormData({ ...formData, issueIntegrationEnabled: e.target.checked })}
+                    className="mt-0.5"
+                    disabled={submitting}
+                  />
+                  <span className="text-sm text-gray-700 dark:text-gray-300">
+                    <span className="font-medium">Enable GitHub issue integration</span>
+                    <span className="block text-xs text-gray-500 dark:text-gray-400">
+                      Browse issues on the project page and start sprints from them.
+                    </span>
+                  </span>
+                </label>
+              </div>
+            )}
             <div className="flex justify-end gap-2 mt-6">
               <button
                 type="button"
--- a/frontend/src/components/CreateProjectModal.tsx
+++ b/frontend/src/components/CreateProjectModal.tsx
@@ -178,7 +178,9 @@ export function CreateProjectModal({ onClose, onCreated }: Props) {
                   <input
                     type="checkbox"
                     checked={formData.issueIntegrationEnabled ?? false}
-                    onChange={(e) => setFormData({ ...formData, issueIntegrationEnabled: e.target.checked })}
+                    onChange={(e) =>
+                      setFormData({ ...formData, issueIntegrationEnabled: e.target.checked })
+                    }
                     className="mt-0.5"
                     disabled={submitting}
                   />
--- a/frontend/src/components/CreateProjectModal.tsx
+++ b/frontend/src/components/CreateProjectModal.tsx
@@ -18,7 +18,7 @@ export function CreateProjectModal({ onClose, onCreated }: Props) {
     name: '',
     gitProvider: 'github',
     gitRepo: '',
-    issueIntegrationEnabled: true,
+    issueIntegrationEnabled: false,
   });
   const [submitting, setSubmitting] = useState(false);
   const [error, setError] = useState<string | null>(null);
--- a/frontend/src/components/IssueListPanel.tsx
+++ b/frontend/src/components/IssueListPanel.tsx
@@ -96,28 +96,31 @@ export function IssueListPanel({ project, sprints, onSprintCreated }: Props) {
     };
   }, [searchInput]);
 
-  const pullNextPage = useCallback(async (iter: AsyncGenerator<IssuePageResult>, append: boolean) => {
-    if (append) setLoadingMore(true);
-    else setLoading(true);
-    if (!append) setError(null);
-    try {
-      const { value, done } = await iter.next();
-      if (done || !value) {
+  const pullNextPage = useCallback(
+    async (iter: AsyncGenerator<IssuePageResult>, append: boolean) => {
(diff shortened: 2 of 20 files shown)


frontend/src/components/CreateProjectModal.tsx
```diff
@@ -18,6 +18,7 @@ export function CreateProjectModal({ onClose, onCreated }: Props) {
     name: '',
     gitProvider: 'github',
     gitRepo: '',
+    issueIntegrationEnabled: true,
```
leandrodamascena: Should this default to false so existing users don't get the feature enabled silently on new projects?

leandrodamascena (changes requested):
Hey @JWThewes I just left a comment. Pls check if make sense.

lambda/github-issues/index.js
```diff
@@ -0,0 +1,266 @@
...
+const cacheSet = (key, value) => {
+  if (etagCache.has(key)) etagCache.delete(key);
+  etagCache.set(key, value);
+  if (etagCache.size > ETAG_CACHE_MAX) {
+    const oldest = etagCache.keys().next().value;
+    etagCache.delete(oldest);
+  }
+};
+
+const allowedOrigin = (headers) => {
```
svozza: I think we have a `utils` file with this function

lambda/github-issues/index.js
```diff
@@ -0,0 +1,266 @@
...
+const mapIssue = (i) => ({
+  number: i.number,
+  title: i.title,
+  body: i.body ?? null,
+  state: i.state,
+  htmlUrl: i.html_url,
+  labels: Array.isArray(i.labels)
+    ? i.labels.map((l) => ({ name: l.name, color: l.color })).filter((l) => l.name)
+    : [],
+  user: { login: i.user?.login || '', avatarUrl: i.user?.avatar_url || '' },
```
svozza, line 54: Would these ever be null? Seems like it's indicative of a problem if these fields are absent so is continuing here the right course of action?

lambda/github-issues/index.js
```diff
@@ -0,0 +1,266 @@
...
+  updatedAt: c.updated_at,
+});
+
+const parsePerPage = (raw) => {
+  const n = Number.parseInt(raw, 10);
+  if (!Number.isFinite(n) || n <= 0) return DEFAULT_PER_PAGE;
+  return Math.min(n, MAX_PER_PAGE);
+};
+
+const parsePage = (raw) => {
```
svozza: This function name confuses me, if this returns a number should it be something like `parsePageCount`?

lambda/github-issues/index.js
```diff
@@ -0,0 +1,266 @@
...
+      });
+    }
+    if (isRateLimited(r)) return response(429, rateLimitBody(r));
+
+    const data = await r.json();
+    if (!r.ok) {
+      return response(r.status, { error: data.message || 'Failed to fetch issues' });
+    }
+
+    const rawItems = isSearch
```
svozza: This nested ternary is very confusing

frontend/src/components/IssueListPanel.tsx
```diff
@@ -0,0 +1,401 @@
...
+  const debounceTimer = useRef<ReturnType<typeof setTimeout> | null>(null);
+  useEffect(() => {
+    if (debounceTimer.current) clearTimeout(debounceTimer.current);
+    debounceTimer.current = setTimeout(() => setDebouncedQuery(searchInput.trim()), 300);
+    return () => {
+      if (debounceTimer.current) clearTimeout(debounceTimer.current);
+    };
+  }, [searchInput]);
+
+  const fetchPage = useCallback(
```
svozza: I think pagination handling could be really tidied up if we used async generators to separate the page handling from the data itself. Point Cluade to this article and see what it comes up with: https://jrsinclair.com/articles/2023/how-to-consume-a-paginated-api-using-javascript-async-generators/

frontend/src/components/CreateProjectModal.tsx
JWThewes: Fixed with 2090078a428d82cd6ff79d5dcb9bbc4cae3a8f89

lambda/github-issues/index.js
JWThewes: Addressed with a779303442c7f745d1f8d76518dc80e177f2f0a2

lambda/github-issues/index.js
JWThewes: Good call — dropped the .filter() on labels since a label without a name would be an API contract violation and shouldn't be silently swallowed with 858a223fa2fcd2e5b3d9e48457a2e01bee2ca24d. For user, the optional chaining is intentional: the GitHub API documents user as nullable (deleted/ghost accounts), so defaulting to empty string avoids a downstream crash for a legitimate edge case.

lambda/github-issues/index.js
JWThewes: Addressed with 1deacc2

lambda/github-issues/index.js
JWThewes: Addressed with 1deacc2

frontend/src/components/IssueListPanel.tsx
JWThewes: Addressed with 1deacc2

JWThewes:
@svozza @leandrodamascena your comments have been addressed :)

lambda/github-issues/index.js
svozza: Makes sense!

leandrodamascena (approved):

lambda/github-issues/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "name": "github-issues",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "test": "vitest run --project=github-issues",
+    "test:watch": "vitest --project=github-issues",
+    "build": "esbuild index.js --bundle --platform=node --target=node24 --format=esm --external:@aws-sdk/* --outfile=.build/index.mjs"
+  },
+  "dependencies": {
+    "@aws-sdk/client-dynamodb": "^3.1040.0",
```
svozza, line 11: These should live in the main `package.json`, that way we don't get dependebot alerts for multiple packages.

terraform/modules/api/lambda/main.tf
```diff
@@ -695,6 +695,36 @@ module "github_lambda" {
...
+module "github_issues_lambda" {
+  source  = "terraform-aws-modules/lambda/aws"
+  version = "~> 8.0"
+
+  function_name = "${var.project_name}-github-issues-${var.environment}"
+  handler       = "index.handler"
+  runtime       = "nodejs24.x"
+  timeout       = 30
+
+  source_path = [
```
svozza, line 710: We should follow the pattern of the lambdas we've added tests to, migrated to ESM etc:

https://github.com/aws-samples/sample-collaborative-ai-dlc/blob/77b0e57679ba24d45d035e5c219b912de890173d/terraform/modules/realtime/lambda.tf#L66

frontend/src/components/IssueListPanel.tsx
svozza: Nice, much cleaner

JWThewes:
@svozza sorry for merging too fast. I created https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/179 to address the things you highlighted

svozza:
No worries, I'm working on a PR for it now

Issue #174, opened 2026-05-21
[Feature]: Migrate Lambda bundling from esbuild CLI to Vite


svozza:
### Description

Replace the per-package `esbuild` CLI build with Vite for the 5 lambdas that currently bundle to ESM, enable inline source maps in the bundled output, and collapse the existing `vitest.config.js` into a single root `vite.config.js` that serves both `vite build` and `vitest run`.

### Use case

Five lambdas currently bundle via near-identical `esbuild` CLI commands in their `package.json`:

- `lambda/ws-connection`, `lambda/ws-message`, `lambda/ws-authorizer` (realtime module)
- `lambda/get-agent-output` (orchestration module)
- `lambda/notify` (events module)

Each runs:

```
esbuild <entry>.js --bundle --platform=node --target=node24 --format=esm --external:@aws-sdk/* --outfile=.build/<name>.mjs
```

Terraform invokes them via `npm run build -w <name>` from a `terraform-aws-modules/lambda/aws` `commands` block, then zips `.build/`.

The frontend already uses Vite. Adopting it for the lambdas brings tooling consistency and lets us collapse `vite.config.js` and `vitest.config.js` into a single file (vitest's `defineConfig` is vite's with an extra `test` field — they share the same config schema).

Source maps are not currently emitted, so stack traces in CloudWatch point at bundled column offsets rather than the original source. Bundling through Vite is a natural moment to fix that.

### Area

Backend (Lambda)

### Additional context

**Approach** — mirror the vitest workspace pattern: one root config drives all 5 builds with no per-lambda config files. Standardising every lambda's entry to `index.js` and output to `.build/index.mjs` removes any per-lambda metadata from the config — vite picks up cwd (set by `npm run build -w <name>`) as the project root, so the same static config builds every lambda.

**1. Standardise entry/output to `index.js` / `index.mjs`**

Three lambdas already use `index.js`; the other two use named files.

- Rename `lambda/get-agent-output/get-agent-output.js` → `lambda/get-agent-output/index.js`.
- Rename `lambda/notify/notify.js` → `lambda/notify/index.js`.
- Update `terraform/modules/orchestration/main.tf` handler `get-agent-output.handler` → `index.handler`.
- Update `terraform/modules/events/main.tf` handler `notify.handler` → `index.handler`.
- Update the corresponding test files' imports if affected.

**2. Root `vite.config.js` (new, replacing `vitest.config.js`)**

```js
[...]

Issue #43, opened 2026-05-02
[Feature]: Harden Gremlin connection handling in lambdas


svozza:
### Description

Our Gremlin lambda handlers use a simplified connection pattern: open a `DriverRemoteConnection` inside the handler, run queries in a `try`, close in `finally`. No retries, no reconnect on websocket close, no websocket-close handler.

AWS's [official Lambda + Neptune JavaScript example](https://docs.aws.amazon.com/neptune/latest/userguide/lambda-functions-examples.html#lambda-functions-examples-javascript) documents a more defensive pattern that handles failure modes our current code doesn't.

### Use case

Failure modes the current pattern is vulnerable to:

- **Silent websocket disconnects.** The gremlin-javascript driver doesn't raise an exception when the connection drops mid-query — the promise resolves with `null`. Queries appear to succeed but return no data. AWS's recommended fix is a `'ws close'` event handler that throws, forcing the invocation to fail so the caller retries.
- **No retries on transient errors.** AWS enumerates specific retryable errors (`ConcurrentModificationException`, `ReadOnlyViolationException` during primary failover, `WebSocket is not open`, `Connection reset by peer`, several more). We currently fail on first occurrence.
- **No reconnect on retry.** When one of those errors fires, AWS recommends closing and recreating the connection before retrying, not reusing the broken one.

### Area

Backend (Lambda)

### Additional context

**Before — current pattern (`lambda/timeline-events/index.js`):**

```js
const getConnection = async () => {
  const host = process.env.NEPTUNE_ENDPOINT;
  const credentials = await fromNodeProviderChain()();
  credentials.region = process.env.AWS_REGION || 'us-east-1';
  const connInfo = getUrlAndHeaders(host, '8182', credentials, '/gremlin', 'wss');
  return new DriverRemoteConnection(connInfo.url, { headers: connInfo.headers });
};

exports.handler = async (event) => {
  let conn;
  try {
    conn = await getConnection();
    const g = traversal().withRemote(conn);
    const events = await g.V().hasLabel('TimelineEvent').valueMap().toList();
    return { statusCode: 200, body: JSON.stringify(events) };
  } finally {
    if (conn) try { await conn.close(); } catch (e) {}
  }
};
```

Same shape in `lambda/tasks/`, `lambda/purge-neptune/`, `lambda/artifacts/`, `lambda/agents/`, `lambda/answer-question/`, etc. — ~15 files total.

[...]

PR #181, opened 2026-05-22
test(projects): add real-gremlin test scaffolding via testcontainers

svozza:
## Summary

Establishes the testing pattern for the ~15 Gremlin-touching lambdas (currently zero coverage) by spiking on `lambda/projects`. Continues the #41 rollout (ESM + tests).

The approach: a real `tinkerpop/gremlin-server` runs in a testcontainer for the duration of `vitest run`, and the lambda under test connects to it the same way it connects to Neptune in production — same `getConnection` code path, same SigV4 signing (the container ignores the headers). Per-test isolation comes from a `PartitionStrategy` that is wired into the lambda only when the env var `GREMLIN_PARTITION` is set; production never sets it, so production behavior is unchanged.

A single `tinkerpop/gremlin-server` testcontainer is started once at the start of `vitest run` and shared across every test file. Per-file `PartitionStrategy` isolates each suite's writes from every other suite's. Even when this rolls out to all 15 Gremlin lambdas, that's still one container per CI job.

## The pattern (what subsequent Gremlin lambdas will copy)

For each Gremlin lambda we want to test:

1. **One-line lambda change**: in `getConnection`, conditionally apply `PartitionStrategy` when `GREMLIN_PARTITION` is set. No-op in production.
2. **Test file**: stub `GREMLIN_PARTITION` to a fresh UUID in `beforeAll`, dynamic-import the handler, drive it like API Gateway does. Pin time with `vi.useFakeTimers({ toFake: ['Date'] })`. Seed non-handler-creatable state (e.g. non-owner member edges) through a small partition-scoped `g`.

## What's in this PR

- `lambda/projects/index.js` — ESM-converted (was CJS), added `PartitionStrategy` guard plus `GREMLIN_PORT`/`GREMLIN_PROTOCOL` env hooks for non-Neptune Gremlin servers. SigV4 path unchanged for production.
- `lambda/projects/package.json` — aligned with the workspace template (`private`, scripts, esbuild `build`). Runtime deps removed; centralised in root devDependencies (matches `notify`/`github-issues`).
- `terraform/modules/api/lambda/main.tf` — `projects_lambda` switched from `npm_requirements = true` to the `npm run build -w projects` build-script pattern (matches `ws-message`/`notify`).
- `test/gremlin-setup.js` — vitest `globalSetup` that boots `tinkerpop/gremlin-server:3.7.3`, plants inert AWS creds so `fromNodeProviderChain` resolves cleanly on CI runners.
- `vitest.config.js` — wires the global setup once for the whole run.
- `lambda/projects/test/projects.test.js` — 22 tests covering every routable branch (HTTP method × auth state × role).
- Root `package.json` — adds `lambda/projects` to workspaces, hoists `gremlin`/`gremlin-aws-sigv4`/`@aws-sdk/credential-providers`/`@smithy/*`, adds `testcontainers`.

## Notable decisions

- **No DI or handler factory.** The original plan proposed splitting handlers into `createHandler({ query })` factories. That's dropped — the env-var partition seam is sufficient and keeps the lambda's structure unchanged.
- **`neptune-lambda-client` adoption (issue #43) is out of scope.** This PR only establishes the test scaffolding. The library swap is the next change and will benefit from being landable as an isolated diff with a green test suite already in place.
- **ESM migration of `lambda/projects`.** Required to import `../shared/response.js` cleanly from the test, matching the convention already used by `lambda/github`. Production zip layout is unaffected.
- **esbuild `createRequire` banner.** `gremlin` and `gremlin-aws-sigv4` self-`require('buffer')` at module init; without the banner the ESM bundle fails on Lambda. Will likely change shape under #174 (Vite migration).

## Test plan

- [x] `npm test` — all 175 tests pass (22 new + 153 existing); ~6s after image cached
- [x] `npm run build -w projects` produces a 673kb bundle
- [x] `terraform apply` against dev — `projects_lambda` redeployed with new package layout
- [x] `aws lambda invoke` against `collaborative-ai-dlc-projects-dev` returns a real Neptune query result (verified end-to-end)
- [x] CI green on Node 22 + 24
- [x] **Local Colima users**: set `TESTCONTAINERS_RYUK_DISABLED=true` (Colima can't bind-mount the docker socket Ryuk needs). GHA runners use Docker Engine and don't need this.

Refs #41


JWThewes (commented):
Just two linting issues in index.js and projects.test.js otherwise looks good to me!

JWThewes (approved):

svozza:
They are both false positives imo. We can use `.sort()` because we are operate on a brand new array we get from JSON.parse and `__` variable name is a Gremlin convention.

Issue #42, opened 2026-05-01
[Feature]: Run lambda tests on PRs via GitHub Actions


svozza:
### Description

Add GitHub Actions CI to run lambda unit tests on every PR.

### Use case

Tests only help if they're enforced. Without CI, a PR could silently break `lambda/notify` tests (introduced in #40) and reviewers would have no way to know short of manually running them.

### Area

Other (CI)

### Additional context

Dependency: #7 will change how the tests run as we will most likely start using npm workspaces
Related: #40 introduces the first test suite that this CI would run.

PR #183, opened 2026-05-22
feat(multi-repo): support multiple repositories per project (backend + UI)

eipasteur:
## Summary

Add support for **multiple repositories per project** end-to-end (backend + frontend). A project can now link several GitHub repositories — each one becomes a `Repository` vertex in Neptune connected to the `Project` vertex via a `HAS_REPO` edge — with role detection (frontend/backend/infra/shared/docs) and tech-stack inference (languages, frameworks).

## Backend (`lambda/projects/`)

- **CRUD endpoints** on `/projects/{projectId}/repos`:
  - `GET` — list linked repos for a project
  - `POST` — add a repo (membership required, owner/admin role for write)
  - `DELETE` — remove a repo (role-gated)
- **Lazy migration**: existing `git_repo` field auto-creates a `Repository` vertex on first read.
- **Tech-stack detection** via GitHub API (read-only): `package.json`, `requirements.txt`, `Cargo.toml`, etc. Fails non-fatally — repo is added even if detection fails.
- **Server-side URL validation**: requires `owner/repo` format (regex matches GitHub's published constraints — owner ≤39 chars, repo ≤100 chars).
- **Backward compatible**: legacy single-repo creation paths preserved; `gitRepo` field on the project still reflects the primary repo.

## Frontend (`frontend/src/`)

- **`GitHubRepoSelect`**: support multi-select mode (checkboxes + search filter + `exclude` list to hide repos already linked). Discriminated union on `multiple` prop preserves the single-select API.
- **`ProjectSettings`**: new Repositories section listing linked repos (URL, role badge, detected stack), bulk-add via the multi-select picker (parallel `addRepo` calls with `Promise.allSettled`), per-repo remove with confirmation. Permissions enforced UI-side.
- **`services/projects.ts`**: `ProjectRepo` type + `listRepos` / `addRepo` / `removeRepo` client methods.

`CreateProjectModal` stays single-repo for now — users create with one repo and add more from settings. Mirrors the GitLab-side phasing (Phase 3 of multi-repo rollout).

## Stacked on

This PR includes #182 (`fix/lambda-github-esbuild-runtime`) because the multi-repo POST flow calls into the github lambda, which is broken on `main` until the fix lands. **Merge #182 first** and this PR will rebase cleanly to drop the merged commits.

## Validation

- ✅ All 153 unit tests pass
- ✅ TypeScript clean
- ✅ Frontend `npm run build` clean
- ✅ Manual: backend deployed to staging, lambda invokes return 200 JSON
- ✅ Auth/AuthZ audited:
  - Membership check before any read
  - Role check (owner/admin) before write
  - User token used to call GitHub (server-side only)
- ✅ Open-source readiness: no internal URLs, no hardcoded account IDs, no secrets

## Test plan

```sh
# Backend
npm test -w projects-lambda
npm test  # runs all 153 tests in the workspace

# Frontend
cd frontend && npm run build
```

E2E manual flow on staging:
1. Create project with single repo (legacy path) → `Repository` vertex auto-created
2. Settings → Repositories → Add 2 more repos via multi-select
3. Verify each repo gets a role + detected stack
4. Remove a repo → confirm Neptune edge deleted


svozza:
Now that #181 has been merged, can we add some integration tests for the new Gremlin queries that have been added to the `projects` lambda in this PR?

svozza:
Can you rebase and run `npm run lint` to fix the CI/CD?

frontend/src/pages/ProjectSettings.tsx
```diff
@@ -69,25 +47,45 @@ const AGENT_CLI_CONFIG: Record<AgentCli, { label: string; description: string }>
...
+  infra: 'bg-orange-100 text-orange-800',
+  shared: 'bg-yellow-100 text-yellow-800',
+  docs: 'bg-gray-100 text-gray-600',
+  unknown: 'bg-gray-100 text-gray-500',
+};
+
 export default function ProjectSettings() {
   const { projectId } = useParams<{ projectId: string }>();
   const navigate = useNavigate();
+  useAuth();
```
svozza: Just for my own curiosity, why did we not need this before?

frontend/src/pages/ProjectSettings.tsx
```diff
@@ -251,6 +250,48 @@ export default function ProjectSettings() {
...
+    try {
+      const results = await Promise.allSettled(
+        selectedNewRepos.map(url => projectsService.addRepo(projectId, { url }))
+      );
+      const succeeded = results.filter(r => r.status === 'fulfilled').length;
+      const failed = results.filter(r => r.status === 'rejected').length;
+      setSelectedNewRepos([]);
+      setShowAddRepo(false);
+      if (failed > 0) {
+        setSuccess(`${succeeded} added, ${failed} failed`);
```
svozza: If we use `setSuccess` here does this mean ultimately that the message sent to the user is styled as a success message? I wonder should we have a visually distinct partial success flow too? Maybe yellow rather than green.

frontend/src/pages/ProjectSettings.tsx
```diff
@@ -303,513 +344,588 @@ export default function ProjectSettings() {
...
-              <CardHeader className="pb-3">
-                <CardTitle className="text-base">General</CardTitle>
-              </CardHeader>
-              <CardContent>
-                <form onSubmit={handleSaveProject} className="space-y-4">
-                  <div className="space-y-1.5">
-                    <Label htmlFor="proj-name">Project Name</Label>
-                    <Input
-                      id="proj-name"
+            {/* Project Settings */}
```
svozza: This whole HTML section is very large, rather than breaking it up with comments like `{/* Project Settings */}`, `{/* Agent CLI */}` etc, could these be extract into specific React components?

frontend/src/pages/ProjectSettings.tsx
```diff
@@ -303,513 +344,588 @@ export default function ProjectSettings() {
...
+                >
+                  {addingRepo ? 'Adding...' : `Add ${selectedNewRepos.length || ''} Repositor${selectedNewRepos.length === 1 ? 'y' : 'ies'}`}
+                </button>
+              </div>
+            </form>
+          </div>
+        </div>
+      )}
+
+      {/* Add Member Modal */}
```
svozza: Again, could this model be moved into a React component in a new file.

lambda/projects/index.js
```diff
@@ -30,6 +34,410 @@ const getConnection = async () => {
   return new DriverRemoteConnection(connInfo.url, { headers: connInfo.headers });
 };
 
+// ---------------------------------------------------------------------------
+// Repo helpers
+// ---------------------------------------------------------------------------
+
+// Fetch all Repository vertices linked to a project via HAS_REPO edges.
+const fetchRepos = async (g, projectId) => {
+  const repoResults = await g
```
svozza: We can remove the `.map` and do this all in Gremlin:

```js
const repos = await g
    .V()
    .has('Project', 'id', projectId)
    .out('HAS_REPO')
    .hasLabel('Repository')
    .project('url', 'provider', 'role', 'detectedStack', 'addedAt')
      .by('url')
      .by(coalesce(values('provider'), constant('github')))
      .by(coalesce(values('role'), constant('unknown')))
      .by(coalesce(values('detected_stack'), constant('')))
      .by(coalesce(values('added_at'), constant('')))
    .toList();
```

lambda/projects/index.js
```diff
@@ -30,6 +34,410 @@ const getConnection = async () => {
...
+    .V()
+    .has('Project', 'id', projectId)
+    .outE('HAS_MEMBER')
+    .inV()
+    .has('User', 'id', userId)
+    .hasNext();
+  if (!isMember) return response(403, { error: 'Access denied' });
+
+  if (httpMethod === 'DELETE') {
+    const roleEdge = await g
```
svozza: I think we can remove`getVal` if we tweak the Gremlin query to just check if the edge exits for `owner` and `admin` role:

```js
const allowed = await g
    .V()
    .has('Project', 'id', projectId)
    .outE('HAS_MEMBER')
    .has('role', P.within('owner', 'admin'))
    .inV()
    .has('User', 'id', userId)
    .hasNext();

  if (!allowed) {
    return response(403, { error: 'Only project owners and admins can remove repositories' });
  }
```

lambda/projects/index.js
```diff
@@ -30,6 +34,410 @@ const getConnection = async () => {
...
+    const role = getVal(roleEdge.value, 'role') || 'member';
+    if (role !== 'owner' && role !== 'admin') {
+      return response(403, { error: 'Only project owners and admins can remove repositories' });
+    }
+
+    const repoUrl = event.queryStringParameters?.url;
+    if (!repoUrl) return response(400, { error: 'url query parameter is required' });
+    const decoded = decodeURIComponent(repoUrl);
+
+    const repoExists = await g
```
svozza: We can simplify this to one query if we allow deletes of non-existant repos to return a 200, not sure we get much benefit from return 404, I think:

```
await g
    .V()
    .has('Project', 'id', projectId)
    .out('HAS_REPO')
    .has('Repository', 'url', decoded)
    .drop()
    .iterate();

  return response(200, { removed: decoded });
```

If we want to preserve the 404 behaviour then we can still do this in one query:

```
const dropped = await g
    .V()
    .has('Project', 'id', projectId)
    .out('HAS_REPO')
    .has('Repository', 'url', decoded)
    .aggregate('x')
    .drop()
    .cap('x')
    .next();

  if (dropped.value.length === 0) {
    return response(404, { error: 'Repository not found on this project' });
  }
  return response(200, { removed: decoded });
```

lambda/projects/index.js
```diff
@@ -30,6 +34,410 @@ const getConnection = async () => {
...
+      'git_repo',
+    );
+    await ensureLegacyRepoMigrated(g, projectId, legacyGitRepo);
+    const repos = await fetchRepos(g, projectId);
+    return response(200, repos);
+  }
+
+  // POST /projects/{projectId}/repos
+  if (httpMethod === 'POST') {
+    const roleEdge = await g
```
svozza: We can do a similar simplification as the other `owner` / `admin` query above

lambda/projects/index.js
```diff
@@ -30,6 +34,410 @@ const getConnection = async () => {
...
+      }
+    }
+
+    const newRepoId = `repo-${randomUUID()}`;
+    const addedAt = new Date().toISOString();
+    const repoRole = data.role || detection.role || 'unknown';
+    const provider = data.provider || 'github';
+    const detectedStack = data.detectedStack || detection.summary || '';
+
+    await g
```
svozza, line 542: We should be able to combine these two queries:

```
await g
    .addV('Repository')
      .property('id', newRepoId)
      .property('url', data.url)
      .property('provider', provider)
      .property('role', repoRole)
      .property('detected_stack', detectedStack)
      .property('added_at', addedAt)
      .as('r')
    .V().has('Project', 'id', projectId)
      .addE('HAS_REPO').to('r')
    .iterate();
```

lambda/projects/index.js
```diff
@@ -108,21 +530,29 @@ exports.handler = async (event) => {
...
-            name: getVal(v, 'name'),
-            gitProvider: getVal(v, 'git_provider') || 'github',
-            gitRepo: getVal(v, 'git_repo'),
-            agentCli: getVal(v, 'agent_cli') || 'kiro',
-            issueIntegrationEnabled: getVal(v, 'issue_integration_enabled') === 'true',
-            createdAt: getVal(v, 'created_at') || new Date().toISOString(),
-            userRole: getVal(e, 'role') || 'member',
-          };
-        });
+        const projects = await Promise.all(
```
svozza: If one of these fails the whole operation will, can we use `Promise.allSettled`?

lambda/projects/index.js
```diff
@@ -140,13 +587,41 @@ exports.handler = async (event) => {
...
 
+        // Create Repository vertices and HAS_REPO edges
+        const reposOut = [];
+        for (const repo of inputRepos) {
+          const repoId = `repo-${randomUUID()}`;
+          const addedAt = new Date().toISOString();
+          const repoRole = repo.role || guessRole(repo.url);
+          const provider = repo.provider || data.gitProvider || 'github';
+
+          await g
```
svozza, line 872: Again these two queries can be combined, I think.

frontend/src/pages/ProjectSettings.tsx
JWThewes: As there is no other reference to auth/user/token in the file I think this can be removed

JWThewes:
We have to take care of the issue integration. Today we're only loading issues from the legacy primary project.repo linked GitHub repo.
I do see three options:
1. Minimum: add a banner in IssueListPanel when project.repos.length > 1 clarifying that issues come from the primary repo only — and file a follow-up. Cheap, ships now.
  2. Repo selector: dropdown at the top of the panel (defaulting to primary) so the user picks one repo at a time. Keeps query model simple; sprint linkage needs a repoUrl field added next to issueNumber to avoid #42 collisions.
  3. Aggregated view: union issues across all linked repos with a per-issue repo badge and a repo filter. Most useful UX, but requires changes to githubIssuesService.listPages (parallel fan-out + merging + pagination semantics) and a stronger composite key for sprint linkage.
  
  I would vote for 3. This is a bigger change but it would offer the most consistent user experience

frontend/src/pages/ProjectSettings.tsx
eipasteur: Removed — the ESM version has no auth/user/token reference in this file; imports are router + services + types only.

frontend/src/pages/ProjectSettings.tsx
eipasteur: Done — partial outcomes go through buildAddReposFeedback(succeeded, failed), which returns kind:'warning' when any repo fails. That renders a distinct amber banner (bg-yellow-50 / border-yellow-200 / text-yellow-800) via setWarning, separate from the green success path. Covered by ProjectSettings.test.ts.

frontend/src/pages/ProjectSettings.tsx
eipasteur: Agreed the page is large. Scoping that extraction as a dedicated follow-up rather than widening this multi-repo PR, to keep the diff reviewable. Sections stay comment-delimited until then.

frontend/src/pages/ProjectSettings.tsx
eipasteur: Same follow-up as the section extraction above — the modal moves into its own component in that refactor.

lambda/projects/index.js
eipasteur: Done in the ESM rewrite — fetchRepos now uses .project('url','provider','role','detectedStack','addedAt').by(coalesce(values(...), constant(default))), so defaults are applied in-query. The trailing .map() only marshals the driver's per-row Map into a plain object (the gremlin JS driver returns Map), not value extraction.

lambda/projects/index.js
eipasteur: Kept this multi-step on purpose. A Repository vertex can be shared across projects, so we drop the HAS_REPO edge first, then drop the vertex only if no other HAS_REPO edge still references it. The suggested .out('HAS_REPO').has('Repository','url',decoded).drop() deletes the vertex unconditionally, which would corrupt other projects linked to the same repo. We also read the target's role to promote a new primary after removing the primary one, and to preserve the 404. Happy to adopt the aggregate('x').cap('x') form for the 404 itself, but the edge-then-conditional-vertex split needs to stay.

lambda/projects/index.js
eipasteur: Done — same hasNext() pattern on the POST /repos gate.

lambda/projects/index.js
eipasteur: Done — single traversal: addV('Repository')...as('r').V().has('Project',...).addE('HAS_REPO').to('r'). Same pattern in the create-project loop and in the legacy-repo migration.

lambda/projects/index.js
eipasteur: Done — the project-list enrichment uses Promise.allSettled and omits (with a logged warning) any project that fails to enrich, so one bad project no longer 500s the whole list.

lambda/projects/index.js
eipasteur: If this is the addV+addE pair, it's now a single traversal (see above). If you meant a different pair here, let me know and I'll fold it.

lambda/projects/index.js
svozza: Makes sense

frontend/src/pages/ProjectSettings.tsx
svozza: Yeah, good call. Can we create an issue for this so it doesn't get lost?

lambda/projects/index.js
svozza: Looks good now

eipasteur:
> Now that #181 has been merged, can we add some integration tests for the new Gremlin queries that have been added to the `projects` lambda in this PR?

Done — `lambda/projects` runs against a real `tinkerpop/gremlin-server` testcontainer (`test/gremlin-setup.js`, shared via vitest `globalSetup`, isolated per-file with `PartitionStrategy`). 60 tests cover POST/GET/PUT/DELETE + the `/repos` routes and their Gremlin traversals.

eipasteur:
> We have to take care of the issue integration. Today we're only loading issues from the legacy primary project.repo linked GitHub repo. I do see three options: [...] I would vote for 3.

Option 3 is implemented in this branch. `githubIssuesService.listAcrossRepos` fans out across every linked repo with `Promise.allSettled`, merges and sorts by most-recently-updated, tags each issue with its source `repo`, and surfaces per-repo failures instead of failing the whole load. The panel renders a per-repo badge + a repo filter (defaulting to all). For sprint linkage I used the composite key you flagged — `repo#issueNumber` — with legacy sprints (no `repository`) keyed under the primary repo so they keep matching, which covers the #42-collision case. Aggregation lives in `githubIssues.ts` / `IssueListPanel.tsx`; happy to walk through the pagination/cap semantics if you want to tune `PER_PAGE`.

eipasteur:
> Can you rebase and run `npm run lint` to fix the CI/CD?

Done — rebased onto current `main`. `npm run lint` (oxlint) is clean and `npm run format:check` (oxfmt) passes; only pre-existing non-blocking warnings remain (`from_` dangle in sprints/tasks, `toSorted` suggestions). `npm test` is green (223 tests across the lambda suites).

frontend/src/pages/ProjectSettings.tsx
eipasteur: Filed #235.

eipasteur:
**Force-pushed** — rebased onto current `main` and reimplemented the multi-repo backend in **ESM**.

What changed vs the previous head (`a16c820`):
- Gremlin consolidation feedback folded in: `project`/`coalesce`, single `addV`+`addE`, `hasNext` role checks, `Promise.allSettled` — replies inline.
- One reasoned pushback on the repo-delete: it drops the `HAS_REPO` edge then the `Repository` vertex **only if no other project references it** (shared-vertex safety), so it can't stay a single unconditional `drop()`.
- Issue integration: option 3 (aggregated view across all linked repos + per-repo badge + filter, composite `repo#issueNumber` sprint key) is implemented.
- ProjectSettings component extraction tracked as a follow-up in #235.
- `npm run lint` / `format:check` clean, `npm test` green (223 tests, real gremlin-server testcontainer).

eipasteur:
Rebased onto current `main` and reconciled with the work that landed since this branch was opened:

- **#200 (Tracker provider abstraction)** — dropped the standalone GitHub-issue-aggregation commits (`githubIssues.ts` was removed on `main`); the "issues across a project's repos" need is now expressed through #200's per-binding tracker model. Multi-repo **project** management (`/repos` routes, `fetchRepos`, `repos` on the project) is kept and merged alongside #200's `trackers` (both coexist on the project).
- **#215 (per-agent review verdicts)** — the multi-PR review (`extractPrs`/`PrInfo`, per-PR tabs) is re-implemented on top of #215's reworked `ReviewPage`; the now-removed `review.riskScore`/`riskReasoning` references are gone, `PARTIAL` is deduped against #215's.
- **#226 (pool-worker module extraction)** — the multi-repo construction prompt block is ported into the new `construction-orchestrator-prompt.js` module.
- **#205 (scoped MCP servers)** — `mcp-server-graph` reconciled (multi-repo `pr_state`, `repository` on CodeFile, #215 Review schema preserved).

Verified locally: `tsc -b` clean, **431 tests** pass, `oxlint`/`oxfmt` clean, `terraform validate` OK.

Validated end-to-end on staging via the multi-repo e2e harness (`review-cycle`): multi-repo project creation, full inception→construction→review, and the blind+full review pipeline all work. The harness surfaced one pre-existing multi-repo PR-creation timing gap (a repo whose sprint branch isn't ahead-of-base at trigger time doesn't get a PR) — tracked separately in #239.

Component extraction of `ProjectSettings` (incl. re-adding the repo-management UI on the modernized #200 surface) remains deferred to #235.

svozza (dismissed):

JWThewes (dismissed):

JWThewes (approved):

PR #184, opened 2026-05-22
feat(ui): unified sidebar + project cache + homepage with Issues/Iterations

eipasteur:
## Summary

Rework of the user experience around projects: global sidebar with status dots, stale-while-revalidate cache for projects + sprints, and project homepage that aligns Issues (via #171) and Iterations in parallel columns.

### `hooks/useProjectsCache.ts` — stale-while-revalidate cache

- `useSyncExternalStore` pattern: projects cached 120s (TTL) with auto-refresh, sprints cached 60s per-project.
- Avoids redundant `/api/projects` calls when navigating between sidebar entries; returns stale view immediately while a background refresh validates.
- Pure client-side logic — no backend changes.

### Layout (`AppShell` + `AppSidebar` + `SprintPipelineBar`)

- **Global sidebar** lists all projects with live status dots (running/waiting/failed/idle). Clicking navigates to the project page.
- **Observability section** in sidebar shows active iterations with status icons; clicking navigates to per-iteration observability page.
- **SprintPipelineBar** (new) replaces the old vertical PipelineView with a horizontal progress-bar pipeline (per-phase progress indicators, responsive).
- **AppHeader** streamlined — removed observability button (accessed via sidebar now).

### `pages/Project.tsx` — project homepage

- Two horizontally aligned cards: **Repository** info and **Iterations** (count + active + last activity timestamp).
- Headers share `h-7` height + matching spacing so columns line up.
- "Start" iteration button on the right of the Iterations title.
- Live updates via `useProjectSprintsCache`.
- Removed redundant "Agent status" card (visible from observability page).

### `pages/Dashboard.tsx`

- Refactored to consume `useProjectsCache` instead of fetching directly. Error + retry states, loading skeleton.

### `lib/sprintStatus.ts` (new)

- `effectiveSprintStatus` helper: normalizes sprint phase + agent status into a single status enum (`running | waiting | completed | failed | passed | idle`).
- Shared across sidebar, pipeline bar, observability, and project pages for consistent status semantics.

## Validation

- ✅ TypeScript clean
- ✅ `npm run build` clean
- ✅ oxlint 0 errors
- ✅ oxfmt passes
- ✅ Cache hooks tested: stale-while-revalidate works, no double-fetch, cache invalidation on `refresh()` call
- ✅ Sidebar status dots verified on multiple projects
- ✅ Issues/Iterations alignment verified

## Stacked

PR #185 (observability) is stacked on this branch.


frontend/src/hooks/useProjectsCache.ts
```diff
@@ -0,0 +1,195 @@
...
+  sprintsVersion++;
+  sprintsListeners.get(projectId)?.forEach((fn) => fn());
+}
+
+function isStale(entry: { fetchedAt: number } | null, ttl: number): boolean {
+  if (!entry) return true;
+  return Date.now() - entry.fetchedAt > ttl;
+}
+
+async function fetchProjects(): Promise<ProjectWithSprint[]> {
```
svozza, line 58: Not for now but we should look at using something like [Tanstack Query](https://tanstack.com/query/latest) or [SWR](https://swr.vercel.app/) for API request like this, you get caching for free and bunch of other QOL improvements. Would probably reduce the size of this file by 50% or more

frontend/src/pages/Project.tsx
```diff
@@ -1,276 +1,321 @@
...
+};
+
+const STATUS_LABEL: Record<string, string> = {
+  running: 'Running',
+  waiting: 'Waiting for input',
+  completed: 'Completed',
+  failed: 'Failed',
 };
 
+function formatRelativeTime(dateStr: string | null): string {
```
svozza, line 70: We might want to use a dedicated time library like for this this sort of stuff. It's notoriously full of edge cases (I used to maintain a date time library). Something like datefns, which is highly treeshakeable so we only ship the functions that we use: https://date-fns.org/v4.3.0/docs/formatDistanceToNowStrict. 

```js
function formatRelativeTime(dateStr: string | null): string {
  if (!dateStr) return '';
  const date = new Date(dateStr);
  const ms = Date.now() - date.getTime();
  if (ms < 60_000) return 'just now';
  return formatDistanceToNowStrict(date, { addSuffix: true, roundingMethod: 'floor' });
}
```

frontend/src/pages/Project.tsx
```diff
@@ -1,276 +1,321 @@
...
-              {sprints.length} sprint{sprints.length !== 1 ? 's' : ''}
-            </p>
-          </div>
-          <Button onClick={() => setShowCreate(true)} size="sm" className="gap-1.5">
-            <Plus className="h-3.5 w-3.5" />
-            New Sprint
-          </Button>
+  if (!project) return <div className="p-6">Project not found</div>;
+
+  return (
```
svozza, line 187: Can we look at breaking this large block of HTML up into React components where it makes sesne?

svozza:
Can we do `npm run lint` in the frontend to fix the CI/CD?

JWThewes:
If we create issues for the cache and time stuff and run the linter it looks good to me!

frontend/src/pages/Project.tsx
```diff
@@ -109,205 +140,226 @@ export default function Project() {
...
+      <div className="grid md:grid-cols-2 gap-3">
+        <Card>
+          <CardContent className="px-4 py-3">
+            <div className="flex items-center gap-2 text-xs text-muted-foreground mb-1">
+              <FolderGit2 className="h-3 w-3" />
+              Repository
+            </div>
+            {project.gitRepo ? (
+              <>
+                <p className="text-sm font-medium truncate">{project.gitRepo}</p>
```
jeromevdl: maybe we can add a link to github repo?

frontend/src/pages/Project.tsx
```diff
@@ -109,205 +140,226 @@ export default function Project() {
...
-            <CardContent className="flex flex-col items-center justify-center py-12">
-              <div className="h-12 w-12 rounded-full bg-muted flex items-center justify-center mb-3">
-                <Zap className="h-6 w-6 text-muted-foreground" />
+        )}
+
+        <Card className="flex flex-col">
+          <CardHeader className="pb-3">
+            <div className="flex items-center justify-between gap-3 min-h-7">
+              <div className="flex items-center gap-2">
+                <CardTitle className="text-sm">Iterations</CardTitle>
```
jeromevdl, line 302: Image

Why do we have 2 "Iterations" blocks ? Could we merge them somehow ? Also can we align on the vocabulary "Iteration" vs "Sprint" => the button says Sprint and the titles "Iteration"

Proposition:

Image

and the empty state:

Image

I'm not even sure we need to display the repo (it's in the settings....)

frontend/src/hooks/useProjectsCache.ts
eipasteur: Tracked in #276. Keeping the hand-rolled version for now to limit scope here but the migration makes sense as a next step.

frontend/src/pages/Project.tsx
eipasteur: Switched to `formatDistanceToNow` from date-fns in this PR.

frontend/src/pages/Project.tsx
eipasteur: Tracked in #277. Keeping it as-is for this PR to avoid inflating the diff but will extract components in a follow-up.

frontend/src/pages/Project.tsx
eipasteur: Added — repo name is now a link to GitHub.

frontend/src/components/layout/SprintPipelineBar.tsx
```diff
@@ -0,0 +1,316 @@
...
+
+        {sprint?.prUrl && (
+          <Tooltip>
+            <TooltipTrigger asChild>
+              <Button
+                variant="ghost"
+                size="sm"
+                className="h-7 w-7 p-0"
+                aria-label="Open pull request on GitHub"
+                onClick={() => window.open(sprint.prUrl!, '_blank', 'noopener,noreferrer')}
```
JWThewes: Can we keep the previous behavior here and route to the review page instead of opening sprint.prUrl directly? sprint.prUrl is only the denormalized first PR for a sprint; in multi-repo sprints this shortcut sends the user to just one repo PR and hides the rest. The old sidebar intentionally navigated to review so the full multi-PR list stayed visible.

frontend/src/hooks/useProjectsCache.ts
```diff
@@ -0,0 +1,247 @@
...
+
+export async function getProjectsWithSprints(): Promise<ProjectWithSprint[]> {
+  if (projectsCache && !isStale(projectsCache, PROJECTS_TTL)) return projectsCache.data;
+  await revalidateProjects();
+  if (projectsCache) return projectsCache.data;
+  throw new Error(projectsError ?? 'Failed to load projects');
+}
+
+async function revalidateSprints(projectId: string, force = false) {
+  if (sprintsFetching.has(projectId)) return;
```
JWThewes: I think the forced refresh path needs to win over any in-flight request. Right now a refresh after create/delete/update can be a no-op if the initial sprint fetch is already running, which means a pre-mutation response can repopulate the cache and the UI will stay stale until the TTL/poll. Can we queue a follow-up fetch when force is true, or tag requests so older responses cannot overwrite newer refreshes? Same concern applies to invalidateProjects while projectsInflight exists.

frontend/src/hooks/useProjectsCache.ts
```diff
@@ -0,0 +1,247 @@
...
+async function fetchProjects(): Promise<ProjectWithSprint[]> {
+  const projs = await projectsService.list();
+  const results = await Promise.allSettled(
+    projs.map(async (project): Promise<ProjectWithSprint> => {
+      const cached = sprintsCache.get(project.id);
+      if (cached && !isStale(cached, SPRINTS_TTL)) {
+        return { project, latestSprint: latestOf(cached.data) };
+      }
+      try {
+        const sprints = await sprintsService.list(project.id);
```
JWThewes, line 67: This is going to make the always-mounted sidebar call /projects and then /projects/:id/sprints for every project on load and on the poll interval. That scales linearly with project count and makes basic navigation depend on a bunch of unrelated sprint-list calls. Could we expose latestSprint as part of the projects projection, or add a dedicated lightweight endpoint for the sidebar/dashboard and keep full sprint-list loading scoped to the selected project page?

eipasteur:
Thanks for the feedback @JWThewes .

Addressed:

1. SprintPipelineBar: the PR button now routes to the sprint review page
   (/project/:id/sprint/:id/review) instead of opening sprint.prUrl, so the full
   multi-PR list stays visible for multi-repo sprints.

2. useProjectsCache forced-refresh race: a forced refresh now wins over an
   in-flight fetch: it chains a single follow-up after the in-flight one settles
   and returns THAT promise, so callers await post-mutation data rather than a
   possibly pre-mutation response. Concurrent forced callers coalesce; bounded to
   one follow-up (applies to both revalidateProjects and revalidateSprints).

3. Sidebar N+1 (/projects + /projects/:id/sprints per project): this needs
   latestSprint in the /projects projection (or a dedicated endpoint), so it can't
   land cleanly in this frontend PR. Tracked in #286 
   Kept the hand-rolled fetch here to limit scope.

Rebased onto current main (resolved conflicts from #256/#270/#284). tsc + oxlint
(0 errors) + oxfmt all green.

JWThewes (approved):

frontend/src/pages/Project.tsx
jeromevdl: @eipasteur @JWThewes I don't think this was addressed... not commented. The proposed UI is not clear.

JWThewes: Oh totally oversaw that. @eipasteur can you tackle in a separate PR?

eipasteur: I will take a look at this in a new PR yes!

The focus of this PR was more the main side bar on the left than the homepage, which I agree still needs some improvements and refining

eipasteur: @jeromevdl I will use your proposition for the next Frontend PR

Issue #198, opened 2026-05-28
[Feature]: Jira support — Phase 4: migration polish + observability (no deletions)


JWThewes:
Part of #194. Depends on Phase 3.

## Goal

Polish the tracker abstraction now that all three providers (GitHub Issues, Jira Cloud) are live. **No deletion happens here** — this is open source, so the legacy `issue_number` / `issue_url` / `issueIntegrationEnabled` properties, the dual-shape readers, the per-project migration endpoint, the migration banner, and the bulk migration Lambda all stay deployed indefinitely. We have no control over when (or whether) downstream forks migrate, so the cost of removing safety nets is unbounded.

## What this phase actually does

1. **Migration observability** — surface a count of "projects still on the legacy data model" in the admin/observability page so operators of installs with hundreds of projects can tell at a glance how much data is still on the old shape. Driven off the same query the migration banner uses (`issue_integration_enabled='true'` AND no outgoing `HAS_TRACKER`).

2. **Bulk-migrate-from-admin-UI** — promote the bulk migration from a CLI-only Lambda to an admin-page button so operators don't need shell access to run it across many projects at once. Calls the same lambda; results displayed inline.

3. **Run the bulk migration in CI for the demo environment** — when contributors deploy the demo against a clean Neptune from main, the migration is unnecessary; when they deploy against a long-running cluster with pre-#194 data, it runs idempotently. Either way no manual step.

4. **Documentation sweep** — add a "Migrating from legacy issue integration" section to the relevant docs explaining the in-product banner, the bulk lambda, and the fact that legacy data keeps working forever. Cross-link from the GitHub Issues + Jira sections.

5. **Telemetry hooks (optional)** — if there's an opt-in telemetry channel by the time this phase lands, count how many migrations run per release so we can tell if the in-product path is reaching users. No telemetry is added if there isn't already an opt-in channel.

## Why no deletion

Originally this phase was "remove the legacy fields and the shim routes." That can't happen safely:

- We cannot tell when an OSS install has finished migrating, because we don't know about the OSS install at all.
- A fork that hasn't pulled in months might `git pull` and find their existing sprints rendering empty if we'd removed the fallback readers.
- The migration UI + lambda must stay available because we don't know when (or whether) a downstream user will upgrade.

So Phase 4 is permanent-coexistence cleanup: the abstraction is the primary path, the legacy path is the safety net, both stay forever.

## Acceptance criteria

[...]

Issue #194, opened 2026-05-28
[Feature]: Add Jira Cloud support (and a tracker-provider abstraction)


JWThewes:
### Description

Add **Jira Cloud** as a task-tracker provider so a sprint can be started from a Jira issue (Story/Task/Bug) the same way it can be started from a GitHub issue today (#171). In the same change, introduce a generic **tracker provider abstraction** so future trackers (Linear, GitLab issues, Azure Boards, …) plug in cleanly, and migrate the existing GitHub-issues integration onto that abstraction so we stop hardcoding `gitRepo` as the issue source.

### Use case

Many teams plan in Jira, not in GitHub Issues — especially in enterprises and regulated environments. Today they cannot use this platform's "start a sprint from an issue" flow without first mirroring tickets into GitHub. They also expect to keep planning in Jira even after we add multi-repo (#14) and GitLab (#55), where the tracker is decoupled from the code host. Building the abstraction once, behind Jira, keeps subsequent providers cheap to add.

### Area

Backend (Lambda)

## Current state

- `lambda/github-issues/index.js` is hardcoded to GitHub REST and reads `Project.gitRepo` (`owner/repo`) as its issue source.
- `Project` (Neptune vertex) carries a single `git_provider` + `git_repo` + boolean `issue_integration_enabled` (`lambda/projects/index.js:67`-`219`). This conflates **code host** with **task tracker**.
- `Sprint` stores GitHub-specific `issue_number` (numeric) + `issue_url` strings (`lambda/sprints/index.js:28`-`107`). The frontend round-trips them via `IssueListPanel` and `sprintsService.create({issueNumber, issueUrl})`.
- The frontend `IssueListPanel` parses `gitRepo` to `owner/repo` and is hard-bound to `gitProvider === 'github'` (`frontend/src/pages/Project.tsx:243`).
- `git-connections` DynamoDB row is keyed on `userId` only, with a `provider` attribute already present (today only `'github'`). Token resolution lives in `lambda/shared/git-token.js`.

## Scope of this issue

1. **A tracker provider abstraction** with Jira Cloud as the first non-GitHub implementation.
2. **Polymorphic Sprint→tracker-resource link** replacing GitHub-specific `issue_number`/`issue_url`.
3. **Many-to-many Project↔tracker bindings** so one collaborative project can be linked to multiple Jira projects (or other trackers) — and so this composes cleanly with future multi-repo (#14).
[...]

PR #199, opened 2026-05-28
Phase 1: tracker provider abstraction — data model + migration UI (#195)

### LINKED ISSUE #195: [Feature]: Jira support — Phase 1: data model + migration
Part of #194.

## Goal

Land the data-model foundation for the tracker-provider abstraction. After this phase, GitHub Issues works exactly as it does today, but the underlying state is in the new shape and ready for the provider abstraction (Phase 2) and Jira (Phase 3). Existing data is **never broken** — both shapes are read transparently — and projects on the legacy shape get a clear in-product migration path.

## Scope

1. **`tracker-connections` DynamoDB table** — keyed `PK=userId`, `SK=providerInstance` (e.g. `github-issues#public`, `jira-cloud#cloud`). Attributes: `parameterName`, `scope`, `cloudId` (Jira-only), `baseUrl` (Jira-only), `expiresAt` (Jira-only), `createdAt`. Provisioned alongside the existing `git-connections` table; **`git-connections` is left untouched** — it stays as the code-host connection store. See parent issue §3a for full schema.

2. **`HAS_TRACKER` edge in Neptune** — `(Project)-[HAS_TRACKER]->(TrackerBinding)` where TrackerBinding holds `{id, provider, instance, externalProjectKey, displayName, createdAt, createdBy}`. See parent issue §3b.

3. **Polymorphic Sprint→tracker fields** — add `tracker_provider`, `tracker_instance`, `tracker_external_project_key`, `tracker_resource_type`, `tracker_resource_id`, `tracker_resource_url` to the Sprint vertex. The existing `issue_number` and `issue_url` properties are kept on the vertex **forever** — pre-migration data must keep rendering for OSS users on long-running deployments.

4. **Always-on dual-shape code paths** in the sprints + projects lambdas:
   - **Read**: prefer the new `tracker_*` fields when set; fall back to `issue_number`/`issue_url` for unmigrated data. Both paths are unconditional — no feature flag.
   - **Write**: always populate the polymorphic fields. New writes happen on the new shape; legacy writes keep the legacy properties for compatibility.
   - The `Project` API response always includes `trackers: TrackerBinding[]` (empty for unmigrated projects).
   - The `Sprint` API response always includes `tracker: { ... } | null`. `issueNumber`/`issueUrl` are derived from the tracker when it's a github-issues resource, or read from the legacy properties otherwise.

5. **Per-project migration endpoint** `POST /projects/{projectId}/migrate-tracker` (owner/admin only). Backfills this project's `HAS_TRACKER` binding + all its sprints' `tracker_*` fields in one call. Idempotent. Supports `{dryRun: true}` payload. Returns `{ dryRun, projects: {candidates, applied}, sprints: {candidates, applied} }`.

6. **In-product migration UI** in `ProjectSettings.tsx`. The `MigrateTrackerCard` renders only when:
   - `project.issueIntegrationEnabled === true` (legacy flag was set), AND
   - `project.trackers.length === 0` (no migration has run yet).

   Owner/admin sees a "Migrate now" button that calls the endpoint above. After success the page reloads, the trackers array populates, the card disappears. Plain members see explanatory text but no button.

7. **Bulk admin migration Lambda** at `lambda/migrate-tracker-fields/`. Walks every Project + Sprint in Neptune in one shot. Same code path as the per-project endpoint (shared `lambda/shared/tracker-migration.js`). Pattern: identical to `lambda/purge-neptune` (admin utility, VPC-attached, invoked directly via `aws lambda invoke`, reuses the `neptune-reader` IAM role). Useful for operators with many projects who'd rather not click through each one.

   ```bash
   # Dry-run preview
   aws lambda invoke --function-name $(terraform output -raw migrate_tracker_fields_lambda_name) \
     --payload '{"dryRun":true}' --cli-binary-format raw-in-base64-out /tmp/out.json

   # Apply
   aws lambda invoke --function-name $(terraform output -raw migrate_tracker_fields_lambda_name) \
     --payload '{}' --cli-binary-format raw-in-base64-out /tmp/out.json
   ```

## Why this shape (and not a feature flag + ops runbook)

This is open source. We have **zero control** over when, or whether, users migrate. Three consequences:

- **No feature flag.** A flag means the new code is wrong-by-default. Operators forget to flip it and end up with sprints that should have new fields but don't. Always-on dual-shape code is the only safe default.
- **Migration is in-product, not a runbook.** A user upgrading their fork shouldn't need to read release notes to keep their data working. If the legacy data model is detected, the project settings page tells them so and offers a one-click migration. The bulk lambda is a power-tool for operators with many projects, not the primary path.
- **Legacy readers + migration tooling stay forever.** No upgrade window, no deprecation cycle. Phase 4's original "delete legacy fields" goal is impossible — we have to keep both shapes load-bearing for the lifetime of the project.

## Out of scope

- The provider abstraction itself (Phase 2).
- Jira (Phase 3).
- Removing anything (Phase 4 is now about polish, not deletion — see #198).

## Acceptance criteria

- [ ] `tracker-connections` table is provisioned via Terraform and present in dev.
- [ ] Pre-migration projects + sprints render exactly as they did before, with `tracker = null` and `trackers = []`. No user-facing change for unmigrated data.
- [ ] `POST /projects/{id}/migrate-tracker` (owner/admin only): on a legacy project, single call backfills the `HAS_TRACKER` edge and every sprint's `tracker_*` fields. Re-running returns `applied: 0`. Plain members get 403.
- [ ] `MigrateTrackerCard` appears on ProjectSettings for legacy projects and disappears after a successful migration.
- [ ] `migrate_tracker_fields_lambda` deploys and is invokable; payload `{"dryRun":true}` returns counts without writing; payload `{}` applies and re-running is a no-op.
- [ ] Both the per-project endpoint and the bulk lambda call into a single shared helper (`lambda/shared/tracker-migration.js`) so they cannot drift.
- [ ] Frontend type-checks; no console warnings on the settings page in either pre- or post-migration state.
- [ ] All tests pass.

## Dependencies

None. Foundational phase.

## Unblocks

- Phase 2 ([Feature]: Tracker provider abstraction + GitHub Issues port)
- Phase 3 ([Feature]: Jira Cloud provider + UI)

JWThewes:
**Phase 1 of #194 — tracker provider abstraction.** This PR targets `feat/194-jira-tracker-abstraction`, not `main`. The parent branch consolidates all four phases into a single PR against `main` once the phase branches have all merged.

Closes #195.

## Summary

Lays the data-model foundation for Jira (and any future tracker providers) and ships an in-product migration so OSS forks with pre-#194 data can self-serve onto the new shape. After this PR, GitHub Issues works exactly as it does today; nothing user-visible changes for projects on the legacy data model until the user clicks **Migrate now** on the project settings page.

### Why no feature flag

This is open source. We can't tell when, or whether, downstream forks migrate. A flag would mean the new code is wrong-by-default — operators forget to flip it and run with broken state. Always-on dual-shape readers are the only safe default: prefer `tracker_*` when present, fall back to `issue_number`/`issue_url` for unmigrated rows. Both shapes stay readable forever; Phase 4 (#198) doesn't delete anything.

See the parent issue (#194) for the full architectural picture and the *why nothing is ever removed* reasoning.

## What's included

**Data model**
- New `tracker-connections` DynamoDB table, keyed `userId#providerInstance` (sibling to `git-connections`; populated by Phase 3 when Jira lands)
- `HAS_TRACKER` edge + `TrackerBinding` vertex on Neptune (Project ↔ N trackers)
- Polymorphic `tracker_*` properties on Sprint vertices (`tracker_provider`, `tracker_instance`, `tracker_external_project_key`, `tracker_resource_type`, `tracker_resource_id`, `tracker_resource_url`)

**Lambdas**
- Always-on dual-shape readers in `lambda/sprints` and `lambda/projects` — no feature flag; new fields preferred, legacy fields used as fallback
- New `POST /projects/{projectId}/migrate-tracker` route on the projects lambda (owner/admin only, idempotent, supports `{dryRun: true}`)
- New bulk `lambda/migrate-tracker-fields/` admin lambda (VPC-attached, invoked via `aws lambda invoke`, sibling to `purge-neptune`)
- Shared `lambda/shared/tracker-migration.js` so the per-project endpoint and the bulk lambda can't drift
- `lambda/sprints` migrated from CJS to ESM + workspace esbuild bundling (matches projects/github-issues/migrate-tracker-fields)

**Frontend**
- `MigrateTrackerCard` in `ProjectSettings.tsx` — appears only when `project.issueIntegrationEnabled === true && trackers.length === 0`. Owner/admin gets a **Migrate now** button; plain members see explanatory text
- `Project` and `CreateSprintInput` types extended with `trackers: TrackerBinding[]` and the `tracker` payload
- New `projectsService.migrateTracker(projectId, dryRun)` service method

**Infra**
- Terraform variables, outputs, IAM, route, and CORS wiring for the new endpoint and lambda
- API Gateway redeployment trigger updated to include the new `migrate-tracker` resource

## Migration flows (after this PR ships)

| Scenario | What the user sees |
|---|---|
| Fresh deploy, clean Neptune | Nothing. New sprints automatically populate both shapes. The migration card never appears. |
| Upgrade onto cluster with pre-#194 data | Existing sprints render exactly as before (legacy fallback). Owner/admin sees an amber "Migrate to the new tracker data model" card on Project Settings. One click backfills that project + its sprints. |
| Operator with hundreds of projects | Same per-project card *or* one shot via `aws lambda invoke --function-name $(terraform output -raw migrate_tracker_fields_lambda_name) --payload '{}' …`. Phase 4 (#198) promotes this to an admin-UI button. |

## Test plan

- [x] `npm test` — 188/188 passing (was 180; +3 bulk-lambda tests, +5 per-project-endpoint tests, sprint tests reworked for unconditional behavior)
- [x] Per-project migration: legacy project + sprints → migrate → trackers populated → re-run is no-op
- [x] Bulk lambda dry-run reports candidates without writing
- [x] Plain members get 403 from the migration endpoint; non-members get 403
- [x] `terraform validate` passes for `modules/api/lambda` and `modules/git`
- [x] All four lambdas build via `npm run build -w <name>`
- [x] Frontend type-checks (`npx tsc --noEmit`)
- [ ] Manual: deploy to dev and verify the card appears for a legacy project, disappears after migration, and Jira-shaped tracker payloads round-trip
- [ ] Manual: confirm `aws lambda invoke` against the bulk lambda works end-to-end in the dev VPC


lambda/projects/index.js
```diff
@@ -18,6 +19,27 @@ const getVal = (obj, key) => {
   return raw ?? '';
 };
 
+const mapTrackerBinding = (v) => ({
```
svozza: We should be able to remove this mapper and do everything in gremlin:

```js
const fetchProjectTrackers = (g, projectId) =>
    g.V()
      .has('Project', 'id', projectId)
      .out('HAS_TRACKER')
      .hasLabel('TrackerBinding')
      .project('id', 'provider', 'instance', 'externalProjectKey', 'displayName', 'createdAt', 'createdBy')
        .by('id')
        .by('provider')
        .by(coalesce(values('instance'), constant(null)))
        .by(coalesce(values('external_project_key'), constant(null)))
        .by(coalesce(values('display_name'), constant(null)))
        .by(coalesce(values('created_at'), constant(null)))
        .by(coalesce(values('created_by'), constant(null)))
      .toList();
```

lambda/projects/index.js
```diff
@@ -59,10 +81,37 @@ export const handler = async (event) => {
...
     const projectId = pathParameters?.projectId;
     const userId = event.requestContext?.authorizer?.claims?.sub;
     const userEmail = event.requestContext?.authorizer?.claims?.email || '';
+    const isMigrateTracker = httpMethod === 'POST' && path?.endsWith('/migrate-tracker');
+
+    // POST /projects/{projectId}/migrate-tracker — owner/admin only.
+    // Backfills the tracker_* fields on this project's sprints + creates a
+    // synthetic HAS_TRACKER edge if the project still uses the legacy
+    // issue_integration_enabled boolean. Idempotent. See parent issue #194.
+    if (isMigrateTracker) {
```
svozza, line 125: We should be able to simplify this by leaning on Gremlin:

```js
const role = await g
    .V().has('Project', 'id', projectId)
    .outE('HAS_MEMBER').as('e')
    .inV().has('User', 'id', userId)
    .select('e').values('role')
    .next();

  if (role.done) return response(403, { error: 'Access denied' });
  if (role.value !== 'owner' && role.value !== 'admin') {
    return response(403, { error: 'Only project owners and admins can migrate trackers' });
  }
```

lambda/projects/index.js
```diff
@@ -116,21 +167,26 @@ export const handler = async (event) => {
...
-            name: getVal(v, 'name'),
-            gitProvider: getVal(v, 'git_provider') || 'github',
-            gitRepo: getVal(v, 'git_repo'),
-            agentCli: getVal(v, 'agent_cli') || 'kiro',
-            issueIntegrationEnabled: getVal(v, 'issue_integration_enabled') === 'true',
-            createdAt: getVal(v, 'created_at') || new Date().toISOString(),
-            userRole: getVal(e, 'role') || 'member',
-          };
-        });
+        const projects = await Promise.all(
```
svozza: If one of these fails in `Promise.all`everything does, can we use `Promise.allSettled` nd handle partial failure?

lambda/shared/tracker-migration.js
```diff
@@ -0,0 +1,148 @@
...
+    .by(__.in_('HAS_SPRINT').hasLabel('Project').valueMap())
+    .toList();
+};
+
+// Run the migration. When projectId is provided, scoped to that project.
+// When omitted, runs across the whole graph (admin/bulk path).
+//
+// Returns { dryRun, projects: {candidates, applied}, sprints: {candidates, applied} }.
+const runTrackerMigration = async (g, { projectId, dryRun = false } = {}) => {
+  const projectRows = await findProjectsNeedingMigration(g, projectId);
```
svozza: These can be run concurrently with `Promise.all`:

```js
const [projectRows, sprintRows] = await Promise.all([...]);
```

frontend/src/pages/ProjectSettings.tsx
```diff
@@ -372,6 +403,51 @@ export default function ProjectSettings() {
           <div className="text-center py-12 text-muted-foreground text-sm">Loading...</div>
         ) : (
           <>
+            {/* Tracker-abstraction migration banner — appears only when this
+                project still uses the legacy issue_integration boolean and
+                has no HAS_TRACKER edge yet. Disappears once migrated.
+                See parent issue #194 (Jira / tracker provider abstraction). */}
+            {project &&
```
svozza: Could this be abstracted into a React component?

JWThewes:
@svozza I addressed your findings with the latest commit and fixed the lint issues as well

svozza (approved):

svozza:
Nice. Approved!

PR #207, opened 2026-05-31
fix: getting started env/region handling

### LINKED ISSUE #206: [Docs]: Getting Started conflates AWS profile name with environment (dev|prod)
### Description

The README's "Getting Started" passes `<your-profile-name>` as the positional argument to `bootstrap.sh`, `deploy-terraform.sh`, and `deploy-frontend.sh`, but those scripts only accept `dev` or `prod` (the **environment**, not the AWS profile). Step 2 also names the tfvars file `<your-profile-name>.tfvars`, while `deploy-terraform.sh` actually loads `environments/${ENVIRONMENT}.tfvars`. Following the README verbatim fails on step 1 with `Usage: ./scripts/bootstrap.sh [dev|prod]`. 

Bonus: `export REGION=<your-region>` is documented but `bootstrap.sh` hardcodes `REGION="us-east-1"` and ignores it, so the state bucket lands in the wrong region.

Repro:

```text
$ export AWS_PROFILE=my-profile
$ export REGION=eu-central-1
$ ./scripts/bootstrap.sh my-profile
Usage: ./scripts/bootstrap.sh [dev|prod]
$ ./scripts/bootstrap.sh dev
# bucket created in us-east-1, not eu-central-1
```

### Location

- `README.md` → "Getting Started", steps 1–3 and 6
- `scripts/bootstrap.sh` (hardcoded `REGION="us-east-1"`)

### Suggestion

- Replace every `<your-profile-name>` used as a script argument with `dev` (or `prod`); keep `<your-profile-name>` only for `export AWS_PROFILE=...`.
- Step 2: `cp terraform/environments/dev.tfvars.example terraform/environments/dev.tfvars` (filename must match the environment, not the profile).
- Make the README and `bootstrap.sh` agree on region: either drop `export REGION=...` from the README, or change `bootstrap.sh` to honor `${REGION:-us-east-1}`.

### Area

Getting started

eercanayar:
Closes #206.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,14 +37,17 @@ This is a one-time setup. The bootstrap script creates the S3 bucket (with a ran
 ```bash
 export AWS_PROFILE=<your-profile-name>
 export REGION=<your-region>
-./scripts/bootstrap.sh <your-profile-name>
+export ENVIRONMENT=dev    # or: prod
+./scripts/bootstrap.sh $ENVIRONMENT
 ```
+
+> Keep these three variables exported in your shell for the remaining steps.
 
 ### 2. Configure Terraform Variables
 
 ```bash
-cp terraform/environments/dev.tfvars.example terraform/environments/<your-profile-name>.tfvars
-# Edit dev.tfvars with your desired region, etc.
+cp terraform/environments/dev.tfvars.example terraform/environments/$ENVIRONMENT.tfvars
+# Edit only if you need to override defaults; region comes from $REGION.
 ```
 
 ### 3. Deploy Infrastructure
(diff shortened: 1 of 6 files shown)


JWThewes:
I don't think there's any reason to just have dev or prod.
We should remove that restriction and allow any environment name that users want to use.
Also I think the current implementation conflicts with already existing deployments where the terraform state has been written already.

Could you verify with an existing deployment and also remove the dev|prod restriction completly?

JWThewes:
V2 brings a new install mechanism which respects this

PR #213, opened 2026-06-01
chore: add quality guardrails using pre-commit hooks

jeromevdl:
## Add agentic guardrails (local pre-commit checks)

Adds inner-loop guardrails so quality/security checks run locally before each commit, giving humans and AI agents fast feedback instead of waiting for CI.

### What's included
- **Pre-commit framework** (`.pre-commit-config.yaml`) wiring up: oxfmt/oxlint, gitleaks + AWS-credential/private-key detection, semgrep (SAST), and `terraform fmt` + tflint.
- **Hook dispatcher** (`.githooks/pre-commit`) that chains the org's hook, the pre-commit framework, then the affected unit tests (`vitest run --changed`).
- **Coverage report**: `@vitest/coverage-v8` + `npm run test:coverage` (report-only, no threshold); `coverage/` gitignored.
- **Docs**: README "Testing & Code Quality" section; CONTRIBUTING hook setup.

### Setup (one-time, per contributor)
```bash
uv tool install pre-commit
git config core.hooksPath .githooks
```

### ⚠️ These hooks are a convenience, not enforcement
Client-side git hooks are **opt-in** and **bypassable** (`--no-verify`, or simply not installing them). They speed up the author's feedback loop but **cannot guarantee** checks ran. Real enforcement must live server-side — in **CI on PRs** plus **branch protection** requiring those checks to pass before merge. This PR deliberately stays inner-loop only; mirroring these checks into CI/branch protection is a recommended follow-up.

### Notes
- No global/system git config changed — only repo-local `core.hooksPath`.
- Pre-existing formatting issues in the repo were **not** auto-fixed.


svozza:
Given this project is all written in JavaScript could we use Husky as the pre-commit framework so that we don't have to take a dependency on Python in: https://typicode.github.io/husky/

Other than that the pre-commit checks you've added all look reasonable to me.

jeromevdl:
Good remark. 

We can probably mimic what we have here with husky, with a bit of custom work. `pre-commit` is quite convenient and standard.

What cannot be easily replaced is `semgrep` (also python based). Do you think I should also remove it or do you know a node alternative of it ?

And pre-commit handle the download/setup of other tools like gitleaks, which doesn't have real equivalent in nodejs ecosystem...

svozza:
We already have CodeQL enabled in the repo here which serves the same purpose so maybe we could use that?

https://docs.github.com/en/code-security/concepts/code-scanning/codeql/about-the-codeql-cli

jeromevdl:
Migrated this PR from the `pre-commit` framework to **Husky + lint-staged**. 
Going Node-native ended up surfacing a few other gaps worth addressing in the same PR — full breakdown below.

## What changed vs the original approach

| Layer | Before (pre-commit) | After (Husky) |
|---|---|---|
| Hook framework | `pre-commit` (Python, `uv tool install`) | `husky` (npm devDep) |
| Staged-file dispatch | `.pre-commit-config.yaml` | `lint-staged` (npm devDep) |
| JS/TS format + lint | oxfmt, oxlint | oxfmt, oxlint *(unchanged)* |
| Hygiene (whitespace, EOF, YAML/JSON, merge-conflict, large-file) | `pre-commit-hooks` Python utilities | dropped — oxfmt covers JS/TS hygiene; the rest are low-impact and surface elsewhere |
| Secret scanning | gitleaks (Go binary, ~150 detectors) | `secretlint` + `@secretlint/secretlint-rule-preset-recommend` (npm, ~30 detectors covering AWS / GitHub / Anthropic / OpenAI / private keys / DB strings) |
| SAST (local) | semgrep (Python) | optional — `semgrep` or `opengrep` if on PATH, silent skip otherwise |
| SAST (real) | — | **CodeQL default setup** in CI (already running on every PR — confirmed) |
| Terraform fmt + tflint | terraform, tflint binaries | unchanged, with one fix below |
| Dependency audit (SCA) | not present | **`npm audit --omit=dev --audit-level=high`** on root and frontend |
| Frontend type-check | not present | **`tsc -b`** on the frontend project when frontend TS/TSX is staged |
| Affected unit tests | `vitest run --changed` | unchanged |
| git-defender chaining | yes, in `.githooks/pre-commit` | **dropped** — not appropriate for a public sample repo, and Husky's `core.hooksPath` would override system-scope git-defender anyway |

## Why no SAST in the local hook

Tried to find a tool that is simultaneously fast, dataflow-quality, Node-native, and zero-friction. That combination doesn't exist in OSS today:

- **CodeQL CLI** — the right tool, but 1–4 minutes per run + 600 MB install. Unworkable on every commit.
- **Snyk Code** — closest match (real dataflow, npm CLI) but requires a Snyk account + `SNYK_TOKEN`. Not acceptable for an `aws-samples` public repo.
- **opengrep / semgrep OSS** — fast, but pattern-matching only (single-file, no taint). Adds a non-Node binary install.
- **eslint-plugin-security** — broken on ESLint 9, [0% recall on the 2026 benchmark](https://ofriperetz.dev/articles/benchmark-17-eslint-security-plugins-compared).

Conclusion: **CodeQL in CI is the SAST enforcement layer**. The local hook opportunistically runs `semgrep` or `opengrep` if a contributor has them on `PATH`, otherwise silent skip. CodeQL still catches everything in CI.

## Trade-offs to flag explicitly

1. **secretlint < gitleaks in detector breadth** (~30 vs ~150). Acceptable for this repo's surface; mitigated by GitHub's automatic secret scanning on public repos.
2. **No CI mirror for terraform/secret/audit checks** — local hooks are bypassable. Consistent with the original PR's "inner-loop only" scope; CI mirroring is a follow-up.
3. **94 pre-existing checkov findings** in `terraform/` (CloudWatch encryption, ALB HTTPS, log retention, security group egress, ECS read-only filesystems, etc.). Real security issues but out of scope for this PR. **Strongly recommend a dedicated hardening PR.** Documented in `CONTRIBUTING.md`.
4. **JSON/YAML/large-file hygiene checks dropped.** Low impact in practice (CI fails on broken YAML; GitHub renders merge-conflict markers visibly).

## Other things this PR fixes along the way

- **Added `terraform/.tflint.hcl`** wiring up the [tflint AWS provider plugin](https://github.com/terraform-linters/tflint-ruleset-aws) (v0.47.0). Without this, tflint was running with empty default rules — basically a no-op. The hook calls `tflint --init` opportunistically before each run; idempotent no-op when already cached.
- **Frontend coverage gaps closed.** Frontend was outside the npm workspaces and not audited or type-checked locally before. Now both are covered (silent skip if `frontend/node_modules` doesn't exist).
- **Pre-existing `terraform fmt` violation** in `modules/realtime/yjs-server/outputs.tf` fixed (one whitespace alignment).

## Onboarding

Before:
```bash
uv tool install pre-commit
git config core.hooksPath .githooks
brew install gitleaks semgrep tflint
```

After:
```bash
npm install
```

(Plus `npm --prefix frontend install` if you want the frontend audit + typecheck steps to run locally — silent skip if you don't.)

## Design principle: never block, never instruct

Every optional check (frontend, terraform, SAST, even tflint plugin init) silently skips when its tool or dependency tree isn't available. The hook never tells contributors to install anything or run setup commands. CI is the enforcement layer.

## Honest caveat (unchanged from original PR)

> Client-side git hooks are opt-in and bypassable (`--no-verify`, or simply not installing them). Real enforcement lives in CI on PRs plus branch protection. This PR deliberately stays inner-loop only.

---

Verified end-to-end: the migration commit itself ran the new pipeline cleanly (lint-staged, both audits, full vitest suite — 178/178 green in ~7s).

svozza (dismissed):

JWThewes (dismissed):

JWThewes (approved):

Issue #220, opened 2026-06-02
Split runtime libraries out of root devDependencies into dependencies


svozza:
### Description

The root `package.json` has **no `dependencies` block** — every runtime library lives in `devDependencies`, alongside the genuine build/test tooling:

- **Runtime libs (mislabelled):** `gremlin`, `gremlin-aws-sigv4`, `neptune-lambda-client`, `@aws-sdk/client-*`, `@aws-sdk/lib-dynamodb`, `@aws-sdk/credential-providers`, `@smithy/*`
- **Actual tooling (correct):** `esbuild`, `vitest`, `oxlint`, `oxfmt`, `testcontainers`, `aws-sdk-client-mock`, `aws-sdk-client-mock-vitest`

The `dependencies` vs `devDependencies` split is meant to encode runtime-vs-tooling. Right now those labels don't reflect reality.

### Why it's harmless today

Lambdas are bundled by esbuild into a single `index.mjs` (the deployed zip ships **no `node_modules`** — verified: the timeline-events artifact is one `index.mjs`, 0 dependency files). Terraform packages the `.build/` output, not an installed dependency tree, so `npm ci --omit=dev` never strips anything at deploy time. Nothing breaks.

### Why it's still worth fixing

- The taxonomy lies: a reader can't tell which libs are shipped code vs. tooling.
- It's a latent footgun — if any lambda ever moves back to the traditional `npm ci --omit=dev` + ship-`node_modules` packaging, every runtime lib would be silently stripped and the function would crash at runtime.

### Proposed change

Repo-wide, single PR:
- Add a `dependencies` block to root `package.json` and move the runtime libs into it.
- Leave only build/test tooling in `devDependencies`.
- No code or behavior change; bundles are unaffected.

### Area

Backend (Lambda) / build tooling

JWThewes:
Fixed with v2

PR #221, opened 2026-06-02
test(timeline-events): add real-gremlin tests + adopt neptune-lambda-client

svozza:
Applies the real-gremlin testing pattern (#181, #202, #203) to `lambda/timeline-events`, and swaps the hand-rolled connection/SigV4 code for [`neptune-lambda-client`](https://github.com/svozza/neptune-lambda-client). Refs #41.

## Changes

- **CJS → ESM**; runtime deps stripped from the lambda's `package.json` (hoisted at root), package renamed to the workspace convention, added to root `workspaces`.
- **Terraform**: module switched to the `npm run build -w timeline-events` esbuild pattern, `nodejs18.x` → `nodejs24.x` (matches sprint-graph / purge-neptune).
- **Neptune access via `neptune-lambda-client`**: the client is built once at module scope and reused across warm invocations (lazy connect, reconnect on socket drop, retry on transient errors) instead of connecting/closing per request. `useIam` is tied to `GREMLIN_PROTOCOL` so the testcontainer (plain `ws://`) skips signing while production stays `wss://` + SigV4. `close` is exported for test teardown only.

## Tests

Nine tests against a shared `tinkerpop/gremlin-server` testcontainer (partition-isolated), covering:
- `OPTIONS` short-circuit
- empty list for a sprint with no events
- timestamp-descending ordering
- `valueMap` → camelCase field mapping
- sprint isolation (no cross-sprint leakage)
- `POST` create + `HAS_TIMELINE_EVENT` edge wiring
- field/timestamp defaults
- no-op `POST` against a missing sprint
- `405` for unsupported methods

## Verification

- `npm run test -w timeline-events` → 9 passing.
- `lint` / `format:check` clean.
- Deployed to dev (`eu-west-1`) and smoke-tested against real Neptune: `GET` → `200` with `[]`, confirming the library's SigV4 path works end-to-end.

## Follow-up

Root `package.json` keeps `neptune-lambda-client` in `devDependencies` to match the existing convention (all runtime libs live there because esbuild bundles them). Splitting runtime libs into a proper `dependencies` block is tracked separately in #220.


lambda/timeline-events/index.js
```diff
@@ -57,24 +68,26 @@ exports.handler = async (event) => {
...
-          .property('sprint_id', sprintId)
-          .as('e')
-          .addE('HAS_TIMELINE_EVENT')
-          .from_('s')
-          .to('e')
-          .next();
+        await query((g) =>
+          g
+            .V()
+            .has('Sprint', 'id', sprintId)
```
jeromevdl, line 74: existing problem in main, but maybe we should fix it here: if the sprint doesn't exists return an error:

```
const exists = await query((g) => g.V().has('Sprint', 'id', sprintId).hasNext());
if (!exists) return res(404, { error: 'Sprint not found' });
```

The test at line 197 ("does not attach the event when the sprint does not exist") locks this in and validates the bug as a normal behaviour: expect(res.statusCode).toBe(201) + expect(fetched.body).toEqual([]).

lambda/timeline-events/index.js
```diff
@@ -27,28 +39,27 @@ const mapEvent = (v) => ({
...
-          .out('HAS_TIMELINE_EVENT')
-          .hasLabel('TimelineEvent')
-          .order()
-          .by('timestamp', order.desc)
-          .valueMap()
-          .toList();
+        const list = await query((g) =>
+          g
+            .V()
+            .has('Sprint', 'id', sprintId)
```
jeromevdl, line 55: existing problem in main, but maybe we should fix it here: if the sprint doesn't exists return an error:

```
const exists = await query((g) => g.V().has('Sprint', 'id', sprintId).hasNext());
if (!exists) return res(404, { error: 'Sprint not found' });
```

lambda/timeline-events/index.js
jeromevdl, line 67: No input validation on POST, json parse does not validate the input, just the format and will simply fail as a 500 (should be 400)

lambda/timeline-events/test/timeline-events.test.js
jeromevdl, line 1: - No test for the case where body is missing or malformed JSON on POST
- No test for type/title missing

lambda/timeline-events/index.js
svozza: We should be using a validation library like Zod for this sort of validation but this is beyond the scope of this PR.

lambda/timeline-events/index.js
svozza: Beyond the scope of this PR. I will create a separate issue for this.

lambda/timeline-events/index.js
svozza: Beyond the scope of this PR, should be a separate issue.

lambda/timeline-events/test/timeline-events.test.js
svozza: Again, these are due to lack of input validation, this should be done in a separate PR.

lambda/timeline-events/index.js
svozza: Filed as #223 to track the sprint-existence check for both the POST and GET handlers. I've also removed the test that locked in the current behaviour (`does not attach the event when the sprint does not exist`) in this PR.

lambda/timeline-events/index.js
svozza: Filed as #224 to track input validation on POST (return 400 for missing/malformed body and missing `type`/`title`, via a validation library such as Zod).

lambda/timeline-events/index.js
jeromevdl: Makes sense

lambda/timeline-events/index.js
jeromevdl: OK

lambda/timeline-events/test/timeline-events.test.js
jeromevdl: OK, see #224

jeromevdl (approved):

JWThewes (approved):

Issue #219, opened 2026-06-02
Agent can hang during Kiro session startup after fresh deployment


JWThewes:
## Summary
On some fresh deployments, starting an agent can appear to stall before any agent output is produced. The observed log sequence reaches Kiro ACP initialization and starts session creation, then stops before `Session created` / `Sending prompt`.

## Observed logs
```text
[acp] Initialized: Kiro CLI Agent 2.5.0
[acp] Creating session...
[acp] Starting session with 2 MCP server(s)
```

A later healthy run showed `Starting session with 1 MCP server(s)`, suggesting the stall may be related to an extra MCP server loaded from SSM during `session/new`.

## Problems
- ACP JSON-RPC requests did not have timeouts, so `session/new` could wait indefinitely.
- Logs did not include MCP server names/commands, making the extra server hard to identify.
- Final status logging collapsed DynamoDB and Neptune failures into generic `Failed to save status` messages.
- Neptune SigV4 signing used AWS SDK v3 credential objects directly, which is fragile with signer packages expecting older key aliases.

## Proposed fix
- Add bounded timeouts for ACP initialization/session creation/mode switching.
- Keep `session/prompt` unbounded because agent prompts can legitimately run for a long time.
- Log MCP server names/commands before `session/new`.
- Validate extra MCP server SSM config is a JSON array.
- Split status finalizer errors into DynamoDB vs Neptune messages.
- Normalize AWS SDK v3 credentials before Neptune SigV4 signing.

## Branch
`fix/agent-session-new-timeout-logging`

JWThewes:
This is the whole agent log that's available:

June 2, 2026, 13:40
	
[pool-worker] Job done (exit=0, pushSucceeded=false)
	
agent
June 2, 2026, 13:40
	
[pool-worker] Failed to update AgentRun status: Access key and secret key are required
	
agent
June 2, 2026, 13:40
	
Failed to save status: Access key and secret key are required
	
agent
June 2, 2026, 13:40
	
[acp] kiro-cli exited with code 0
	
agent
June 2, 2026, 13:40
	
[acp] << response id=0 ok
	
agent
June 2, 2026, 13:40
	
[acp] Initialized: Kiro CLI Agent 2.5.0
	
agent
June 2, 2026, 13:40
	
[acp] Creating session...
	
agent
June 2, 2026, 13:40
	
[acp] Starting session with 2 MCP server(s)
	
agent
June 2, 2026, 13:40

[...]

PR #227, opened 2026-06-03
Fix agent startup hangs from broken MCP servers

JWThewes:
## Summary

- Add bounded ACP request timeouts for startup/session setup while leaving long-running prompts unbounded.
- Validate and preflight merged MCP server config so optional stdio MCP servers with missing commands are skipped instead of hanging startup.
- Surface friendly agent startup/execution failures through WebSocket events, persisted AgentOutputs, and existing UI stream panels.
- Harden diagnostics by redacting MCP args/URL secrets and avoiding raw error broadcasts.
- Normalize Neptune signer credentials in agent-side graph paths.

## Issue

Refs #219

Broken MCP server configuration could make agents appear to hang during startup, especially when a server required an executable such as `uv` that was not present in the worker image, or when ACP `session/new` waited indefinitely for MCP startup. The previous behavior left the failure mostly in logs and did not give users an actionable UI message.

## Testing

- `npm test` - 19 files, 393 tests passed
- `npm run format:check` - passed
- `npm run lint` - no errors, existing warnings only

### DIFF
--- a/frontend/src/components/settings/McpServersSection.tsx
+++ b/frontend/src/components/settings/McpServersSection.tsx
@@ -85,6 +85,13 @@ export function McpServersSection({
       </CardHeader>
       <CardContent className="space-y-3">
         <p className="text-xs text-muted-foreground">{description}</p>
+        {canEdit && (
+          <p className="text-[11px] text-muted-foreground">
+            Stdio MCP servers can use these agent image commands: <code>node</code>,{' '}
+            <code>npx</code>, <code>uv</code>, <code>uvx</code>, <code>python</code>,{' '}
+            <code>python3</code>. Use an absolute path for any other executable.
+          </p>
+        )}
         <Textarea
           value={value}
           onChange={(e) => {
@@ -97,7 +104,7 @@ export function McpServersSection({
           )}
           spellCheck={false}
           disabled={!canEdit}
-          placeholder='[{"name":"my-tool","command":"npx","args":["-y","my-mcp-server"]}]'
+          placeholder='[{"name":"my-tool","command":"npx","args":["-y","my-mcp-server"],"env":[]}]'
         />
         {error && (
           <p className="text-[11px] text-destructive flex items-start gap-1">
--- a/lambda/agents-ecs/Dockerfile
+++ b/lambda/agents-ecs/Dockerfile
@@ -46,28 +46,29 @@ RUN curl -LsSf https://astral.sh/uv/install.sh | sh \
     && chmod -R a+rX "$UV_PYTHON_INSTALL_DIR"
 
 # Embed AI-DLC workflow rules
-COPY aidlc-rules /opt/aidlc-rules
+COPY agents-ecs/aidlc-rules /opt/aidlc-rules
 
 # Install MCP server
-COPY mcp-server-graph/package.json mcp-server-graph/index.js /opt/mcp-server-graph/
+COPY agents-ecs/mcp-server-graph/package.json agents-ecs/mcp-server-graph/index.js /opt/mcp-server-graph/
 RUN cd /opt/mcp-server-graph && npm install --production
 
 # Install ACP client (includes drivers)
-COPY package.json /opt/acp-client/package.json
+COPY agents-ecs/package.json /opt/acp-client/package.json
 RUN cd /opt/acp-client && npm install --production
 
 ARG IMAGE_TAG=latest
 ENV IMAGE_TAG=${IMAGE_TAG}
 
-COPY acp-client.js pool-worker.js branch-cleanup.js construction-orchestrator-prompt.js /opt/acp-client/
-COPY drivers/ /opt/acp-client/drivers/
+COPY agents-ecs/acp-client.js agents-ecs/pool-worker.js agents-ecs/branch-cleanup.js agents-ecs/construction-orchestrator-prompt.js /opt/acp-client/
+COPY agents-ecs/drivers/ /opt/acp-client/drivers/
+COPY shared/mcp-validator.js /opt/shared/mcp-validator.js
 
 RUN mkdir -p /workspace \
     && chown -R node:node /workspace /opt/aidlc-rules /opt/mcp-server-graph /opt/acp-client /opt/kiro
 
 WORKDIR /workspace
 
-COPY --chown=node:node entrypoint.sh /entrypoint.sh
+COPY --chown=node:node agents-ecs/entrypoint.sh /entrypoint.sh
 RUN chmod +x /entrypoint.sh
 
 USER node
--- a/lambda/agents-ecs/acp-client.js
+++ b/lambda/agents-ecs/acp-client.js
@@ -20,6 +20,7 @@ const { getDriver } = require('./drivers');
 const AGENT_CLI = process.env.AGENT_CLI || 'kiro';
 const driver = getDriver(AGENT_CLI);
 const ACP_VERBOSE = process.env.ACP_VERBOSE === 'true';
+const DEFAULT_REQUEST_TIMEOUT_MS = parseInt(process.env.ACP_REQUEST_TIMEOUT_MS || '120000', 10);
 
 // ---------------------------------------------------------------------------
 // MCP servers — merged from global (SSM), project (Neptune), task (Neptune)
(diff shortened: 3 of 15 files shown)


lambda/agents-ecs/acp-client.js
```diff
@@ -176,13 +321,86 @@ function send(method, params) {
...
+        ? setTimeout(() => {
+            pending.delete(id);
+            reject(new Error(`${method} timed out after ${timeoutMs}ms`));
+          }, timeoutMs)
+        : null;
+    pending.set(id, { method, resolve, reject, timer });
   });
 }
 
+function settlePending(id, settle, value) {
```
svozza, line 306: For my own understanding, what's going on here, loks like we're reimplementting promise functions

lambda/agents-ecs/acp-client.js
```diff
@@ -20,13 +21,132 @@ const { getDriver } = require('./drivers');
...
+    env: Array.isArray(server.env)
+      ? server.env.map((e) => ({ name: e.name, value: `<${(e.value || '').length} chars>` }))
+      : server.env,
+    headers: Array.isArray(server.headers)
+      ? server.headers.map((h) => ({ name: h.name, value: `<${(h.value || '').length} chars>` }))
+      : server.headers,
+  };
+}
+
+function validateMcpServer(server, index) {
```
jeromevdl: there is a `mcp-validator.js` in lambda/shared that already validates during input, why testing it again here (and not using the mcp-validator)?

lambda/agents-ecs/acp-client.js
```diff
@@ -20,13 +21,132 @@ const { getDriver } = require('./drivers');
...
+    }
+    if (server.headers === undefined) server.headers = [];
+    if (!Array.isArray(server.headers)) {
+      throw new Error(`MCP server ${server.name} headers must be an array`);
+    }
+  }
+  return server;
+}
+
+function commandExists(command) {
```
jeromevdl, line 67: should we also test this in the mcp-validator when adding the mcp, instead of waiting that an agent is run? So that we can directly tell, "your mcp won't be able to run (command not found)", rather than accepting blindly and postponing the error?

lambda/agents-ecs/acp-client.js
```diff
@@ -176,13 +321,86 @@ function send(method, params) {
...
+
+function friendlyAgentError(err) {
+  const raw = err && err.message ? err.message : String(err || 'Unknown error');
+  if (/command not found on PATH/i.test(raw)) {
+    const match = raw.match(/MCP server ([^:]+:[^ ]+) command not found on PATH: (.+)$/i);
+    if (match) {
+      return `An MCP server could not start because the required command \`${match[2]}\` is not installed in the agent image. The server was ${match[1]}.`;
+    }
+    return 'An MCP server could not start because one of its required commands is not installed in the agent image.';
+  }
```
jeromevdl, line 331: I think we should defer this to the configuration of the MCP, to make sure we can run it while the user is configuring it, rather than waiting later.

lambda/agents-ecs/acp-client.js
jeromevdl: but we don't have the image in the lambda...:( 
Maybe we can at least put a small list of binaries we now for sure they are available (node, npx, uvx, uv, python3, python, ...) to reduce the risk of typos. And we keep this check here as a real check.

lambda/agents-ecs/acp-client.js
```diff
@@ -20,13 +21,132 @@ const { getDriver } = require('./drivers');
...
+      throw new Error(message);
+    }
+    reportAgentWarning(
+      `Skipping optional ${message}. The agent will continue without this tool server.`,
+    );
+  }
+  return runnable;
+}
+
+function parseMcpServersJson(raw, source) {
```
jeromevdl, line 107: ± duplicate of mcp-validator `validateMcpServersJson`

jeromevdl: worth adding it in the dockerfile ? already shared by lambda functions

lambda/agents-ecs/acp-client.js
JWThewes: Not reimplementing the Promise — these are bookkeeping helpers around the `pending` map. Each entry now carries `{method, resolve, reject, timer}`, and we need three things to happen atomically when a response arrives: delete from the map, clear the timeout, and settle. Without the helper every callsite (response handler, error handler, exit handler) would have to remember to clear the timer, and a missed `clearTimeout` would later fire `reject(timeout)` on an already-resolved request and leak the timer until process exit. Pre-timeout there was nothing to clean up, so this didn't exist before.

JWThewes:
@jeromevdl @svozza I addressed your comments. Could you re-review this one?

lambda/agents-ecs/Dockerfile
```diff
@@ -46,28 +46,29 @@ RUN curl -LsSf https://astral.sh/uv/install.sh | sh \
     && chmod -R a+rX "$UV_PYTHON_INSTALL_DIR"
 
 # Embed AI-DLC workflow rules
-COPY aidlc-rules /opt/aidlc-rules
+COPY agents-ecs/aidlc-rules /opt/aidlc-rules
```
jeromevdl, line 49: we are in agents-ecs/ folder ... ?

lambda/shared/mcp-validator.js
```diff
@@ -15,6 +15,7 @@ const ALLOWED_TYPES = new Set(['stdio', 'http', 'sse']);
 const STDIO_ALLOWED_KEYS = new Set(['type', 'name', 'command', 'args', 'env']);
 const HTTP_ALLOWED_KEYS = new Set(['type', 'name', 'url', 'headers']);
 const NAME_VALUE_KEYS = new Set(['name', 'value']);
+const KNOWN_AGENT_IMAGE_COMMANDS = new Set(['node', 'npx', 'uv', 'uvx', 'python', 'python3']);
```
jeromevdl: KNOWN_AGENT_IMAGE_MCP_COMMANDS

frontend/src/components/settings/McpServersSection.tsx
```diff
@@ -85,6 +85,13 @@ export function McpServersSection({
       </CardHeader>
       <CardContent className="space-y-3">
         <p className="text-xs text-muted-foreground">{description}</p>
+        {canEdit && (
+          <p className="text-[11px] text-muted-foreground">
+            Stdio MCP servers can use these agent image commands: <code>node</code>,{' '}
+            <code>npx</code>, <code>uv</code>, <code>uvx</code>, <code>python</code>,{' '}
+            <code>python3</code>. Use an absolute path for any other executable.
```
jeromevdl, line 92: do we really allow to use something with "/..." ?

jeromevdl:
@JWThewes looks good, just a few comments left

lambda/shared/mcp-validator.js
JWThewes: Done — renamed to `KNOWN_AGENT_IMAGE_MCP_COMMANDS` in 69267b6.

lambda/agents-ecs/Dockerfile
JWThewes: The Docker build context is `lambda/`, not `agents-ecs/` — Terraform sets `source_path = .../lambda` and `docker_file_path = .../lambda/agents-ecs/Dockerfile` (terraform/modules/compute/agents/main.tf:86-87). The context is the parent `lambda/` dir specifically so the image can also `COPY shared/mcp-validator.js` from the sibling `shared/` package. So all source paths are relative to `lambda/`, which is why they need the `agents-ecs/` prefix. I verified all 11 COPY sources resolve correctly against the `lambda/` context.

frontend/src/components/settings/McpServersSection.tsx
JWThewes: Yes, intentionally. The validator accepts a bare command only if it's one of the known image commands (node, npx, uv, uvx, python, python3); anything else must be an absolute path (must contain `/`). This is the escape hatch for custom/mounted binaries at a known path inside the agent image. The MCP config is admin-gated and the command runs inside the sandboxed agent container, so allowing an explicit absolute path doesn't widen the trust boundary — it just lets advanced users point at a binary we don't ship by name. Happy to tighten to an allowlist-only if you'd prefer, but that would block legitimate custom-binary setups.

JWThewes:
@jeromevdl thanks — addressed the latest round:

- **mcp-validator.js**: renamed `KNOWN_AGENT_IMAGE_COMMANDS` → `KNOWN_AGENT_IMAGE_MCP_COMMANDS` (69267b6).
- **Dockerfile `agents-ecs/` prefix**: kept as-is — the build context is `lambda/` (so the image can also COPY `shared/mcp-validator.js`), so the prefix is required. Verified all COPY paths resolve. Details in the inline reply.
- **UI 'absolute path' hint**: intentional escape hatch for custom binaries; bare commands are still allowlist-restricted. Rationale in the inline reply — let me know if you'd rather lock it to allowlist-only.

frontend/src/components/settings/McpServersSection.tsx
jeromevdl: No that's ok. it's a sandboxed environment and if it's not an MCP (system command) it won't work. OK

jeromevdl (approved):
LGTM

PR #236, opened 2026-06-04
feat(review): surface all PRs in multi-repo review with per-PR tabs

eipasteur:
Stacked on #183 (multi-repo base) — review after #183.

Surfaces every `PullRequest` from the sprint graph (one per repo) on the review page: per-repo tabs with state dots, a labeled View PR button, and a scoped checkout command. Sprint actions and the review verdict sit below.

- `sprintGraph.extractPrs()` + `PrInfo` type (filters stale, sorts by repo, carries `pr_state`); unit tests
- `ReviewPage`: per-PR selection scopes View/checkout/comments; cancel-guarded comments fetch; NaN risk guard; `PARTIAL` verdict styling
- `PipelineView`: sidebar PR button navigates to the review page
- `reviews`/`ReviewEditor`: `PARTIAL` added to `ReviewStatus`

### DIFF
--- a/frontend/src/pages/ReviewPage.tsx
+++ b/frontend/src/pages/ReviewPage.tsx
@@ -8,7 +8,7 @@ import { projectsService, type Project } from '@/services/projects';
 import { reviewsService } from '@/services/reviews';
 import { questionsService } from '@/services/questions';
 import { githubService, type PRComment } from '@/services/github';
-import { sprintGraphService } from '@/services/sprintGraph';
+import { sprintGraphService, extractPrs, type PrInfo } from '@/services/sprintGraph';
 import { realtimeService } from '@/services/realtime';
 import { sprintsService } from '@/services/sprints';
 import { agentsService } from '@/services/agents';
@@ -48,6 +48,7 @@ import {
   XCircle,
   Code2,
   GitBranch,
+  GitPullRequest,
   Link,
   AlertTriangle,
   ShieldAlert,
@@ -144,6 +145,8 @@ export default function ReviewPage() {
   } = useSprint();
 
   const [project, setProject] = useState<Project | null>(null);
+  const [prs, setPrs] = useState<PrInfo[]>([]);
+  const [selectedPrId, setSelectedPrId] = useState<string>('');
   const [prComments, setPrComments] = useState<PRComment[]>([]);
   const [prBranch, setPrBranch] = useState('');
   const [prBaseBranch, setPrBaseBranch] = useState('main');
@@ -199,13 +202,13 @@ export default function ReviewPage() {
       sprintGraphService
         .get(sprintId)
         .then((graph) => {
-          const prNode = graph.nodes.find((n) => n.type === 'PullRequest');
-          if (prNode) {
-            setPrBranch((prNode as Record<string, string>).branch || '');
-            setPrBaseBranch((prNode as Record<string, string>).base_branch || 'main');
-          }
+          const found = extractPrs(graph);
+          setPrs(found);
+          setSelectedPrId((prev) =>
+            prev && found.some((p) => p.id === prev) ? prev : (found[0]?.id ?? ''),
+          );
           // Fall back to sprint branch if no PR node yet
-          else if (sprint?.branch) {
+          if (found.length === 0 && sprint?.branch) {
             setPrBranch(sprint.branch);
             setPrBaseBranch(sprint.baseBranch || 'main');
           }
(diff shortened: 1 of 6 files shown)


frontend/src/pages/ReviewPage.tsx
```diff
@@ -376,7 +422,11 @@ export default function ReviewPage() {
                   onAutoSave={async (draft) => {
                     questionsService
                       .update(sprintId, pq.id, { draftAnswer: draft })
-                      .catch(() => {});
+                      .catch((err) => {
```
JWThewes, line 436: The catch is on a question draft autosave — it has nothing to do with the sprint graph or PRs. The comment and console.error label both look copy-pasted from the graph fetch handler. A
  graph fetch failure here would log "Failed to update question draft" — fix the message, or move the verbose handler to the actual graph .catch(() => {}) at line ~218 (which is still
  silent, contradicting the comment's stated motivation).

frontend/src/services/sprintGraph.ts
```diff
@@ -21,3 +21,37 @@ export interface SprintGraph {
...
+  prUrl: string;
+  prNumber: string;
+  repository: string; // owner/repo
+  branch: string;
+  baseBranch: string;
+  state: string; // open | closed | merged | ''
+}
+
+// Collect every PullRequest node from a sprint graph (multi-repo = one per repo).
+// Stale PRs (superseded/merged) are filtered out; result is sorted by repository
```
JWThewes: It's only stale PR's are filtered out - merged PRs are not filtered

frontend/src/pages/ReviewPage.tsx
```diff
@@ -214,16 +217,59 @@ export default function ReviewPage() {
     }
   }, [sprintId, sprint?.branch, sprint?.baseBranch]);
```
JWThewes: The deps [sprintId, sprint?.branch, sprint?.baseBranch] cause a full graph fetch any time those sprint fields change (realtime updates, etc.). The branch fallback is a one-shot init
  concern and doesn't justify refetching the graph. Consider splitting: fetch graph on sprintId only, and keep the fallback assignment in the same effect guarded by prs.length === 0.

frontend/src/pages/ReviewPage.tsx
```diff
@@ -423,13 +473,64 @@ export default function ReviewPage() {
...
+                  size="sm"
+                  className="gap-1.5 ml-auto"
+                  title={activeRepo ? `Open ${activeRepo} #${activePrNumber}` : 'Open pull request'}
+                  onClick={() => window.open(activePrUrl, '_blank')}
+                >
+                  <ExternalLink className="h-3.5 w-3.5" /> {viewPrLabel}
+                </Button>
+              </div>
+              {prs.length > 1 && (
+                <Tabs value={selectedPrId} onValueChange={setSelectedPrId}>
```
JWThewes: Would a ToggleGroup make more sense here?
https://www.radix-ui.com/primitives/docs/components/toggle-group

JWThewes (changes requested):

frontend/src/pages/ReviewPage.tsx
eipasteur: Fixed. The autosave catch now logs "Failed to autosave question draft" (its actual concern), and the verbose handler moved to the real graph .catch at the fetch effect, which is no longer silent.

frontend/src/services/sprintGraph.ts
eipasteur: You're right, the comment was misleading. The backend sets stale=true *and* pr_state='merged' when a PR is merged on GitHub, so the old !stale filter dropped them and the violet "merged" dot could never render. Filter now keeps merged PRs (only superseded stale PRs are dropped) and the comment is corrected. Added a unit test covering the merged case.

frontend/src/pages/ReviewPage.tsx
eipasteur: Done. Split into two effects: the graph fetch now depends on [sprintId] only, and the sprint-branch fallback lives in its own effect guarded by prs.length === 0, so realtime branch/baseBranch updates no longer trigger a full graph refetch.

frontend/src/pages/ReviewPage.tsx
eipasteur: I agree - switched to ToggleGroup (added the shadcn wrapper; @radix-ui/react-toggle-group was already a dep). It's a single-select value picker, which is semantically correct here vs Tabs (no content panels).

JWThewes (approved):

PR #215, opened 2026-06-01
feat: improved code reviews with per-agent verdicts and human approval workflow

jeromevdl:
## Improved code reviews

Splits the review into per-agent verdicts (technical + business), the fix uses the comments in PR, and adds an explicit human approval workflow that completes the sprint.

_Note: This was a PR on the internal gitlab which has not been merged before moving to GitHub_

### What's new

**Per-agent reviews**
- "Blind Review" → **Technical Review** (code-only, technical)
- "Full Review" → **Business Review** (cross-references requirements)
- Each agent has its own status badge (PASS/FAIL/PARTIAL) and risk score (0-10) displayed in its tab
- New `Review` node fields: `blind_status`, `blind_risk_score`, `blind_risk_reasoning`, `full_status`, `full_risk_score`, `full_risk_reasoning`
- Backwards compatible: existing reviews fall back to the legacy `risk_score` / `risk_reasoning` fields

**`review-modify` agent reads PR comments**
- New `get_pr_comments` MCP tool
- Agent reads all PR feedback, categorizes it (clear / ambiguous / trivial), fixes what's clear, asks `ask_question` on ambiguities, posts a `🔧 Review Modify Summary` PR comment
- Modify dialog renamed to "Fix Review Findings"; instruction is now optional

**Human review → sprint completion**
- Pass / Partial / Fail buttons (immediate save, like before, plus the new Partial)
- New "Send to GitHub" button — posts the verdict + comments to the PR
- Sending a **Pass** verdict transitions the sprint to a new `COMPLETED` phase
- After COMPLETED: textarea, verdict buttons, "Send to GitHub", "Kick-Off Review Agents", and "Fix Review Findings" are all disabled (sprint is read-only)

**Misc**
- Pool-worker: strip `task-` prefix when constructing task branch names (bugfix)
- Agent broadcasts now include `executionId` and `agentType`
- Dark-mode prose contrast on agent output cards and inception general info
- Construction page button: "Approve & Move to Review" → "End Construction & Move to Review"
- Removed unused `SprintNav.tsx`

### Schema changes (Neptune)

Additive only:
- `Sprint.phase` enum extended with `COMPLETED`
- `Review` node: 6 new properties (per-agent status + risk fields), `status` enum extended with `PARTIAL`
- No data migration needed — `mapReview` falls back to the legacy `risk_score` / `risk_reasoning` for old reviews

### Deploy notes

Both frontend and backend need to be deployed:
[...]

PR #237, opened 2026-06-04
fix(agents): harden review stale-marking, PR pr_state, validation & PARTIAL

eipasteur:
Stacked on #236 (review) → #183 (multi-repo base). **Draft — pending rebase onto current `main` + reconciliation with #215/#205/#226** (overlaps in pool-worker / mcp-server-graph / reviews).

- agents: define missing `__` (gremlin.process.statics) so review nodes are marked stale on construction re-run
- mcp-server-graph: `pr_state='open'` at PR creation (multi & single repo); clear sprint pr fields on stale; `parseOwnerRepo` validation; whitelist `Review.status`
- reviews: accept PARTIAL; deterministic order-by
- create-pr: robust 422 existing-PR lookup + owner/repo validation
- pool-worker: per-repo coverage in full_review

### DIFF
--- a/lambda/create-pr/create-pr.js
+++ b/lambda/create-pr/create-pr.js
@@ -151,7 +151,11 @@ exports.handler = async (event) => {
 
   try {
     // Get project details from Neptune via GitHub Lambda
-    const [owner, repo] = gitRepo.split('/');
+    const parts = gitRepo.split('/');
+    if (parts.length !== 2 || !parts[0] || !parts[1]) {
+      return { statusCode: 400, body: `Invalid gitRepo "${gitRepo}": expected "owner/repo"` };
+    }
+    const [owner, repo] = parts;
 
     // Create PR using GitHub API
     const prTitle = `AI-DLC: ${branch}`;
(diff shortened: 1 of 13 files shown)


lambda/reviews/index.js
```diff
@@ -48,12 +55,24 @@ exports.handler = async (event) => {
 
     switch (httpMethod) {
       case 'GET': {
-        // Return the active (non-stale) review. If all are stale, return the most recent one.
+        // Return the active (non-stale) review. If all are stale, return the most
```
JWThewes, line 58: The existing comment says "If all are stale, return the most recent one" and the existing fallback is allReviews[allReviews.length - 1]. With the new Order.desc ordering on stale_at, the
  newest stale review is now at index 0 and the oldest at length - 1 — so the fallback now deterministically returns the oldest stale review, contradicting the comment. Pre-PR, ordering
  was unspecified so [length-1] was arbitrary; this PR makes the bug deterministic. Fix: use allReviews[0] (or change to Order.asc).

lambda/create-pr/create-pr.js
JWThewes, line 1: We should add some tests for findByBranch (especially the fork fallback) and the new owner/repo validation

JWThewes (changes requested):
- SHELL_SAFE_REPO_PATTERN is now defined in two places (lambda/projects/index.js:147 and lambda/agents/index.js:54); the agents copy adds length <= 200, the projects copy doesn't. Drift
  waiting to happen — pull into a shared util.
  - parseOwnerRepo in mcp-server-graph and the inline check in create-pr.js:154-157 do the same thing — could be one helper.
  - VALID_REVIEW_STATUSES in mcp-server-graph hardcodes the same array as VALID_STATUSES in reviews/index.js; comment acknowledges "Mirrors VALID_STATUSES" but doesn't share it.
- GIT_AUTHOR_EMAIL/GIT_AUTHOR_NAME overrides aren't wired in Terraform. The comment says "set by the task def" but grep -rn GIT_AUTHOR terraform/ returns nothing. Either land the wiring
  with this change or drop the override until it's needed

lambda/reviews/index.js
eipasteur: Fixed. With Order.desc the newest stale review is at index 0, so the fallback now uses allReviews[0] instead of allReviews[length-1].

lambda/create-pr/create-pr.js
eipasteur: Added tests for findByBranch (precise head-filter path, fork/org-mismatch fallback-to-list, neither-found→500) and for gitRepo validation (missing slash, extra segments, empty string). All 10 tests pass.

eipasteur:
- Extracted to lambda/shared/repo-validation.js — canonical source for both lambdas. agents uses ./shared/repo-validation (runtime zip injection); projects uses ../shared/repo-validation (esbuild bundles it). Hardened projects to match agents' 200-char cap.

- Wired in the ECS task definition with generic defaults ("AI-DLC Agent" / "ai-dlc@example.com"). Variable chain: modules/compute/agents/variables.tf → main.tf module call → terraform/variables.tf. Overridable per deployment via -var. Makes the pool-worker comment accurate.

JWThewes (approved):

PR #238, opened 2026-06-04
chore(infra): multi-repo API routes + frontend deploy/env guards

eipasteur:
Stacked on the agents PR → #236 → #183. **Draft — pending rebase onto current `main`** (overlaps in terraform/api routes).

- API Gateway routes for project repo management + required outputs/variables
- guard the frontend deploy/env-generation scripts (env discovery + init-env safety)

### DIFF
--- a/scripts/deploy-frontend.sh
+++ b/scripts/deploy-frontend.sh
@@ -22,6 +22,18 @@ echo "Deploying frontend for environment: $ENVIRONMENT"
 
 # Get S3 bucket name from Terraform output
 cd "$TF_DIR"
+
+# Safety: the initialized backend MUST match the requested environment before we
+# read the bucket/distribution — otherwise the `s3 sync --delete` below could
+# wipe the wrong environment's bucket.
+INIT_ENV=$(terraform output -raw environment 2>/dev/null || echo "")
+if [[ -n "$INIT_ENV" && "$INIT_ENV" != "$ENVIRONMENT" ]]; then
+    echo "Error: terraform is initialized for '$INIT_ENV' but you requested '$ENVIRONMENT'."
+    echo "Re-init the correct backend first:"
+    echo "  terraform init -backend-config=environments/$ENVIRONMENT.s3.tfbackend -reconfigure"
+    exit 1
+fi
+
 BUCKET_NAME=$(terraform output -raw s3_bucket_name 2>/dev/null || echo "")
 CLOUDFRONT_ID=$(terraform output -raw cloudfront_distribution_id 2>/dev/null || echo "")
 
--- a/scripts/deploy-frontend.sh
+++ b/scripts/deploy-frontend.sh
@@ -6,8 +6,13 @@ ENVIRONMENT=${1:-dev}
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 TF_DIR="$SCRIPT_DIR/../terraform"
 
-# Discover available environments from terraform/environments/*.tfvars
-AVAILABLE_ENVS=$(ls "$TF_DIR/environments/"*.tfvars 2>/dev/null | xargs -n1 basename | sed 's/\.tfvars$//' | grep -v '\.example$' | tr '\n' ' ')
+# Discover available environments from terraform/environments/*.tfvars.
+# The *.tfvars glob never matches *.tfvars.example, so no grep -v needed.
+# Fall back to "dev prod" so a fresh public clone (no .tfvars yet) still works.
+AVAILABLE_ENVS=$(ls "$TF_DIR/environments/"*.tfvars 2>/dev/null | xargs -n1 basename | sed 's/\.tfvars$//' | tr '\n' ' ')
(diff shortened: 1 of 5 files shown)


terraform/main.tf
```diff
@@ -132,7 +132,11 @@ module "dynamodb" {
 module "neptune" {
   source = "./modules/data/neptune"
 
-  name_prefix        = "${var.project_name}-${var.environment}"
+  # State-compatible override: the staging cluster was created with a "-dev"
+  # prefix. Without this the default "${project}-${environment}" would rename
+  # the cluster identifier and force-replace Neptune (data loss). See
+  # .kiro/plans/staging-migration.md.
```
jeromevdl: `.kiro/plans/staging-migration.md` ??

terraform/modules/api/main.tf
jeromevdl, line 1: probably we should have separated the 2 topics in 2 PRs...

terraform/modules/api/routes.tf
jeromevdl, line 1: Was it missing from the multi-repo PR ? How could that work ? Why not adding it to #183 instead (it's not closed)

terraform/main.tf
eipasteur: Removed it

terraform/modules/api/main.tf
eipasteur: Moved the multi-repo API routes out to #183, so this PR is now just the deploy/env script guards.

terraform/modules/api/routes.tf
eipasteur: Moved them into #183.

scripts/generate-env.sh
JWThewes, line 1: Only generate-env.sh was switched to discover envs from *.tfvars. deploy-frontend.sh:9-12, deploy-terraform.sh:8-10, and bootstrap.sh:16-18
  still hardcode dev|prod. So adding staging.tfvars lets you generate-env.sh staging but ./deploy-frontend.sh staging still bails at the hardcoded check. Worse, the new comment in
  generate-env.sh says:

  ▎ # (keep in sync with deploy-frontend.sh — do not hardcode an env allow-list).

  …which is actively misleading because deploy-frontend.sh is still hardcoded.

scripts/deploy-frontend.sh
```diff
@@ -22,6 +22,18 @@ echo "Deploying frontend for environment: $ENVIRONMENT"
 
 # Get S3 bucket name from Terraform output
 cd "$TF_DIR"
+
```
JWThewes, line 30: We have this check two times now.
As deploy-frontend is calling generate-env.sh shouldn't it enough to have it just there?

scripts/generate-env.sh
```diff
@@ -6,13 +6,29 @@ ENVIRONMENT=${1:-dev}
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 TF_DIR="$SCRIPT_DIR/../terraform"
 
-if [[ "$ENVIRONMENT" != "dev" && "$ENVIRONMENT" != "prod" ]]; then
-    echo "Usage: $0 [dev|prod]"
+# Discover available environments from terraform/environments/*.tfvars
+# (keep in sync with deploy-frontend.sh — do not hardcode an env allow-list).
+AVAILABLE_ENVS=$(ls "$TF_DIR/environments/"*.tfvars 2>/dev/null | xargs -n1 basename | sed 's/\.tfvars$//' | grep -v '\.example$' | tr '\n' ' ')
```
JWThewes: ls "$TF_DIR/environments/"*.tfvars ... | grep -v '\.example$'
  The *.tfvars glob can't match dev.tfvars.example (different suffix), so the grep -v never trips.

JWThewes (changes requested):

scripts/generate-env.sh
eipasteur: Removed grep -v '\.example$' from both scripts — the *.tfvars glob never matches *.tfvars.example (different suffix), so the filter was dead. Also removed the misleading "keep in sync with deploy-frontend.sh" comment from generate-env.sh since deploy-frontend.sh is now converted to the same discovery pattern.

scripts/deploy-frontend.sh
eipasteur: Converted all four scripts to environment discovery. generate-env.sh and deploy-frontend.sh drop the dead grep and add a fallback to "dev prod" when no *.tfvars exist (fixes the fresh-clone regression). deploy-terraform.sh and bootstrap.sh replaced their hardcoded dev|prod checks with the same discovery+fallback pattern.

scripts/generate-env.sh
eipasteur: Removed grep -v '\.example$' from both scripts — the *.tfvars glob never matches *.tfvars.example (different suffix), so the filter was dead. Also removed the misleading "keep in sync with deploy-frontend.sh" comment from generate-env.sh since deploy-frontend.sh is now converted to the same discovery pattern.

JWThewes (approved):

PR #255, opened 2026-06-11
feat: show who answered agent questions and link them from the timeline

### LINKED ISSUE #10: [FEATURE]: Show who answered an agent question
When a question has been answered, the UI just shows it as "answered" with no indication of who answered or when. Display (and store ?) the responder's name and timestamp, especially important in multi-user collaboration scenarios.

JWThewes:
Closes #10

## What

When a question is answered it previously just showed as "answered" with no indication of who answered or when. This PR adds responder traceability for multi-user collaboration scenarios:

- **Who answered**: the responder's identity (Cognito sub + display name) and an answered-at timestamp are persisted server-side and shown in the Q&A history as `Answered by {name} · {timestamp}`.
- **Timeline linkage**: `question_asked` / `question_answered` timeline events now carry a `questionId` and are clickable — they scroll to and highlight the question card on the current page, or deep-link to the sprint's Inception page (`#question-{id}`) which renders all questions including the answered history.

### Backend
- `lambda/questions` (PUT): reads the Cognito User Pools authorizer claims (`sub`, `custom:display_name`, `email`) and persists `answered_by`, `answered_by_name`, `answered_at` on the Neptune `Question` vertex; the existing DynamoDB sync now includes `answeredBy`/`answeredByName`. Server-side capture keeps the responder identity authoritative (not client-supplied).
- `lambda/agents` (`POST /agents/{taskId}/questions/{questionId}/answer`): stores `answeredByName` alongside the existing `answeredBy` sub, and now **best-effort syncs the answer + responder fields to the Neptune Question vertex**. Previously this endpoint only wrote DynamoDB, so answers given from the Agent page never reached the sprint pages — the question stayed "pending" there forever and never appeared in Q&A history (pre-existing bug).
- `lambda/timeline-events`: accepts/returns an optional `questionId` (`question_id` property).
- `lambda/submit-question`: stamps `question_id` on the server-created `question_asked` event.

### Frontend
- Q&A history shows the responder + timestamp; question cards carry `id="question-{id}"` DOM anchors.
- All four pages (Inception/Construction/Review/Agent) pass `questionId` when creating `question_answered` timeline events; ReviewPage previously created no event at all (fixed) and AgentPage now also broadcasts `question.answered` + reloads, matching the other pages.
- Timeline items (`ActivityPanel`, `TimelinePanel`) become clickable when an event references a question (new `lib/questionAnchor.ts` + `hooks/useQuestionAnchor.ts`).

## ⚠️ Lambda runtime / build pattern change for `lambda/questions`

To make the questions Lambda testable against the shared gremlin-server testcontainer, it was migrated from raw `gremlin` + `gremlin-aws-sigv4` (hardcoded wss/SigV4) to the `neptune-lambda-client` pattern already used by `timeline-events` and `ws-connection`:

- **Runtime: `nodejs18.x` → `nodejs24.x`**
- **Packaging**: from `npm_requirements` + `shared/` zip-prefix to an ESM workspace bundled with esbuild (`npm run build -w questions` → `.build/index.mjs`); the per-lambda `package-lock.json` was removed in favour of the root workspace
- Terraform: `terraform/modules/api/lambda/main.tf` updated accordingly (a deploy will replace the function code and runtime; env vars, IAM role and VPC config are unchanged)

Handler logic is unchanged by the refactor.

## Tests

- New `lambda/questions` suite (12 tests): answer persistence and mapping of `answeredBy`/`answeredByName`/`answeredAt` (verified via follow-up GET), DynamoDB sync payload, `custom:display_name` → `email` → empty claims fallback, draft answers not stamping a responder, plus GET/POST/404/405/OPTIONS baselines.
- `lambda/timeline-events` tests extended for `questionId` persistence.
- Full suite: 454/454 passing; frontend `tsc` + build clean; `terraform fmt` clean.

## Out of scope

- Backfill of legacy answered questions (they show no responder).
- The vestigial `answer-question` SQS lambda (no producer; already stores `answeredBy`).
- Frontend test infrastructure (none exists in the repo today — deferred to a separate issue).


JWThewes:
Timeline links to question answered.
There you can see the answers and WHO answered when

Screenshot 2026-06-11 at 09 49 34

frontend/src/components/layout/ActivityPanel.tsx
jeromevdl: Warning "unicorn(no-array-reverse)" 

It has no impact as the array is temporary, but to reduce the number of warnings, I'd change to `.toReversed()`

jeromevdl: same here

lambda/agents/index.js
```diff
@@ -1183,16 +1185,35 @@ exports.handler = async (event) => {
...
+            .V()
+            .has('Question', 'id', questionId)
+            .property(cardinality.single, 'structured_answer', structuredAnswerJson)
+            .property(cardinality.single, 'answered_by', userId)
+            .property(cardinality.single, 'answered_by_name', userName)
+            .property(cardinality.single, 'answered_at', new Date().toISOString())
+            .next();
+        });
+      } catch (e) {
+        console.error('Neptune question sync failed:', e.message);
```
jeromevdl, line 1224: Should we do more than just logging? DLQ? metrics/alarm in cloudwatch ?

frontend/src/hooks/useQuestionAnchor.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+export function useQuestionAnchor(ready: boolean) {
+  const location = useLocation();
+
+  useEffect(() => {
+    if (!ready) return;
+    const match = location.hash.match(/^#question-(.+)$/);
+    if (!match) return;
+    // Defer until after the current render commits the anchors
+    const t = window.setTimeout(() => scrollToQuestion(match[1]), 150);
```
jeromevdl: what happens if 150 ms is not enough ?

lambda/agents/index.js
```diff
@@ -1183,16 +1185,35 @@ exports.handler = async (event) => {
...
-          UpdateExpression: 'SET #s = :s, structuredAnswer = :a, answeredBy = :u, answeredAt = :t',
+          UpdateExpression:
+            'SET #s = :s, structuredAnswer = :a, answeredBy = :u, answeredByName = :n, answeredAt = :t',
           ExpressionAttributeNames: { '#s': 'status' },
           ExpressionAttributeValues: {
             ':s': 'answered',
             ':a': structuredAnswerJson,
             ':u': userId,
+            ':n': userName,
             ':t': Date.now(),
```
jeromevdl, line 1201: answeredAt is `Date.now()` here (current time in ms) and `new Date().toISOString()` (string) in the other lambda (question). Isn't there a problem ? We should probably keep the same format everywhere (not sure if there is a common practice in the project but ISO would be more standard...)

terraform/modules/api/lambda/main.tf
```diff
@@ -685,17 +685,16 @@ module "questions_lambda" {
...
     {
-      path             = "${path.module}/../../../../lambda/questions"
-      npm_requirements = true
-    },
-    {
-      path          = "${path.module}/../../../../lambda/shared"
-      prefix_in_zip = "shared"
+      path = "${path.module}/../../../../lambda/questions"
+      commands = [
+        "cd ../.. && npm run build -w questions",
```
jeromevdl: `cd ../..` can be problematic, we should use `${path.root}`

frontend/src/pages/AgentPage.tsx
```diff
@@ -132,13 +138,20 @@ export default function AgentPage() {
   const handleAnswerQuestion = async (questionId: string, answer: StructuredAnswer) => {
```
jeromevdl: `handleAnswerQuestion` is more or less the same in the 4 pages, maybe extract it in a common hook `useAnswerQuestion` to avoid duplication and having to modify the 4 each times

jeromevdl (commented):
Few comments and one doubt on the `answeredAt` type which is not consistent (string vs long)

frontend/src/components/layout/ActivityPanel.tsx
JWThewes: Done in dd13383 — went one step further: `findLastIndex` expresses the intent directly and avoids the copy entirely (bumped tsconfig `lib` to ES2023 for it, target stays ES2022).

JWThewes: Done in dd13383 — same treatment, this one is now a plain `toolCalls.findLast(...)`.

lambda/agents/index.js
JWThewes: Good point. A DLQ does not really fit here — this is a synchronous API request path (no queue involved), and the failure is recoverable from the UI: the DynamoDB write already succeeded so the agent sees the answer; only the sprint pages would lag (question stays pending there and can be re-answered).

In dd13383 I added a CloudWatch metric via Embedded Metric Format (`collaborative-ai-dlc/NeptuneQuestionSyncFailure`, dimension `Lambda=agents`) on the failure path, so an alarm can be attached without new dependencies or infra changes. Happy to add the `aws_cloudwatch_metric_alarm` in Terraform too if you think it belongs in this PR.

frontend/src/hooks/useQuestionAnchor.ts
JWThewes: Then the scroll silently did nothing — fair catch. In dd13383 the one-shot timeout is replaced with a bounded retry: it polls for the anchor every 100ms and gives up after ~2s (covers slow renders; the cap avoids polling forever when the hash references a deleted/stale question).

lambda/agents/index.js
JWThewes: Each store is internally consistent, they just differ from each other (pre-existing): the agent-questions DynamoDB table uses epoch millis everywhere (`createdAt` in submit-question, the answer-question lambda, and `frontend/src/services/agents.ts` types them as `number`), while the Neptune graph uses ISO-8601 strings throughout, which is what the Q&A history renders.

Switching the DynamoDB write to ISO would mix string and number values in the same attribute across old/new rows, which seemed worse than the status quo. I added a comment in dd13383 documenting why the two writes differ. If you prefer, a repo-wide normalization to ISO (with readers tolerating both) could be a follow-up issue.

terraform/modules/api/lambda/main.tf
JWThewes: Fixed in dd13383 using `cd ${abspath("${path.module}/../../../..")}` — anchored on the module file location and absolute, so it does not depend on the working directory (`path.root` would anchor on the root module instead, which moves around between `terraform/` and `terraform/environments/*`). Note the relative `cd ../..` pattern exists in 14 other lambda build blocks that pre-date this PR; happy to convert them all in a follow-up so this PR stays scoped.

frontend/src/pages/AgentPage.tsx
JWThewes: Done in dd13383 — extracted `frontend/src/hooks/useAnswerQuestion.ts`, used by all four pages. It also folds in the `handleDismissQuestion` handler, which was duplicated verbatim across the same four pages. The Agent page keeps its one real difference via an optional `submitAnswer` override (it answers through the agents endpoint instead of the questions API); the other three use the default.

JWThewes:
@jeromevdl I addressed your comments - would be cool if you can check again

jeromevdl (approved):

jeromevdl:
Thanks @JWThewes. Yes sorry, I noticed things that were there before this PR. I let you decide if we should clean them more broadly (terraform, date format). For the date, just wanted to be sure we don't try to compare things that are not comparable.... and I fear we try at some point.

JWThewes:
> Thanks @JWThewes. Yes sorry, I noticed things that were there before this PR. I let you decide if we should clean them more broadly (terraform, date format). For the date, just wanted to be sure we don't try to compare things that are not comparable.... and I fear we try at some point.

I'll create an issue out of this. We should fix that but I wouldn't do it as part of this PR.

PR #261, opened 2026-06-11
feat(realtime): global membership-scoped token enforcement (discussions PR 1/8)

JWThewes:
## Summary

PR 1 of the discussions feature ([plan §4/§13]): closes the pre-existing realtime authorization gap **globally** before any chat feature ships. Both realtime fabrics historically verified only "valid Cognito JWT" — any signed-in user could join any Yjs doc or app-WS scope, and any connected client could broadcast arbitrary spoofable events via `ws-message`.

### Connect path + session lifecycle (§4a)
- **New HMAC scope tokens** (10 min TTL): `base64url(payload).hmac` minted by the new `lambda/discussions` workspace (`POST /sprints/{id}/realtime-token`, `POST /projects/{id}/realtime-token`) after a `HAS_MEMBER` membership check. Secret: Terraform `random_password` → SSM SecureString `/{project}/{env}/realtime-doc-secret`.
- **yjs-server**: verifies signature, expiry, scope coverage (deny-by-default doc-name extractor covering every existing prefix), and principal binding (`docToken.sub === JWT sub`); force-closes sockets at token expiry (close code `4401`). Secret injected as ECS secret.
- **ws-connection**: same four checks at `$connect`; stores `tokenExp` on the connection row.
- `DOC_TOKEN_ENFORCE` env is an operational kill switch only (default enforcing).

### Send path (§4b, D10)
- **ws-message**: broadcast target is the sender's *registered* document (body `documentId` advisory only); two-event client allowlist (`question.answered`, `sprint.phaseChanged`); legacy scan-all `broadcast` + client `notification` removed; sends from expired connections rejected.
- **AppSidebar payload-blind fix** (round-4 blocker): `sprint.phaseChanged` handler re-fetches the sprint and navigates from the *fetched* phase only.
- **`realtime.ts` send() spread-order fix**: payload fields can no longer clobber the route action. Side note: this also un-breaks the client `sprint.phaseChanged` reload hint, which was silently dropped by ws-message before (the payload's `action` overwrote `broadcastToDocument` on the wire).

### Fan-out filters (§4a, round-4 finding)
`tokenExp > now` applied **at send time** in ws-message, ws-connection disconnect fan-out, notify, submit-question, mcp-server-graph, and acp-client — whose 10 s connection cache now stores `{connectionId, tokenExp}` pairs and filters per send, never at cache-fill time.

### Frontend
Shared `lib/realtimeToken.ts` (cached, single-flight, pre-expiry refresh) plugged into `useYjsDocument` and `realtime.ts` with proactive reconnect before expiry.

## Demo gate (plan §13)
- non-member / forged / expired / other-user token → rejected on both fabrics
- spoofed `discussion.message` / `agent.chunk` via broadcastToDocument → dropped
- allowlisted reload-hints still work; spoofed `sprint.phaseChanged` with a bogus phase causes at most a re-fetch, never wrong navigation
- expired connection inside the 10 s broadcast-cache window receives nothing
- Yjs socket closes at expiry (4401) and recovers via reconnect
- all existing collab features still work (inception draft, question drafts, presence, agent stream)

## Tests
- `lambda/yjs-server/test`: token sign/verify vectors + **doc-name inventory test** enumerating every doc prefix in the codebase, pinning both extractor copies (yjs standalone + shared) in sync
- `lambda/ws-connection` / `lambda/ws-message`: full authorization matrices (forged/expired/scope/sub-mismatch, allowlist, registered-target binding, expiry filters)
- `lambda/discussions`: token issuance against the gremlin testcontainer (member/admin/owner/non-member/404/401)
- `lambda/agents-ecs`: acp-client cache invariants
- 548 backend tests pass; frontend typecheck + build green; `terraform validate` clean

## Deploy notes
- Yjs ECS image + agent pool image are content-hashed — rebuilt automatically on apply; recycle the pool after deploy per the existing release procedure
- Existing connection rows without `tokenExp` get a fan-out grace (≤1 h TTL); new connects always stamp it


JWThewes:
This PR is part of a lot more PR's that are coming for a disucssions feature. The discussions feature will allow teams to discuss about questions, prompts, or sprints in general. It will be a realtime chat where you can include the agent as well.
Discussions will be stored in the graph and can be used later one by the agents as they might contain valuable information

JWThewes:
# Discussions: contextual team chat with AI assist — stack overview (PRs 1–8)

This change introduces **Discussions** — threaded, real-time conversations anchored directly to the things you work on in a sprint — delivered as a stacked chain of 8 PRs (#261 → #268). This comment summarizes what is available after merging the whole stack.

Screenshot 2026-06-12 at 11 24 19

## What you get after merging all 8 PRs

### 💬 Contextual discussion threads
- Start a discussion on any sprint entity: inception, questions, requirements, user stories, tasks, reviews, and general info — via a **Discuss** button on the respective cards (#262, #264)
- Real-time chat between all sprint members: live message sync, typing indicators, and presence dots; messages survive reloads and reach peers even if the sender disconnects (#264)
- Markdown rendering (GitHub-flavored) in messages (#264)

### 🔔 Unread tracking & mentions
- Per-user unread badges on every Discuss button and a first-unread divider when opening a thread; unread state only clears when messages are actually visible on screen (#263, #265)
- `@`-mentions with a member combobox; mentioned users get an in-app toast with jump-to-thread (online, in-app) (#263, #265)

### ✅ Thread lifecycle & moderation
- **Resolve / reopen** threads with an optional resolution summary and an outcome-message pointer — "what did we decide" becomes durable and is surfaced to agents as team decisions (#263, #265)
- **Redaction** (admin/owner only): message content is purged and replaced with an audit stub; redactions propagate even to clients that missed the live event (#263, #265)
- Timeline events for discussion started, resolved, and redacted messages (#262, #263)

### 🔎 Discovery
- New **Discuss tab** in the Activity Panel: thread list with anchor badges, unread counts, open/resolved filter, and sprint-wide text search (#263, #265)
- Discussion threads appear as nodes in the Sprint Graph (#262, #267)

### ✨ In-thread AI assist
- ✨ menu in the chat input (or `/` in an empty input): **Suggest an answer** (question threads, advice-only), **Summarize discussion**, **Explain in detail**, or a free-form prompt (#266, #267)
- The assistant streams its reply live into the thread, reads thread context and the sprint graph itself, and its runs show up in the agent run history — without ever interfering with phase progress (#266, #267)
- Agents in *all other phases* are nudged to consult discussion resolutions as team decisions (#267)

### 🔒 Security hardening (whole platform, not just discussions)
- **Membership-scoped realtime tokens** close a pre-existing gap: previously any signed-in user could join any realtime document; now both realtime fabrics (Yjs + app WebSocket) enforce short-lived, HMAC-signed, membership-checked scope tokens (#261)
- **No client-spoofable events**: all broadcast events (`question.answered`, `sprint.phaseChanged`, discussion events) are now server-origin only; the client event allowlist is empty (#261, #268)
- Atomic creation/message guards prevent duplicate threads/messages under concurrency; keyset pagination keeps large threads fast (#262)

### 📚 Docs & ops
- New user guide `docs/using-the-platform/discussions.md` (role matrix, mentions, assist expectations) (#267)
- Architecture docs updated with the real token-enforcement model plus a doc-secret rotation runbook (#267, #268)

## PR chain

| PR | Scope |
|----|-------|
| #261 | Realtime authorization: membership-scoped tokens on both fabrics |
| #262 | Graph data model + core REST with atomic guards |
| #263 | Resolve, redact, read cursors, search, mentions (backend) |
| #264 | Chat UI core (sheet, thread, live sync) |
| #265 | Unread, mentions, resolve/redact UI, Discussions tab |
| #266 | Assist backend: pool-worker discussion phase |
| #267 | Assist UI, run-history visibility, docs |
| #268 | Server-origin events, empty client allowlist |

**Verification across the stack:** 614+ backend tests green, frontend typecheck + production build clean, `terraform validate` clean.

frontend/src/components/layout/AppSidebar.tsx
```diff
@@ -36,17 +36,32 @@ export function AppSidebar() {
     loadSprint();
   }, [loadSprint]);
 
-  // Re-fetch sprint on agent/phase events and auto-navigate on phase change
+  // Re-fetch sprint on agent/phase events and auto-navigate on phase change.
+  //
+  // INVARIANT (plan §4b): `sprint.phaseChanged` is an allowlisted CLIENT-origin
+  // event until PR 6 — allowlisted client events must never carry rendered or
```
jeromevdl, line 42: what are those plans § ? also unclear why we split into so many PRs (how to test?) and maybe comments should not contain this kind of info...

JWThewes: I'll remove these references. They're coming out of the implementation plan.

Smaller PRs for easier review. You can deploy the phase 8 branch as this contains everything.
So you can test. I just tried to break it down. If you prefer Incan create one large PR though? @jeromevdl

frontend/src/hooks/useYjsDocument.ts
```diff
@@ -64,6 +87,18 @@ export function useYjsDocument(documentId: string | null, userName?: string, use
         console.log('Yjs WebSocket connected');
         reconnectAttempts = 0;
 
+        // Proactively reconnect shortly before the scope token expires — the
+        // server force-closes the socket at expiry (close code 4401), so
+        // beating it keeps the session seamless.
+        if (tokenRefreshRef.current) clearTimeout(tokenRefreshRef.current);
+        tokenRefreshRef.current = window.setTimeout(() => {
```
jeromevdl, line 94: playing with timeouts is a bit fragile. I don't have better solution but it seems a bit light to keep the session live

PR #270, opened 2026-06-12
feat: discussions — entity-anchored team chat with in-thread agent assist

JWThewes:
# Discussions — entity-anchored team chat with in-thread agent assist

One review for the complete discussions feature. It was developed and pre-reviewed as a stack of 8 PRs (#261–#268, all marked merged into this branch); this PR rolls the whole feature up against `main`. The granular commit history of every phase is preserved on this branch, and each stack PR retains its individual review discussion.

## What it does

A chat-like discussion thread can be attached to any sprint-scoped entity (question, requirement, user story, task, review, general info, the inception prompt, or the sprint itself). Team members chat in real time (Yjs live sync, typing/presence awareness), every message is durably persisted as a vertex in Neptune, and the project's configured agent CLI can be invoked inside the thread ("Suggest an answer", "Summarize discussion", "Explain in detail", or a free-form instruction), streaming its reply into the chat and persisting it as a regular message.

User documentation: `docs/using-the-platform/discussions.md`.

### 1. Realtime channel authorization — global (#261)
Closes the pre-existing gap where any signed-in user could join any Yjs doc or app-WS scope and broadcast arbitrary events:
- Short-lived HMAC scope tokens (10 min TTL) minted by `lambda/discussions` after a membership check; verified at connect on both fabrics (yjs-server ECS + ws-connection), with expiry enforcement and principal binding
- `ws-message` hardened: broadcast target is the sender's registered document only; legacy scan-all `broadcast` and client `notification` paths removed
- Token-expiry liveness filters applied at send time in every fan-out path (ws-message, ws-connection, notify, submit-question, mcp-server-graph, acp-client)
- Frontend: shared cached/single-flight token fetcher with pre-expiry reconnect

### 2. Graph data model + core REST (#262)
- New Neptune vertices `Discussion` and `DiscussionMessage`, edges to the anchor entity and authors; one thread per entity (schema future-proofed for titled threads)
- `lambda/discussions` REST: atomic get-or-create, message append, keyset pagination + change deltas
- Neptune cannot unique-constrain ids, so thread creation, message append, and assist concurrency are serialized via DynamoDB conditional writes (one `discussion-locks` table)
- Messages flow server-first: REST persists to Neptune (source of truth), the backend fans out full payloads over WS, clients merge into Yjs by id — Yjs is a live-sync optimization, never the durability path

### 3. Thread features (#263)
- Resolve / reopen with optional resolution summary; admin/owner message redaction
- Per-user unread state via per-thread composite cursor `(lastReadAt, lastReadMessageId)`
- Sprint-scoped bounded search; @-mentions with in-app notification fan-out (online users)

### 4–5. Chat UI (#264, #265)
- Non-modal `DiscussionPanel` hosted in the right activity panel, grouped message blocks, load-older pagination, first-unread divider, visibility-gated read marking (IntersectionObserver)
- `DiscussButton` entry points on every supported entity (cards, question blocks, sprint header) with unread badges and ongoing-thread indicator
- Activity-panel Discussions tab listing all sprint threads; mention toasts; typing indicators via Yjs awareness
- Responsive: resizable inline column on lg+, overlay below

### 6–7. Agent assist (#266, #267)
- Assist runs as a pool-worker `discussion` phase — same CLIs, credentials, streaming, and MCP graph tools as real agent runs; no separate Bedrock integration
- `POST .../assist` → per-thread DynamoDB lock (heartbeat-renewed, takeover-safe) → pool dispatch; 202 + `executionId` for stream correlation
- Explicit "starting" UI state (pool pickup can take 15–60 s), streaming bubble, persisted reply with requested-by caption; assists never touch the Sprint's `current_*` agent status
- `suggest-answer` is advice-only and only offered on question-anchored threads — it never modifies the question
- Agent runs get a discussions nudge in phase prompts so they can self-serve thread context via MCP

### 8. Server-origin event hardening (#268)
- `question.answered` and `sprint.phaseChanged` are now emitted server-side by the questions/sprints/agents lambdas (`lambda/shared/ws-fanout.js`); the frontend never broadcasts events
- The ws-message client-event allowlist is **empty** — connected clients cannot inject any event; handlers treat events as payload-blind reload hints

## Product decisions (recorded for reviewers)

- Member-level resolve/assist (audited; role configuration is a v2 candidate)
- Mentions are online/in-app only (offline/email delivery is v2)
- No assist cost cap: per-thread lock + pool capacity are the limiters; every assist is audited via AgentRun history
- Message editing is out of scope (admin redaction is in; author self-edit is not)

## Infrastructure

- New: `discussion-locks` + realtime-token SSM secret; discussions lambda + routes; assist dispatch wiring in the agents lambda/pool
- DynamoDB tables and WS fan-out env vars threaded through `terraform/modules/api` and `terraform/modules/realtime`
- Yjs ECS image and agent pool image are content-hashed — rebuilt on apply; recycle the pool after deploy per the existing release procedure

## Testing

- 627 backend tests pass (27 suites, incl. gremlin testcontainer integration tests for the discussions lambda, full authorization matrices for ws-connection/ws-message, assist-lock concurrency, fanout filters)
- Frontend typecheck + lint clean
- Each stack PR carried its own demo gate; see the individual PRs for per-phase test plans

## Review notes

- This branch is synced with `main` (merge commit `6c3ba9a`): main's `useAnswerQuestion` refactor (#255) was adopted with its client-side `question.answered` broadcast removed (dead since phase 8 — the event is server-origin), and main's MCP startup fix (#227) is included
- Phases 1–3 were squash-merged within the stack (#261–#263); phases 4–8 landed via the rollup merge `ee760f7` whose tree provably equals the stack head


lambda/discussions/index.js
```diff
@@ -0,0 +1,1457 @@
...
+    if (method === 'POST' && path.endsWith('/assist')) {
+      return await invokeAssist(event, res);
+    }
+    if (method === 'PUT' && path.endsWith('/{discussionId}')) {
+      return await updateDiscussion(event, res);
+    }
+
+    return res(404, { error: 'Not found' });
+  } catch {
+    console.error('discussions handler error');
```
jeromevdl, line 91:
```suggestion
  } catch (err) {
    console.error('discussions handler error', err);
```

lambda/ws-message/index.js
```diff
@@ -1,76 +1,101 @@
...
+  const registeredDocumentId = senderRow.documentId.S;
+  if (body.documentId && body.documentId !== registeredDocumentId) {
+    // Body value is advisory only — log the mismatch, use the registered doc.
+    console.warn(
+      `documentId mismatch from ${connectionId}: body="${body.documentId}" registered="${registeredDocumentId}"`,
+    );
+  }
+
+  await broadcastToDocument(registeredDocumentId, message, connectionId);
+  return { statusCode: 200 };
```
jeromevdl: With an empty `CLIENT_EVENT_ALLOWLIST` this is dead code...

Not sure to understand why you updated it and kept it. 

The complete lambda is not really used except as a defensive mechanism, so maybe we can shrink it to just the defensive action (return everything). We can use git later on if needed to retrieve the code ?

Or did I misunderstand the purpose of it?

frontend/src/components/discussion/DiscussionProvider.tsx
```diff
@@ -0,0 +1,284 @@
...
+    setDiscussions([]);
+    reloadDiscussions();
+  }, [reloadDiscussions]);
+
+  // Refresh on server fanout — covers other users' messages, resolves
+  // and redactions. Cheap: one list query per event.
+  useEffect(() => {
+    if (!sprintId) return;
+    const unsubs = [
+      realtimeService.on('discussion.message', () => reloadDiscussions()),
```
jeromevdl: on super chatty sessions, that can become an issue. 10 people in the room, sending messages in parallel will make that much calls to Neptune list (multiplied by the number of projects in parallel). 

I think API GW, Lambda and probably Neptune too can scale to absorb the traffic, so we can probably keep this for the time being but worth adding a comment on a scalability option

Or do something like:

```
// Coalesce bursts of fan-out events into a single list query. A sprint-wide
// message broadcasts to every connected member, and each would otherwise
// fire its own GET /discussions (a non-trivial Neptune traversal). Trailing
// debounce collapses rapid chatter into one refresh per ~600ms window.
const reloadTimerRef = useRef<number | null>(null);
const scheduleReload = useCallback(() => {
    if (reloadTimerRef.current) clearTimeout(reloadTimerRef.current);
    reloadTimerRef.current = window.setTimeout(() => {
      reloadTimerRef.current = null;
      reloadDiscussions();
    }, 600);
  }, [reloadDiscussions]);

  // Clear any pending reload on unmount.
  useEffect(
    () => () => {
      if (reloadTimerRef.current) clearTimeout(reloadTimerRef.current);
    },
    [],
  );
```

```
useEffect(() => {
    if (!sprintId) return;
    const unsubs = [
      realtimeService.on('discussion.message', () => scheduleReload()),
      realtimeService.on('discussion.updated', () => scheduleReload()),
    ];
    return () => unsubs.forEach((u) => u());
  }, [sprintId, scheduleReload]);
```

jeromevdl: 600ms can be reduced if it looks laggy

frontend/src/components/discussion/DiscussionsTab.tsx
```diff
@@ -0,0 +1,205 @@
...
+  sprint: 'Sprint',
+  inception: 'Inception',
+  question: 'Question',
+  requirement: 'Req',
+  userstory: 'Story',
+  task: 'Task',
+  review: 'Review',
+  generalinfo: 'Info',
+};
+
```
jeromevdl, line 26:
```suggestion

const EMPTY_DISCUSSIONS: Discussion[] = [];

```

frontend/src/components/discussion/DiscussionsTab.tsx
```diff
@@ -0,0 +1,205 @@
...
+          status: statusFilter === 'all' ? undefined : statusFilter,
+        })
+        .then((r) => setSearchResults(r.results))
+        .catch(() => setSearchResults([]))
+        .finally(() => setSearching(false));
+    }, 350);
+    return () => clearTimeout(timer);
+  }, [query, statusFilter, sprintId]);
+
+  const discussions = ctx?.discussions ?? [];
```
jeromevdl:
```suggestion
  const discussions = ctx?.discussions ?? EMPTY_DISCUSSIONS;
```

frontend/src/components/discussion/DiscussionsTab.tsx
```diff
@@ -0,0 +1,205 @@
...
+  const filtered = useMemo(
+    () => discussions.filter((d) => statusFilter === 'all' || d.status === statusFilter),
+    [discussions, statusFilter],
```
jeromevdl, line 73: See suggestions above to avoid this lint finding (performance issue)

frontend/src/lib/discussion.ts
```diff
@@ -0,0 +1,129 @@
...
+};
+
+export const compareByUpdated = (a: DiscussionMessage, b: DiscussionMessage): number => {
+  if (a.updatedAt !== b.updatedAt) return a.updatedAt < b.updatedAt ? -1 : 1;
+  if (a.id !== b.id) return a.id < b.id ? -1 : 1;
+  return 0;
+};
+
+export const sortMessages = (messages: DiscussionMessage[]): DiscussionMessage[] =>
+  [...messages].sort(compareByCreated);
```
jeromevdl:
```suggestion
  [...messages].toSorted(compareByCreated);
```

lambda/agents-ecs/pool-worker.js
jeromevdl, line 1: please address the linter findings if they make sense

docs/concepts/architecture.md
```diff
@@ -79,13 +79,20 @@ flowchart TB
...
+The Yjs server is small: on every WebSocket upgrade it pulls the Cognito JWT from the URL query string, verifies it in-process, then verifies a short-lived HMAC-signed **scope token** (issued by the discussions Lambda only to project members, bound to the caller's Cognito identity, covering exactly the requested sprint/project) before opening or joining a `Y.Doc` instance keyed by the URL path, and broadcasts CRDT sync (`type 0`) and awareness (`type 1`) updates between connected clients. Unknown document-name formats are rejected, and each socket is force-closed when its scope token expires (close code 4401) — the client reconnects with a fresh token, so membership is re-validated at most every ten minutes. A removed member can therefore keep an already-open session for at most the remaining token life. Live document state lives **only in process memory**; the server makes zero S3 or DynamoDB writes. When the last client disconnects, the doc is destroyed sixty seconds later.
```
jeromevdl, line 82: great stuff

frontend/src/components/discussion/DiscussionPanel.tsx
```diff
@@ -0,0 +1,277 @@
...
+    }
+  };
+
+  const redact = async (messageId: string) => {
+    if (!discussion) return;
+    try {
+      const redacted = await discussionsService.redact(sprintId, discussion.id, messageId);
+      applyMessages([redacted]);
+    } catch (err) {
+      console.error('Redact failed:', err);
```
jeromevdl, line 145: The user won't know it failed... I notice this here but there are probably other places where we just write the error on the console. In some cases, it can be useful to inform the user. 
In this case I think it is important as redacting can be for security / legal / compliance reason so should be visible.

frontend/src/components/discussion/DiscussionInput.tsx
```diff
@@ -0,0 +1,285 @@
...
+                  onClick={() => runAssist('custom', assistInstruction.trim())}
+                >
+                  <Sparkles className="h-3 w-3 mr-1" />
+                  Ask
+                </Button>
+              </div>
+            </PopoverContent>
+          </Popover>
+        )}
+        <textarea
```
jeromevdl, line 222: Image

The textarea is too small

jeromevdl: happens only on first render, then it's ok:

Image

lambda/ws-message/index.js
JWThewes: Good catch — you understood it correctly. The allowlist was permanently empty, so everything from the sender-row lookup down (token check, documentId-mismatch logging, `broadcastToDocument` fan-out) was unreachable, and there are no `realtimeService.send()` callers in the frontend either. All realtime events are server-origin via `lambda/shared/ws-fanout.js`.

Shrank the lambda to just the defensive action: absorb the $default/sync/notification routes, log, and drop. The broadcast machinery is recoverable from git history if client-origin broadcast is ever revived.

Fixed in 8ac5074.

frontend/src/components/discussion/DiscussionProvider.tsx
JWThewes: Agreed — adopted your debounce approach. Each fanout was sprint-wide, so an N-member room turned one message into N parallel `GET /discussions` Neptune traversals, multiplied across concurrent sprints.

Added the trailing-debounce `scheduleReload` (extracted `RELOAD_DEBOUNCE_MS = 600` as a named constant), wired `discussion.message` / `discussion.updated` through it, and folded the timer cleanup into the existing subscription effect's teardown so one place owns the timer lifecycle.

Fixed in 399b560.

JWThewes: Kept it at 600ms for now and called that out in the code comment ("lower the delay if it feels laggy"), with the value pulled out as the `RELOAD_DEBOUNCE_MS` constant so it's a one-line tweak. Easy to dial down once we see it in a real session. (399b560)

frontend/src/components/discussion/DiscussionsTab.tsx
JWThewes: Applied suggestions

lambda/agents-ecs/pool-worker.js
JWThewes: Done — all six oxlint findings in this file were pre-existing (none from this PR), but they all made sense to clean up:

- **no-shadow** (×2): the `.filter((f) => …)` callback param shadowed the enclosing `for…of` loop var `f` in the two rule-copy loops. Renamed the param to `name`.
- **no-underscore-dangle** (×4): dropped the leading underscore from `availableClis`, `cliAuthErrors`, `neptuneVal`, and `safeRulesDest`. Module scope already makes them private, so the underscore added nothing. The DynamoDB attribute names in the update expressions were never underscored and are unchanged.

While verifying the repo-wide lint gate I also hit an unrelated `no-unused-vars` **error** in `discussions/index.js` — the earlier clear-text-logging fix dropped `err` from the `console.error` but left `catch (err)`. Switched it to an optional catch binding (`catch {}`) in the same commit so the lint gate is green.

`pool-worker.js` now reports 0 oxlint findings; tests pass. Fixed in 0e32729.

frontend/src/components/discussion/DiscussionInput.tsx
```diff
@@ -0,0 +1,285 @@
...
+    },
+    [value, autoGrow],
+  );
+
+  const send = useCallback(() => {
+    const trimmed = value.trim();
+    if (!trimmed || disabled) return;
+    // Only count mentions whose @label text survived editing.
+    const mentions = [...mentioned.entries()]
+      .filter(([, label]) => trimmed.includes(`@${label}`))
```
jeromevdl: we could have several people mentioned (for example: john / johnny) with this regex

frontend/src/hooks/useYjsDocument.ts
jeromevdl: is it normal to have one global doc here for the full component. If `documentId` changes, we should probably have another one to avoid any stale data... can we move it to the `useEffect`?

I know it was already before this PR...

lambda/agents-ecs/mcp-server-graph/index.js
jeromevdl: finding to fix

lambda/discussions/index.js
```diff
@@ -0,0 +1,1462 @@
...
+    }
+
+    return res(404, { error: 'Not found' });
+  } catch (err) {
+    const rawMessage = err instanceof Error ? err.message : 'unknown error';
+    const safeMessage = String(rawMessage).replace(
+      /(secret|token|password|authorization|apikey|api[-_ ]?key)\s*[:=]\s*\S+/gi,
+      '$1=[REDACTED]',
+    );
+    console.error(`discussions handler error: ${safeMessage}`);
```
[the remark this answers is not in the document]
jeromevdl: suppress this warning if you think the redaction is enough, or make it stronger.

lambda/discussions/index.js
jeromevdl, line 1: This new lambda is pretty big (~ 1500 loc), could we split it somehow? handlers / business (service) / ddb (data access layer), to make it more readable / diggestable. It would also simplify tests...

frontend/src/hooks/useYjsDocument.ts
jeromevdl:
```suggestion
  const doc = useMemo(() => new Y.Doc(), [documentId]);
```

with the import and it works !

jeromevdl (changes requested):
Globally, the code is clean and the feature is really nice.
- ⚠️ Frontend issue, see https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/270/changes#r3427923286
- There are a few lint findings and comments (the new lambda is quite big...) but overall it is well written, tested and the feature is really cool.

frontend/src/components/discussion/DiscussionPanel.tsx
JWThewes: Agreed, especially for redaction — silently failing on a security / legal / compliance action is the worst case, since the operator walks away believing sensitive content was removed when it wasn't.

Added a `redactError` state set in the `catch` and rendered it in a dismissible inline banner (same style as the existing assist-error banner), with copy that makes the failure explicit: "Couldn't redact the message: … The content was NOT removed — please try again." Cleared on dismiss, on a new redact attempt, and when switching threads.

Scoped this fix to redaction since that's the high-stakes one. You're right that there are other `console.error`-only spots — agree they're worth a sweep, but I'd rather do that as a focused follow-up than widen this PR. Happy to file a tracking issue if you'd like.

Fixed in cd5b2c6.

frontend/src/hooks/useYjsDocument.ts
JWThewes: `useMemo` import in `useYjsDocument.ts` added with f3426d5

frontend/src/components/discussion/DiscussionInput.tsx
JWThewes: Good catch — and it was actually worse than a regex issue: it was a plain `trimmed.includes('@' + label)` substring check, so `@johnny` also matched the `john` label and would have notified john.

Extracted the logic into a tested `mentionedUserIds()` helper in `lib/discussion.ts` that requires each `@label` to sit on a token boundary — preceded by start-of-string or a non-mention char, and not followed by another mention char — mirroring the input's `@`-token charset (`[\w.+-]`). The label is regex-escaped so dotted email-derived labels like `john.doe` match literally.

Added unit tests for the prefix-collision case (`@johnny` → only johnny), both-mentioned, start-of-string, bare label (no `@`), email-glued `foo@john`, and regex-special / dotted labels.

Fixed in 6dcea2c.

lambda/agents-ecs/mcp-server-graph/index.js
JWThewes: Fixed in ba677b4.

The finding is CodeQL `js/clear-text-logging` (high): the `console.error` override mirrored every stderr arg into `/tmp/mcp-graph.log` via `fs.appendFileSync`. Env-derived sensitive data — `GIT_TOKEN`, or repo URLs that can embed credentials (`https://<token>@host/...`) — can reach `console.error` and was being persisted in clear text.

That log file was write-only — nothing in the repo ever reads `/tmp/mcp-graph.log` — so I removed the override and `LOG_FILE` entirely rather than trying to scrub secrets before writing. Plain `console.error` still logs to stderr for all ~30 call sites (MCP servers log to stderr by design, since stdout is the protocol channel). `fs` stays imported — it's still used for the discussion-posted marker file.

agents-ecs tests pass (50).

frontend/src/components/discussion/DiscussionInput.tsx
JWThewes: Fixed in 29404e3. `autoGrow()` (which clamps the height to 40-160px) only ran on change / mention-insert / send, never on mount — so the `rows={1}` box rendered a hair too short at first paint. Added a mount `useEffect(autoGrow)` so the correct height is applied on first render.

JWThewes: Right — your "only on first render" observation pinned it exactly: the first keystroke fired `autoGrow` and snapped it to 40px, which is why it self-corrected. Running `autoGrow` once on mount applies that same clamped height at first paint, so it's correct before any typing. (29404e3)

lambda/discussions/index.js
JWThewes: Done in 5fc33e3. Split along the layers it already had internally, keeping behavior identical:

- `constants.js` — literals + the load-time takeover-safety invariant
- `clients.js` — shared Neptune (query/close) + DDB/SSM/Lambda clients
- `mappers.js` — pure helpers and Neptune→DTO mappers (no I/O — the most unit-testable bits)
- `data-access.js` — Neptune graph + DynamoDB reads/writes
- `services.js` — sprint authorization, WS fan-out, secret resolution
- `handlers.js` — one HTTP handler per route
- `index.js` — slim router (1457 → 94 lines)

It's a pure move + re-export — no logic changed. `handler` and `close` are still exported from `index.js`, and the takeover-safety invariant still fires on import, so all 64 existing tests pass unchanged and the esbuild bundle still builds (it follows the new `./` imports). `handlers.js` is still ~880 lines (the 10 routes), but it's now cleanly separated from the data-access and service layers, which was the main readability/testability win you were after — and the pure `mappers.js` is now independently testable.

I kept the test file as-is for this commit so the split is provably behavior-preserving; happy to follow up with targeted unit tests against `mappers.js` / `data-access.js` now that the seams exist, if you'd like.

JWThewes:
Addressed all of @jeromevdl comments

lambda/discussions/index.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+    if (method === 'POST' && path.endsWith('/redact')) {
+      return await redactMessage(event, res);
+    }
+    if (method === 'POST' && path.endsWith('/assist')) {
+      return await invokeAssist(event, res);
+    }
+    if (method === 'PUT' && path.endsWith('/{discussionId}')) {
+      return await updateDiscussion(event, res);
+    }
```
jeromevdl, line 87: Do you know Powertools can handle this:

https://docs.aws.amazon.com/powertools/typescript/latest/features/event-handler/http/

Maybe not in this PR but worth simplifying our APIs with this... it handle cors and lots of things.

jeromevdl (approved):
Good for me now, all comments addressed. Good job!

PR #287, opened 2026-06-30
chore: remove unused Step Functions orchestration (dead code)

### LINKED ISSUE #283: [Chore]: Remove unused Step Functions orchestration (dead code)
## Summary

The `agent-workflow` Step Functions state machine and its supporting resources are **dead code** — nothing ever starts an execution. Agent orchestration is actually driven by the ECS pool-worker re-triggering the agents Lambda, not by Step Functions. This issue tracks removing the unused SFN plumbing.

## Why it's dead

- No `StartExecution` call exists anywhere in the codebase (lambdas or ECS). A repo-wide search for `StartExecution` / `SFNClient` / `client-sfn` finds only `lambda/answer-question` (a `SendTaskSuccess` callback that depends on a `taskToken` that is never produced).
- The live orchestration loop is: the construction agent runs in the ECS **pool-worker**, which on exit invokes the **agents Lambda** (`AGENTS_LAMBDA_NAME`) to re-trigger the `construction-orchestrator` phase (see `lambda/agents-ecs/pool-worker.js` → `refreshGitTokenIfNeeded`/orchestrator re-trigger, and the `trigger_pr_creation` MCP tool in `lambda/agents-ecs/mcp-server-graph/index.js`). Step Functions is not in this path.
- The `state_machine_arn` variable is plumbed into `module.api` / `module.api.lambda` but is **never consumed** (no env var, no IAM references it) — pure dead wiring.

## Dead resources to remove

**Terraform — `terraform/modules/orchestration/`**
- `aws_sfn_state_machine.agent_workflow` (`main.tf`)
- `aws_iam_role.step_functions` + `aws_iam_role_policy.step_functions` (`main.tf`)
- `aws_cloudwatch_log_group.step_functions` (`main.tf`)
- outputs `state_machine_arn`, `state_machine_name`, `step_functions_role_arn` (`outputs.tf`)
- the `states:SendTaskSuccess` / `states:SendTaskFailure` IAM statement in `questions.tf`

**Terraform — wiring**
- `terraform/main.tf`: the `state_machine_arn = ...` line passed into `module.api`, and the synthesized ARN at the top of the file
- `terraform/outputs.tf`: `agent_workflow_state_machine_arn` output
- `terraform/modules/api/variables.tf` and `terraform/modules/api/lambda/variables.tf`: the unused `state_machine_arn` variables

**Lambda**
- `lambda/answer-question/answer-question.js`: remove the `SFNClient` import, the `@aws-sdk/client-sfn` dependency, and the `if (question.Item.taskToken) { SendTaskSuccess... }` block (the task-token branch is unreachable without the state machine).
- `lambda/get-agent-output` + its `get_agent_output_lambda` module/role in `orchestration/main.tf`: **referenced only by the dead state machine** (the `CheckAgentOutput` state) and never invoked at runtime. Likely dead too — verify nothing else calls it before removing.

## Must be kept (not part of the SFN)

- `create_pr_lambda` (and its role) in the orchestration module — it is live, invoked by the construction MCP server (`trigger_pr_creation`); it has real outputs consumed in `terraform/main.tf`.
- `lambda/answer-question` itself — only the SFN task-token branch is dead, the rest of the handler is live.

## Notes

- After removal, consider whether the `orchestration` module name still makes sense (it would then only hold `create_pr` / question wiring).
- A `terraform plan` should show only destroys for the SFN state machine, its role, log group, and the unused variables — no changes to live agent behavior.
- Confirmed there is no `StartExecution` caller, so removing the state machine cannot break any runtime flow.

jeromevdl:
## Summary

Remove the entire `orchestration` Terraform module and all associated dead code — the Step Functions state machine was never triggered (no `StartExecution` call exists in the codebase).

## What was removed

**Dead Step Functions resources:**
- `aws_sfn_state_machine.agent_workflow` + IAM role/policy + CloudWatch log group
- `get-agent-output` lambda (only referenced by the dead state machine)
- `answer-question` lambda + SQS queue (no producer exists — answering is handled inline by the `agents` and `questions` lambdas)
- All `state_machine_arn` wiring (variables, outputs, module plumbing)

**Dead code deleted:**
- `terraform/modules/orchestration/` (entire module)
- `lambda/answer-question/answer-question.js`
- `lambda/get-agent-output/` (directory + tests)
- `get-agent-output` workspace from root `package.json`

## What was moved

The `create-pr` lambda (the only live resource in the old orchestration module) was relocated to `terraform/modules/api/lambda/` where all other lambdas live. Resources renamed to `create-pull-request` to avoid naming conflicts during the state transition.

## Verification

- `terraform validate` passes
- All 750 tests pass
- Deployed successfully
- No remaining references to `state_machine`, `get_agent_output`, `step_functions`, or `SFNClient`

## Net change

```
16 files changed, 67 insertions(+), 1179 deletions(-)
```

Closes #283


terraform/main.tf
```diff
@@ -13,6 +13,25 @@ terraform {
...
+  from = module.orchestration.module.create_pr_lambda
+  to   = module.lambda.module.create_pr_lambda
+}
+
+moved {
+  from = module.orchestration.aws_iam_role.create_pr
+  to   = module.lambda.aws_iam_role.create_pr
+}
+
+moved {
```
eipasteur: I have a question: that moved block goes `aws_iam_role_policy` --> `aws_iam_role_policy_attachment`, which are different types. Are you sure this works on existing deployments ? did you actually test it against a state that already has the orchestration module, or just a fresh apply? 
I think cross-type moves usually fail with _Unable to Move Resource State_, but on a clean state it wouldn't show up.

The role name changes too (`create-pr` -> `create-pull-request`), so it gets replaced anyway — could we just drop that block? Perms look the same either way. 
Note: I haven't tested this on my side, so I might be missing something.

eipasteur:
A small detail: `create-pr.js:7` still mentions "Step Functions response shape".

terraform/main.tf
jeromevdl: True, I had this before I changed the name but it failed anyway, forgot to remove it.

terraform/main.tf
```diff
@@ -13,6 +13,25 @@ terraform {
...
+  from = module.orchestration.aws_iam_role_policy.create_pr
+  to   = module.lambda.aws_iam_role_policy_attachment.create_pr_basic
+}
+
```
jeromevdl:
```suggestion
```

eipasteur (approved):

PR #288, opened 2026-06-30
feat: add GitRepoLink component with provider icons

jeromevdl:
*Description of changes:*

- Create `GitRepoLink` component (icon + linked repo name, supports github/gitlab, noLink option for clickable containers)
- Extract GitHub/GitLab SVG icons into shared `icons/git-providers.tsx`, replacing deprecated lucide-react brand icons
- Update `Dashboard`, `Project`, `CommandPalette`, `ProjectDiagram`, and `StartInceptionModal` to use `GitRepoLink`
- Update CreateProjectModal and trackerProviders to use shared icons

| Location | Link active? |
|----------|-------------|
| Project detail page (repo card) | Yes |
| StartInceptionModal | Yes |
| Dashboard (grid/list) | No (card is clickable) |
| CommandPalette | No (item is clickable) |
| ProjectDiagram | No (card is clickable) |

Note for @eipasteur: `StartInceptionModal` and `InceptionView` are marked unused. Didn't want to cleanup in this PR but what are these 2 files?


JWThewes:
Noticed one regression while playing with longer repo paths: the root of `GitRepoLink` is `inline-flex`, which sizes to its content and never shrinks to the parent's width — so the `truncate` on the inner span never actually kicks in.

Concretely: on the Project page the repo name used to be a `<p className="truncate">` and truncated fine, and in the Dashboard grid cards the old block-level flex div constrained the span. With a nested GitLab path like `group/subgroup/some-long-project-name` the text now overflows the card in both places.

Suggestion: constrain the container in both branches (link and `noLink`):

```tsx
className={cn('inline-flex min-w-0 max-w-full items-center gap-1', className)}
```

That keeps the component drop-in and restores truncation everywhere it's used.

JWThewes:
@jeromevdl a great focused PR - just a minor visual thing here. Otherwise looks great!

jeromevdl:
> Noticed one regression while playing with longer repo paths: the root of `GitRepoLink` is `inline-flex`, which sizes to its content and never shrinks to the parent's width — so the `truncate` on the inner span never actually kicks in.
> 
> Concretely: on the Project page the repo name used to be a `<p className="truncate">` and truncated fine, and in the Dashboard grid cards the old block-level flex div constrained the span. With a nested GitLab path like `group/subgroup/some-long-project-name` the text now overflows the card in both places.
> 
> Suggestion: constrain the container in both branches (link and `noLink`):
> 
> ```tsx
> className={cn('inline-flex min-w-0 max-w-full items-center gap-1', className)}
> ```
> 
> That keeps the component drop-in and restores truncation everywhere it's used.

should be fixed now:

image

JWThewes (approved):

PR #319, opened 2026-07-17
fix: enforce authorization on v1 history reads

### LINKED ISSUE #318: Enforce project authorization on legacy v1 history reads
## Problem

Legacy v1 history endpoints require Cognito authentication but do not consistently enforce current project membership or validate that nested sprint, task, and execution identifiers belong to the project in the request path.

This can allow an authenticated user with retained or disclosed object identifiers to access another project's historical agent metadata, questions, or output. One GET status path also performs lazy status reconciliation, making authorization necessary before any read or write.

## Scope

- Enforce current project membership for the legacy agent history GET routes.
- Validate project-to-sprint and sprint-to-execution/task relationships.
- Authorize direct execution/question lookups from stored project ownership metadata.
- Return only fields required by the UI; do not expose internal callback tokens.
- Audit adjacent legacy v1 read handlers for the same missing tenant boundary.

## Acceptance criteria

- Non-members receive 403 before protected project data is returned or mutated.
- A sprint from another project cannot be supplied under an authorized project path.
- Direct execution and agent-question reads derive and enforce project ownership.
- Cross-project regression tests cover the affected routes.
- Existing project members retain read access to v1 history.

JWThewes:
## Summary

- enforce current project membership across legacy v1 history reads
- bind nested sprint and entity identifiers to their parent project or sprint
- derive project ownership for direct agent output and question lookups
- authorize before lazy status reconciliation writes
- allowlist agent-question response fields so internal callback tokens are never returned

## Security impact

Prevents authenticated users from reading or triggering history reconciliation for projects where they are not current members. It also closes equivalent authorization gaps in adjacent legacy sprint history handlers.

## Verification

- npm test: 111 files, 2,353 tests passed
- focused authorization tests: 7 files, 60 tests passed
- npm run lint
- npm run format:check
- production bundles built for all affected Lambdas

Closes #318

### DIFF
--- a/lambda/agents/index.js
+++ b/lambda/agents/index.js
@@ -23,6 +23,7 @@ import {
   InvokeAgentRuntimeCommand,
 } from '@aws-sdk/client-bedrock-agentcore';
 import gremlin from 'gremlin';
+import { PartitionStrategy } from 'gremlin/lib/process/traversal-strategy.js';
 import { fromNodeProviderChain } from '@aws-sdk/credential-providers';
 import { getUrlAndHeaders } from 'gremlin-aws-sigv4/lib/utils.js';
 import { buildResponse } from '../shared/response.js';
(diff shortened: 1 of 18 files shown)


jeromevdl (approved):
good for me

PR #321, opened 2026-07-17
fix: least privilege

jeromevdl:
Changes:

- Remove wildcard resource fallbacks in IAM policies for bedrock-agentcore and Lambda durable execution actions
- Require explicit ARN values rather than allowing fallback to "*" permissions
- Scope sensitive actions to specific resources instead of using wildcards

### DIFF
--- a/terraform/main.tf
+++ b/terraform/main.tf
@@ -36,7 +36,6 @@ provider "awscc" {
   region = var.aws_region
 }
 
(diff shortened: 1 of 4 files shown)


JWThewes (approved):
Looks good to me. I have not deployed it but I guess you deployed?

jeromevdl:
Deployed and tested yes (discussions ok, orchestrator ok)

PR #322, opened 2026-07-18
feat(intents): add intent attachments

jeromevdl:
*Description of changes:*

Add DRAFT-only compose attachments with browser uploads, progress, and collaborator-safe mutations.

Verify and version-pin staged S3 objects before promotion, reconcile ambiguous DynamoDB writes without deleting durable versions, and retain cleanup tombstones for retries.

Materialize committed references into read-only AgentCore workspaces and expose them as untrusted task context in every stage prompt.

Provision attachment routes, IAM permissions, S3 CORS, lifecycle cleanup, and intent/project deletion support.

_Tested on my instance with png and html_
image


terraform/modules/data/s3/main.tf
```diff
@@ -24,6 +24,21 @@ resource "aws_s3_bucket_public_access_block" "artifacts" {
   restrict_public_buckets = true
 }
 
+# Browser attachment uploads use presigned POSTs directly to this bucket. Keep
+# the existing artifact read/PUT methods while adding POST; object access stays
+# private because this only controls browser CORS preflight responses.
+resource "aws_s3_bucket_cors_configuration" "artifacts" {
```
JWThewes: The artifacts bucket already has aws_s3_bucket_cors_configuration.artifacts in terraform/main.tf. 

 S3 has one CORS configuration per bucket, so these resources will overwrite each other, cause persistent Terraform drift, and may remove POST, breaking attachment uploads.

JWThewes:
There’s a collaboration race that can block starting an intent for five minutes.

 If user A reserves an upload, then user B changes the attachments before A commits, A’s commit gets a 409 due to the stale revision. However, A’s pending reservation remains active, counts toward the limits, and causes /start to reject the intent. The UI discards the upload state and offers no retry or cancel path, so everyone must wait for the 300-second TTL.

Could we retry the commit with the latest revision after a 409, and add a way to cancel pending reservations when an upload or retry fails?

JWThewes (changes requested):
Some small comments. Otherwise looks good!

jeromevdl:
Implemented the suggested simplification by removing the client-side attachment commit flow entirely.

The new flow is:

> - browser reserves upload → browser uploads directly to S3 → S3 Object Created
>   → EventBridge → intents Lambda validates and attaches the file
> 
> - The browser no longer has to coordinate a second commit request, retry a stale attachment revision, or cancel reservations. It polls briefly for the attachment to appear after upload.
> 
> - The Lambda now owns the promotion: it verifies the reserved object’s type, size, and version, copies that exact version into the durable attachment prefix, then conditionally updates the intent metadata. Concurrent attachment events retry safely on revision conflicts and remove any losing copied version.
> 
> - A DRAFT intent cannot start while an active upload reservation remains, but abandoned reservations expire normally, so this does not create a permanent lock. Failed asynchronous processing is sent to the attachment DLQ.

I also fixed the S3 CORS configuration so there is a single bucket CORS resource, including POST for the presigned browser upload.

Added a focused handler test covering a concurrent attachment update: it verifies the conflict retry, rollback of the losing S3 copy, and preservation of both attachments.

JWThewes (approved):

PR #333, opened 2026-07-21
feat: project source-control bindings for member intent starts

### LINKED ISSUE #332: Members cannot start intents without a personal GitHub/GitLab connection
## Problem

Project members can only start an intent if they have personally connected GitHub (OAuth or App) or GitLab. For an unconnected member the intent creates and starts (201/202), but the run fails asynchronously:

- The v2 orchestrator resolves the git token from the **starting user's** personal connection (`defaultResolveToken`) and gets `no_connection`.
- `init-ws` then hard-fails with `checkout_failed` (private repo) or `intent_branch_push_failed` (public repo), surfacing as `init_ws_failed` / intent FAILED.

Additionally, raw git tokens are snapshotted into AgentCore invocation payloads and durable execution history, which has no delete API and retains for 90 days.

## Proposal

Replace per-user runtime auth with **project-level source-control bindings**:

- A project owner/admin binds each repository once, choosing the auth type: `github-app` (installation discovered per repo), `github-oauth`, or `gitlab-oauth` (explicitly confirmed delegation of the binder's connection).
- Bindings live in a DynamoDB table holding only opaque credential references — never tokens.
- Intent starts validate bindings up front (409 `SOURCE_CONTROL_NOT_READY` with per-repo reasons instead of async failure).
- The AgentCore runtime becomes token-free: git operations fetch short-lived credentials from a dedicated credential-broker lambda (IAM-scoped to the runtime role, validates execution liveness) via a throwaway `GIT_ASKPASS` helper; provider API operations (PRs, issues, branches) go through a source-control service lambda that never returns tokens.
- Bindings are invalidated automatically when the delegator disconnects, loses scopes, or is removed from the project.

This lets any member start intents on GitHub OAuth, GitHub App, and GitLab OAuth projects, and removes token material from durable history.

JWThewes:
Closes #332

## Problem

Members could not start intents without a personal GitHub/GitLab connection. The intent created and started fine (201/202), but the run failed asynchronously: the orchestrator resolved the **starting user's** personal token (`no_connection`), and init-ws hard-failed with `checkout_failed` / `intent_branch_push_failed`. Raw git tokens were also snapshotted into AgentCore payloads and durable execution history (no delete API, 90-day retention).

## Approach

Per-user runtime auth is replaced with **project-level source-control bindings**:

- **Bind once, per project** — an owner/admin chooses the auth type at step 1 of the create-space flow: **GitHub App** (installation discovered per repo; no personal connection needed — repo discovery runs on App-JWT-minted metadata-read installation tokens via new `/github/app/status|repos` routes), or explicitly confirmed **GitHub OAuth / GitLab OAuth** delegation of the binder's connection. Bindings live in a new DynamoDB table holding only opaque credential references, never tokens, and auto-invalidate when the delegator disconnects, loses scopes, or leaves the project.
- **Fail fast** — intent starts validate bindings up front and return `409 SOURCE_CONTROL_NOT_READY` with per-repo reasons instead of failing mid-run.
- **Token-free runtime** — AgentCore git operations obtain short-lived credentials from a dedicated credential-broker lambda (IAM-scoped to the runtime role; validates execution liveness and repo membership) through a throwaway `GIT_ASKPASS` helper. Tokens no longer enter invocation payloads or durable history.
- **One provider surface** — provider API operations (branches, trees, PRs, issues) route through a source-control service lambda that never returns tokens; the per-provider API proxy routes and the platform-wide GitHub auth mode are removed.

## Testing

- Backend: all workspace suites green (intents, source-control, credential-broker, agentcore, orchestrator, trackers, shared, github/gitlab).
- Frontend: full vitest suite green (54 files / 397 tests) incl. new modal tests for the App path without a personal connection and the unconfigured-App fallback; `tsc -b` clean.
- Terraform: `validate` + `fmt -check -recursive` clean (backend-config init errors pre-exist).

## Notes for reviewers

- The create-space flow now offers **GitHub App vs OAuth at step 1** when both are configured; the App path works end to end for users who never connected GitHub.
- A failed bind leaves the project **unbound** (launch guard blocks starts) instead of deleting it.
- Incident response: invalidate the affected bindings / revoke the token at the provider — unbound/invalid bindings block repository-backed starts. There is deliberately no global kill switch.


jeromevdl (changes requested):
### Gitlab token refresh
When a GitLab access token is near expiry, two git operations can try to refresh it at the same time.
`lambda/shared/git-token.js:125` reads the stored token and refreshes it, but does not prevent another request from doing the same. GitLab replaces the refresh token on a successful refresh. That means the first request succeeds and saves the new token pair, while the second request uses the now-invalid old refresh token and gets invalid_grant.

This PR can trigger that situation because it asks the credential broker for credentials separately for each git operation (`lambda/agentcore/git-auth.js:42`). Parallel construction lanes can therefore refresh the same GitLab connection simultaneously. The failed request is treated as a broken credential and invalidates the project’s source-control binding (`lambda/shared/source-control-bindings.js:281`), blocking repository work until an owner rebinds it.

Please make refresh single-flight per GitLab connection, or re-read the stored token when a refresh fails before marking the binding invalid. Add a test that issues two credential-broker requests concurrently for an expired GitLab token.

### Project-bound GitLab API operations lost the existing 401 refresh-and-retry behavior.
The GitLab provider retries only when ctx.onRefresh is provided (`lambda/shared/git-providers/gitlab.js:54`). The new source-control path invokes providers with only `{ token }` (`lambda/source-control/index.js:365`), so an early/clock-skew/revoked access-token 401 is not refreshed. Instead it invalidates the binding as `provider_unauthorized` (`lambda/shared/source-control-bindings.js:277`). The old personal GitLab handler still has `onRefresh`; the PR’s project-bound path does not.

### Validation
✅ Tested with gitlab end-to-end, clone => branch => PR all worked, didn't face any of the edge case above

JWThewes:
@jeromevdl I addressed your comments and also fixed the GH App issue. after pulling this, the previously-failed binding should verify on the next "Bind and verify" click there's no config change needed on the GitHub App side

jeromevdl (approved):
OK, comments fixed and binding verified

PR #335, opened 2026-07-21
fix: make durable PR waits callback-driven

### LINKED ISSUE #334: Prevent durable PR waits from exhausting execution operation limits
## Problem

PR-per-unit executions currently poll pull request state with durable timers. A long review wait can therefore consume the durable execution's operation budget even though no application work is happening.

The observed execution reached Lambda's 3,000-operation limit and failed with `ExecutionLimitExceeded`, while its process row remained `RUNNING`. During that wait:

- the provider had already merged the reviewed `comparison-feature` head, but local state still showed the unit and PR as ready;
- the completed `views` construction stages and draft PR were still valid;
- the watchdog would not notice the terminal durable execution until the configured local expiry;
- the existing repair path would reset completed stages and replay successful construction.

## Expected behavior

- A PR review wait persists one callback and does not grow durable operations while provider state is unchanged.
- Provider reconciliation wakes that callback only for merge, closure, reviewed head/target movement, or queued authenticated feedback.
- Concurrent scheduler and feedback wakeups deliver a callback at most once.
- Terminal durable executions are detected while their local process rows are still active.
- Repair reconciles remote PR state, validates merged reviewed-head ancestry, preserves completed stages and artifacts, and resumes an existing draft PR through the current orchestrator alias.
- GitHub, GitLab, multi-repository, and stale-run ownership cases behave consistently.

## Proposed implementation

- Add a sparse DynamoDB maintenance index for active executions and parked unit PR waits.
- Replace durable timer polling with persisted callback ownership and an EventBridge-driven provider reconciler.
- Extend the durable watchdog to inspect every indexed active execution.
- Make lane repair provider-aware and state-preserving.
- Add focused regression coverage for long waits, transitions, feedback wakeups, duplicate delivery, early durable failure, and recovery.

## Dependency

This work uses the project-level source-control service and authentication model introduced by #333.

## Acceptance criteria

- Long unchanged PR waits create no repeated durable operations.
- A merged reviewed head is marked merged only after ancestry validation.
- Completed construction stages retain their state, attempts, and artifacts during repair.
- An existing draft PR resumes without creating a replacement or rerunning construction.
- The repaired execution advances after that PR merges.

JWThewes:
Closes #334

## Problem

PR-per-unit review used durable timer polling. One production-like execution exhausted Lambda's 3,000-operation limit while waiting for review, leaving process state active after the durable execution had failed.

The affected intent also demonstrated why repair must preserve state: one reviewed unit PR had merged remotely while local state lagged, and another unit's construction stages and draft PR were already complete and reusable.

## Approach

- Replace repeated durable PR polling with one persisted callback per observed provider state.
- Add sparse DynamoDB maintenance partitions for active executions and parked PR waits.
- Reconcile GitHub/GitLab state on a one-minute schedule and wake only for merge, closure, branch movement, or queued authenticated feedback.
- Atomically claim callback delivery so scheduler and feedback wakeups cannot double-resume a run.
- Detect terminal durable executions before their configured local expiry without overwriting a concurrently completed process row.
- Reconcile provider truth during repair, validate merged reviewed-head ancestry, preserve successful stages and artifacts, and resume existing draft PRs through the current orchestrator alias.

## Recovery verification

The deployed dev recovery was exercised against the failed execution that motivated #334:

- the prior durable execution was confirmed `FAILED` with `ExecutionLimitExceeded`;
- the merged reviewed head was confirmed reachable from the intent branch and the corresponding unit was marked merged;
- ten completed construction stage rows remained `SUCCEEDED` at attempt 0 and no artifacts were archived;
- the existing draft PR was reconciled in place, promoted to ready, and parked on exactly one persisted callback in the sparse `PR_WAITS` index.

The review PR was deliberately left unmerged.

## Testing

- `npm test` - 117 files, 2,331 tests passed
- focused shared/orchestrator/intents suites - 44 files, 887 tests passed
- commit hook - formatting, lint, secret scan, production dependency audits, and 561 targeted tests passed
- affected `intents` and `v2-orchestrator` Lambda bundles built successfully
- `terraform fmt -check -recursive terraform`

### DIFF
--- a/lambda/intents/index.js
+++ b/lambda/intents/index.js
@@ -1198,11 +1198,14 @@ export const handler = async (event) => {
     await ingestAttachmentUpload(event);
     return { ok: true };
   }
+  if (event?.action === 'repair-durable-executions') {
+    return runDurableExecutionWatchdog(event);
+  }
   if (
-    event?.action === 'repair-durable-executions' ||
+    event?.action === 'reconcile-provider-state' ||
     (event?.source === 'aws.events' && event?.['detail-type'] === 'Scheduled Event')
   ) {
-    return runDurableExecutionWatchdog(event);
+    return runScheduledMaintenance(event);
   }
   if (event.httpMethod === 'OPTIONS') return response(200, {});
 
@@ -1402,6 +1405,15 @@ export const handler = async (event) => {
             state: 'QUEUED',
             summary,
           }).catch(() => {});
+          const wait = await store.getUnit(intentId, sectionIndex, unitSlug).catch(() => null);
+          if (wait?.prWaitCallbackId) {
+            await wakeUnitPrWait(wait, {
+              reason: 'queued_feedback',
+              detail: { batchId, commentCount: selected.length },
+            }).catch((error) =>
+              console.error('Queued feedback PR-wait wake failed:', error.message),
+            );
+          }
         }
         return response(created.created ? 202 : 200, mapFeedbackBatch(created.item));
       }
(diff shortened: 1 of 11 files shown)


lambda/intents/index.js
```diff
@@ -4282,37 +4481,178 @@ const runDurableExecutionWatchdog = async ({ candidates: injectedCandidates = nu
...
+          const provider =
+            pr.provider || meta.repoProviders?.[pr.repository] || meta.gitProvider || 'github';
+          const status = await sourceControlOperation({
+            projectId: meta.projectId,
+            provider,
+            repo: pr.repository,
+            operation: 'pr-status',
+            args: { number: pr.number },
+          });
+          const observation = normalizePrSnapshot(pr, status);
```
eipasteur, line 4628: `sourceControlOperation` returns `response.result` directly, and both GitHub and GitLab providers return `null` on non-2xx responses (for example, a deleted PR or transient 5xx). That `null` can reach `normalizePrSnapshot(pr, status)`, where `status.state` throws a `TypeError`.

The surrounding `try/catch` records the error and continues, so this degrades rather than crashes. However, the wait remains parked and retries with the same error every minute.

Could we explicitly handle a missing status before normalizing it and add a test for `{ ok: true, result: null }`, ideally with an actionable error such as `PR status unavailable`?

JWThewes: Addressed in dbd025e. The reconciler now checks for a missing or invalid PR status before normalization and reports `PR status unavailable for <repository>#<number>`, while leaving the wait parked for a later retry. I also added regression coverage for an `{ ok: true, result: null }` response, including assertions that the callback is not invoked and the wait is not cleared. Verified with the full intents suite (201 tests), oxlint, and oxfmt.

terraform/modules/api/lambda/main.tf
```diff
@@ -2098,15 +2098,15 @@ module "intents_lambda" {
 
 resource "aws_cloudwatch_event_rule" "intents_durable_watchdog" {
   name                = "${var.project_name}-intents-durable-watchdog-${var.environment}"
-  description         = "Repair v2 intents whose durable orchestrator expired while process state remained active"
-  schedule_expression = "rate(1 day)"
+  description         = "Reconcile parked PR waits and detect terminal durable orchestrators"
+  schedule_expression = "rate(1 minute)"
```
eipasteur, line 2102: Non-blocking: The maintenance rule now runs every minute, while the Lambda timeout remains 120 seconds. `runPrWaitReconciler` and the watchdog execute sequentially, with up to 100 items each; a slow provider response can therefore make one pass overlap the next scheduled invocation.

The DynamoDB claims make that overlap safe, so this is primarily about wasted Lambda concurrency and provider/API quota. Could we either bound the reconciler\u2019s per-item concurrency or expose `PR_WAIT_RECONCILER_LIMIT` and `DURABLE_WATCHDOG_LIMIT` as Terraform-configurable environment variables so deployments can tune the workload without changing code?

eipasteur (approved):

PR #341, opened 2026-07-22
Fix(frontend): UI improvements

jeromevdl:
*Issue #, if available:*

*Description of changes:*
- Simplified review-stage controls and hid the side panel while creating an intent.
- Improved work products with persistent chronological/reverse ordering across phases, stages, and documents.
- Made review identified items actionable with previews and item-level discussions.
- Improved intent tracker labels and added links between referenced work artifacts in previews.
- Hide generated units while still in inception, only show at construction and after

*Validation:*
- Deployed and tested 
- Git pre commit hook

### DIFF
--- a/frontend/src/components/intent/UnitLaneBoard.tsx
+++ b/frontend/src/components/intent/UnitLaneBoard.tsx
@@ -630,9 +630,8 @@ export function isBeforeConstructionPhase(
   workflowPhases: PhaseNode[] | null,
 ): boolean {
   if (!currentPhasePath || !workflowPhases) return false;
-  const currentPhase = workflowPhases.find((phase) => phase.path === currentPhasePath);
   const constructionPhase = workflowPhases.find((phase) => phase.phaseId === 'construction');
-  return Boolean(currentPhase && constructionPhase && currentPhase.order < constructionPhase.order);
+  return Boolean(constructionPhase && currentPhasePath < constructionPhase.path);
 }
 
 export function UnitLaneBoard({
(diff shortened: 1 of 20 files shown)


JWThewes:
I have to review in detail but one thing: we wanted to add discussion to the intent drafting. If the sidebar is gone we have no place to show that anymore, right? Any idea about this @jeromevdl

JWThewes:
Ok it's just for the intent create page not the draft page. So that should work

jeromevdl:
@JWThewes yes, it is only at creation. Then in compose it is available.

I've added another small fix to hide units while in inception.

frontend/src/components/intent/UnitLaneBoard.tsx
```diff
@@ -624,10 +625,29 @@ export function isFanoutActive(detail: { events?: { type: string }[] } | null):
...
 }
 
+export function isBeforeConstructionPhase(
+  currentPhasePath: string | null,
+  workflowPhases: PhaseNode[] | null,
+): boolean {
+  if (!currentPhasePath || !workflowPhases) return false;
+  const currentPhase = workflowPhases.find((phase) => phase.path === currentPhasePath);
+  const constructionPhase = workflowPhases.find((phase) => phase.phaseId === 'construction');
+  return Boolean(currentPhase && constructionPhase && currentPhase.order < constructionPhase.order);
```
JWThewes: `order` is only the position among sibling phases, not the global workflow order. For a valid nested tree with inception at `01.02` and construction at `02`, both have `order === 2`, so this returns false and the unit board becomes visible during inception. The inverse can also hide it after construction. Could we compare the zero-padded `path` values instead (as `ProvenanceTree` does), and add a nested-phase test that covers both sides of construction?

JWThewes (dismissed):

frontend/src/components/intent/UnitLaneBoard.tsx
jeromevdl: fixed

JWThewes (approved):

PR #342, opened 2026-07-22
feat(bitbucket): add Bitbucket Cloud as a git provider (v2/AgentCore)

smoell:
End-to-end Bitbucket Cloud support across provider, Lambda, Terraform and frontend: OAuth connect, repo listing (/user/workspaces, CHANGE-2770), recursive getTree, just-in-time token refresh (both dispatch paths), SSM Intelligent-Tiering for JWT tokens, and automated PR creation at fan-in. Validated live: multi-unit intent -> 3 units/3 waves merged -> automated Bitbucket PR.

*Issue #, if available:*

*Description of changes:*


frontend/src/components/CreateProjectModal.tsx
jeromevdl: Bitbucket needs its own `bitbucket-oauth` auth type here and throughout the source-control binding contract. Please add it to `AUTH_TYPES`/`AUTH_TYPE_PROVIDER` and canonical repository support in `lambda/shared/source-control-bindings.js`, and to the frontend `SourceControlAuthType`. 

Binding verification also calls `getRepositoryAccess` and `getAuthenticatedUser`; the Bitbucket adapter must implement them before Bitbucket projects can be created.

jeromevdl:
`lambda/shared/source-control-bindings.js` was not updated following the last merge

jeromevdl:
`frontend/src/services/sourceControl.ts` was not updated following the last merge

jeromevdl:
`lambda/shared/source-control-credentials.js` was not updated following the last merge

lambda/shared/git-providers/bitbucket.js
jeromevdl, line 1: Review this file to implement the binding changes from last merge (`getRepositoryAccess` and `getAuthenticatedUser`). 

Please check https://github.com/aws-samples/sample-collaborative-ai-dlc/tree/pr-342 (8ff56d8b4e3245b75b51354a0b9fb206409e4f22)

lambda/trackers/index.js
jeromevdl, line 1: https://support.atlassian.com/bitbucket-cloud/docs/understand-bitbucket-issues/ => bitbucket issues will be removed from bitbucket cloud in August 2026 and users are encouraged to move to Jira. We can also remove the tracker in collaborative aidlc.

Also see https://github.com/aws-samples/sample-collaborative-ai-dlc/tree/pr-342 for this (460c5f7533f8c0a2498c8e672f234d56b95cfd06) .

terraform/main.tf
```diff
@@ -341,7 +346,12 @@ module "agentcore" {
   websocket_execution_arn     = module.realtime.websocket_execution_arn
   aidlc_repo_ref              = var.aidlc_repo_ref
   bedrock_model               = var.bedrock_model
-  kiro_model                  = "claude-opus-4.6"
+  # The Kiro CLI shipped in the AgentCore image only accepts the model id
+  # "auto" (it maps to Kiro's own default) — a Bedrock-style id like
+  # "claude-opus-4.6" is rejected with "Model '…' does not exist. Available
+  # models: auto" and fails the stage. Seed "auto"; override per-CLI at runtime
+  # via Admin → Agent Settings (the cli-models SSM param is ignore_changes).
+  kiro_model = "auto"
```
jeromevdl: Unrelated change. Also I'm using Kiro in Collaborative AIDLC with current config without error. Can we remove this from this PR.

jeromevdl:
`lambda/shared/source-control-credentials.js:249`:

Bitbucket Git-over-HTTPS requires `x-token-auth`, not `x-access-token`. Please also include Bitbucket in the 401 refresh callback immediately below; otherwise a rejected-but-not-yet-expired access token cannot recover.

lambda/shared/git-providers/bitbucket.js
```diff
@@ -0,0 +1,952 @@
...
+};
+
+// ---------------------------------------------------------------------------
+// OAuth
+// ---------------------------------------------------------------------------
+
+const oauth = {
+  secretEnvName: 'BITBUCKET_OAUTH_SECRET_NAME',
+  redirectUriEnvName: 'BITBUCKET_REDIRECT_URI',
+  scopes: 'account repository repository:write pullrequest pullrequest:write',
```
jeromevdl: Please request and require the `email` scope. The Bitbucket binding implementation needs the authenticated user’s email for commit attribution; `account` alone does not grant it.

frontend/src/components/admin/tabs/TrackersTab.tsx
```diff
@@ -22,7 +22,7 @@ interface Props {
 
 // Trackers that piggyback on a source-control OAuth app instead of having
 // credentials of their own.
-const GIT_BACKED_TRACKERS = new Set(['github-issues', 'gitlab-issues']);
+const GIT_BACKED_TRACKERS = new Set(['github-issues', 'gitlab-issues', 'bitbucket-issues']);
```
jeromevdl, line 25: https://support.atlassian.com/bitbucket-cloud/docs/understand-bitbucket-issues/ => bitbucket issues will be removed from bitbucket cloud in August 2026 and users are encouraged to move to Jira. We can also remove the tracker in collaborative aidlc.

Also see https://github.com/aws-samples/sample-collaborative-ai-dlc/tree/pr-342 for this (https://github.com/aws-samples/sample-collaborative-ai-dlc/commit/460c5f7533f8c0a2498c8e672f234d56b95cfd06) .

frontend/src/components/CreateProjectModal.tsx
jeromevdl: Please define each provider’s supported source-control auth options and default in `frontend/src/services/sourceControl.ts`, then consume that metadata here. This keeps the modal and source-control contract aligned.

terraform/modules/api/lambda/main.tf
```diff
@@ -1456,6 +1463,83 @@ module "gitlab_lambda" {
...
+
+resource "aws_iam_role_policy" "bitbucket_connector" {
+  name = "bitbucket-oauth-and-token-storage"
+  role = aws_iam_role.bitbucket_connector.id
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Effect   = "Allow"
+        Action   = ["dynamodb:GetItem", "dynamodb:PutItem", "dynamodb:DeleteItem"]
```
jeromevdl: Once Bitbucket project bindings exist, this role also needs the bindings-table permissions (`Query` and `UpdateItem`, including its index) and `SOURCE_CONTROL_BINDINGS_TABLE` environment variable. 
git-handler invalidates bindings on disconnect, missing scopes, and refresh failure; this role currently has no access to do so. Mirror the GitLab connector role.

frontend/src/components/admin/tabs/SourceControlTab.tsx
```diff
@@ -33,6 +33,7 @@ export function SourceControlTab({ providers, providersLoading, onProvidersChang
   }
 
   const gitlabConfigured = isConfigured(providers, 'gitlab-issues');
+  const bitbucketConfigured = isConfigured(providers, 'bitbucket-issues');
```
jeromevdl, line 36: If we remove issues, we should have another way to configure bitbucket oauth. (gitlab is using the tracker config, which is not a good idea, we should have a dedicated endpoint, see a2f11d5fa52804aed94a1a91415aa303de694dd2)

lambda/shared/git-providers/bitbucket.js
```diff
@@ -0,0 +1,952 @@
...
+  }
+
+  return files;
+};
+
+const getFileContents = async (ctx, repoId, filePath, branch = 'main') => {
+  const { workspace, repoSlug } = splitWorkspaceRepo(repoId);
+  const res = await bbFetch(
+    ctx,
+    `${API_BASE}/repositories/${workspace}/${repoSlug}/src/${encodeURIComponent(branch)}/${encodeURIComponent(filePath)}`,
```
JWThewes: `encodeURIComponent(filePath)` encodes the `/` separators too, so a nested file like `src/App.java` becomes `src%2FApp.java` and the request 404s. Only top-level files work right now. Encode per segment instead:

```js
filePath.split('/').map(encodeURIComponent).join('/')
```

(`branchWebUrl` in `frontend/src/components/intent/CodeSection.tsx` already does it this way.) Same question for `encodeURIComponent(branch)` here and in the other endpoints — worth testing with a `feature/x` style branch name, since user branches commonly contain slashes.

lambda/shared/git-providers/bitbucket.js
```diff
@@ -0,0 +1,952 @@
...
+        method: 'POST',
+        body: JSON.stringify({ type: 'merge' }),
+      },
+    );
+
+    if (mergeRes.ok) return 'merged';
+
+    // Handle merge conflicts
+    if ([400, 409, 422].includes(mergeRes.status)) {
+      return 'conflict';
```
JWThewes, line 1017: When the merge fails with a conflict we return `'conflict'` but leave the temporary PR we just created open. Every conflicted engine merge will litter the repo with an open "Merge X into Y (auto)" PR. Please decline the temp PR (`POST /pullrequests/{id}/decline`) before returning — same for the error path below.

lambda/shared/git-providers/bitbucket.js
```diff
@@ -0,0 +1,952 @@
...
+const constructionBranchPrefix = (branch) => `${branch}--task-`;
+
+const listConstructionTaskBranches = async (ctx, repoId, branch) => {
+  const { workspace, repoSlug } = splitWorkspaceRepo(repoId);
+  const prefix = constructionBranchPrefix(branch);
+
+  // Bitbucket doesn't have a direct search, so we get all branches and filter
+  const res = await bbFetch(
+    ctx,
+    `${API_BASE}/repositories/${workspace}/${repoSlug}/refs/branches?pagelen=100`,
```
JWThewes: This only reads the first page (`pagelen=100`) and doesn't follow `data.next`, unlike `getTree` above. Since `getUnmergedConstructionTaskBranches` gates fan-in PR creation on this list, a task branch that lands beyond page 1 would silently escape the merged-completeness check. Please paginate here (and in `listBranches`) the same way `getTree` does.

lambda/shared/git-providers/bitbucket.js
```diff
@@ -0,0 +1,952 @@
...
+    throw new Error(`Failed to compare ${sourceBranch} against ${targetBranch}: ${errorText}`);
+  }
+  const data = await res.json();
+
+  // If there are no commits in source that aren't in target, it's merged
+  return !Array.isArray(data.values) || data.values.length === 0;
+};
+
+// Real ancestor check — page through GET /commits/{descendantRef} until we find ancestorSha
+const isCommitAncestor = async (ctx, repoId, ancestorSha, descendantRef) => {
```
JWThewes, line 637: This pages through the entire commit history of `descendantRef` looking for one hash — O(history) API calls inside a 30s Lambda, so it'll time out on long-lived repos. Bitbucket has a merge-base endpoint (`GET /2.0/repositories/{workspace}/{repo_slug}/merge-base/{ancestorSha}..{descendantRef}`) that answers this in one call: if the merge-base equals `ancestorSha`, it's an ancestor. Same concern applies to `isBranchMergedInto` above.

lambda/shared/git-providers/bitbucket.js
```diff
@@ -0,0 +1,952 @@
...
+const listPRComments = async (ctx, repoId, prId) => {
+  const { workspace, repoSlug } = splitWorkspaceRepo(repoId);
+  const [generalComments, inlineComments] = await Promise.all([
+    listPaginated(
+      ctx,
+      `${API_BASE}/repositories/${workspace}/${repoSlug}/pullrequests/${prId}/comments?pagelen=100`,
+    ),
+    listPaginated(
+      ctx,
+      `${API_BASE}/repositories/${workspace}/${repoSlug}/pullrequests/${prId}/comments?pagelen=100&q=inline!=null`,
```
JWThewes: This fetches the full comment list twice — once unfiltered and once with `q=inline!=null` — and then filters both in memory anyway. The first request already returns every comment (inline ones included), so one pass partitioned by `c.inline` gives the same result at half the API calls:

```js
const all = await listPaginated(ctx, `...comments?pagelen=100`);
const general = all.filter((c) => !c.inline).map(mapComment);
const inline = all.filter((c) => c.inline).map(mapInlineComment);
```

lambda/shared/git-providers/bitbucket.js
```diff
@@ -0,0 +1,952 @@
...
+// ---------------------------------------------------------------------------
+// PR comments
+// ---------------------------------------------------------------------------
+
+const mapComment = (c) => ({
+  id: c.id,
+  type: 'issue',
+  body: c.content?.raw || '',
+  user: {
+    login: c.user?.username || c.user?.nickname || '',
```
JWThewes: `user.username` was removed from Bitbucket Cloud API payloads back in 2019 (GDPR changes), so this first branch is always undefined and we always fall through to `nickname`. Suggest dropping the dead field and using `nickname || display_name`. Also, `bot: user.type === 'team'` flags workspace/team accounts, not bots — probably better to just hardcode `bot: false` than mislabel org accounts.

JWThewes (commented):
A few additional findings on top of Jerome's review — mostly in the provider module. Details inline.

lambda/shared/git-providers/bitbucket.js
```diff
@@ -0,0 +1,952 @@
...
+    mergeable: null, // Bitbucket doesn't expose this directly
+    mergeableState: null,
+    mergedAt: pr.state === 'MERGED' ? pr.updated_on : null,
+    closedAt: pr.state === 'DECLINED' ? pr.updated_on : null,
+    updatedAt: pr.updated_on ?? null,
+    title: pr.title ?? '',
+  };
+};
+
+const setPullRequestDraft = async (ctx, repoId, prId, draft) => {
```
JWThewes, line 921: Bitbucket Cloud PRs have a native `draft` boolean these days (both in the PR payload and settable via PUT), so the `Draft:` title-prefix emulation shouldn't be needed — and it's fragile: a user editing the title flips our notion of draft state. Worth checking whether the native field covers this; if it does, `getPullRequestStatus`'s `draft` detection should read it too.

terraform/modules/api/lambda/main.tf
```diff
@@ -2301,11 +2449,36 @@ module "v2_orchestrator_lambda" {
...
+    V2_PROCESS_TABLE      = var.v2_executions_table_name
+    BLOCKS_TABLE          = var.blocks_table_name
+    AGENTCORE_RUNTIME_ARN = var.agentcore_runtime_arn
+    GIT_TOKEN_SSM_PREFIX  = "${var.project_name}/${var.environment}/git-token"
+    # Git-token resolution: resolveToken reads the starter's
+    # connection row from these tables, then the SSM token. The IAM statements
+    # were wired from day one but these env vars were MISSING — getGitConnection
+    # threw on TableName undefined, resolveToken degraded to an empty token, and
+    # a private-repo clone silently fell back to `git init` (see init-ws).
+    GIT_CONNECTIONS_TABLE          = var.git_connections_table_name
```
JWThewes: These env-var additions (GIT_CONNECTIONS_TABLE, the GitLab secret, GitHub App params) fix a real pre-existing GitLab/GitHub App bug per the comments — but that makes them an unrelated fix riding in a 2,200-line Bitbucket PR, same as the `kiro_model` change Jerome flagged. Could you split the GitLab/GitHub-App orchestrator fix into its own PR? It deserves review on its own terms, and it shouldn't get reverted if this PR needs rework.

terraform/modules/api/lambda/main.tf
```diff
@@ -1499,8 +1583,13 @@ module "trackers_lambda" {
     GITLAB_REDIRECT_URI            = var.gitlab_redirect_uri
     SOURCE_CONTROL_FUNCTION        = module.source_control_lambda.lambda_function_name
     SOURCE_CONTROL_BINDINGS_TABLE  = var.source_control_bindings_table_name
-    ENVIRONMENT                    = var.environment
-    CORS_ALLOWED_ORIGINS           = var.cors_allowed_origins
+    BITBUCKET_OAUTH_SECRET_NAME    = var.bitbucket_oauth_secret_name
+    BITBUCKET_REDIRECT_URI         = var.bitbucket_redirect_uri
+    # GitHub App auth (github-issues app-mode config)
+    GITHUB_APP_CONFIG_PARAM            = var.github_app_config_param_name
```
JWThewes: The trackers Lambda gains `GITHUB_APP_CONFIG_PARAM` / `GITHUB_APP_PRIVATE_KEY_SECRET_NAME` here, but I don't see a matching IAM statement added to the trackers role in this PR, and there's no Bitbucket connection either — the trackers code path doesn't read these today as far as I can tell. Leftover from an experiment? If it's needed, it needs the IAM grant; otherwise please drop it.

lambda/shared/git-handler.js
```diff
@@ -89,6 +89,10 @@ export const createGitHandler = (provider, routes) => {
                 : {}),
             }),
             Type: 'SecureString',
+            // Bitbucket access/refresh tokens are JWTs (~2800+ chars) that
+            // exceed the SSM Standard-tier 4096-char cap. Intelligent-Tiering
+            // promotes to Advanced only when needed (no-op for short tokens).
+            Tier: 'Intelligent-Tiering',
```
JWThewes, line 95: Heads-up that this now applies to GitHub/GitLab token writes too, not just Bitbucket. Behaviorally it's a no-op under 4KB, but Intelligent-Tiering promotions to Advanced carry a per-parameter monthly cost — probably worth a one-line note in the PR description so operators aren't surprised.

terraform/main.tf
smoell: Reverted — this was environment-specific debugging that shouldn't ride on this PR. kiro_model is back to the upstream claude-opus-4.6 seed. Thanks for catching it.

terraform/modules/api/lambda/main.tf
smoell: Removed both GITHUB_APP_CONFIG_PARAM and GITHUB_APP_PRIVATE_KEY_SECRET_NAME from the trackers lambda. Confirmed the trackers handler never reads them and the trackers IAM role has no matching grant — they were dead env vars introduced by the merge. The legitimate GitHub-App wiring on the github_connector role is untouched.

lambda/shared/git-providers/bitbucket.js
smoell: Both implemented in bitbucket.js and exported. getAuthenticatedUser reads /user (login via nickname/display_name) plus /user/emails for the primary confirmed email (commit attribution), falling back to a …@users.noreply.bitbucket.org form when the email lookup is unavailable so attribution never blocks a run. getRepositoryAccess returns default branch + visibility from the repo object and maps the caller's permission from /user/permissions/repositories (admin/write/read), degrading to read-only if that probe fails. Tests added for both (write access + read-only fallback + noreply fallback).

frontend/src/components/CreateProjectModal.tsx
smoell: Both implemented in bitbucket.js and exported. getAuthenticatedUser reads /user (login via nickname/display_name) plus /user/emails for the primary confirmed email (commit attribution), falling back to a …@users.noreply.bitbucket.org form when the email lookup is unavailable so attribution never blocks a run. getRepositoryAccess returns default branch + visibility from the repo object and maps the caller's permission from /user/permissions/repositories (admin/write/read), degrading to read-only if that probe fails. Tests added for both (write access + read-only fallback + noreply fallback).

lambda/shared/git-providers/bitbucket.js
smoell: Added email to oauth.scopes and requiredConnectionScopes. This is what lets getAuthenticatedUser resolve the user's email for commit attribution. Note: existing Bitbucket connections will need to reconnect to pick up the new scope — I'll call that out in the docs (Block 4).

lambda/shared/git-providers/bitbucket.js
smoell: Fixed. `getFileContents` now encodes each path segment via `filePath.split('/').map(encodeURIComponent).join('/')`, so nested files like `src/main/App.java` keep their literal slashes instead of becoming `src%2Fmain%2FApp.java`. Test covers a nested path + a reserved char within a segment.

lambda/shared/git-providers/bitbucket.js
smoell: Fixed. `mergeBranch` now declines the temporary auto-merge PR (`POST …/pullrequests/{id}/decline`) on both the conflict and error paths before returning, so a failed engine merge no longer leaves an open "Merge X into Y (auto)" PR behind. Best-effort decline (a failed decline won't mask the merge outcome). Test asserts the decline fires on conflict.

lambda/shared/git-providers/bitbucket.js
smoell: Fixed. Both `listBranches` and `listConstructionTaskBranches` now follow `data.next` like `getTree`. Test proves a task branch on page 2 is included in `getUnmergedConstructionTaskBranches` (previously it would silently escape the fan-in merged-completeness check).

lambda/shared/git-providers/bitbucket.js
smoell: Fixed. `isCommitAncestor` now uses `GET …/merge-base/{ancestorSha}..{descendantRef}` — a single call (ancestor iff merge-base == ancestorSha), replacing the O(history) commit-log walk that could time out the 30s Lambda. Handles short-hash ancestors via prefix match. Re: `isBranchMergedInto` — it already uses a single `commits?exclude=…&pagelen=1` call (not an O(history) walk), so it isn't affected by the same timeout concern. Happy to switch it to merge-base too for symmetry if you'd prefer.

lambda/shared/git-providers/bitbucket.js
smoell: Fixed. `listPRComments` now makes one paginated pass partitioned by `c.inline`, dropping the redundant `q=inline!=null` request. Test asserts exactly one network call + correct issue/review partitioning.

lambda/shared/git-providers/bitbucket.js
smoell: Fixed. Author login now uses `nickname || display_name` (dropped `username` — GDPR-removed from Bitbucket Cloud since 2019, always undefined), and `bot` is now `false` rather than `type === 'team'` (which mislabeled workspace/org accounts). Applied to `mapComment`, `mapInlineComment`, and `addPRComment`.

lambda/shared/git-providers/bitbucket.js
smoell: You were right — switched to the native `draft` boolean. `createPullRequest` sends `draft` in the payload, `getPullRequestStatus` reads `pr.draft`, and `setPullRequestDraft` sets it via PUT (no title mangling). Also added the `(draft=true OR draft=false)` clause to `findPullRequest` so draft PRs aren't hidden by the List-PR endpoint ([[BCLOUD-23659](https://jira.atlassian.com/browse/BCLOUD-23659)](https://jira.atlassian.com/browse/BCLOUD-23659)) — closes the latent "draft unit-PR missed at fan-in" bug.

terraform/modules/api/lambda/main.tf
smoell: Done (ca021b6). The bitbucket_connector role now mirrors gitlab_connector: added dynamodb:UpdateItem + dynamodb:Query and the source_control_bindings_table ARN (+ /index/*), plus the SOURCE_CONTROL_BINDINGS_TABLE env var on the bitbucket lambda — so git-handler can invalidate bindings on disconnect / missing-scope / refresh failure.

lambda/trackers/index.js
smoell: Splitting this into its own follow-up PR rather than folding it into this one. Removing the bitbucket-issues tracker also requires a replacement admin surface for the Bitbucket OAuth-app credentials (new GET/PUT /bitbucket/oauth-config endpoint) and rewiring the Connect-button config gating, which currently keys off the tracker-provider status (/trackers/providers). That's a focused, self-contained change set that deserves its own review — I'll link it here.

JWThewes:
Thanks for the quick turnaround on the last round the adapter fixes (nested paths, pagination, merge-base, declining the temp PR, native draft flag) all look good and the tests you added cover them properly. 
That said, I tried to trace a Bitbucket space end-to-end and it doesn't get off the ground yet. The adapter is solid, but the integration layer around it has a few holes:
1. Bitbucket bindings are rejected by the API
lambda/source-control/index.js:483 still has ['github', 'gitlab']. You added bitbucket-oauth everywhere else in the binding contract, but this allowlist guards PUT /projects/{id}/source-control — so picking Bitbucket in the UI returns 400 Invalid source-control selection for bitbucket. No binding means no clone, no PR, no intent start. Everything else in this PR is currently unreachable through the product.
The existing test only asserts github/gitlab, which is why CI is green. Worth extending it so this can't happen again.
This also makes me think the "validated live" run in the description happened before the #333 merge — could you re-run it on the current head?
2. Two lambdas can't refresh Bitbucket tokens
source_control and credential_broker are both missing BITBUCKET_OAUTH_SECRET_NAME and the Secrets Manager grant for the Bitbucket secret (terraform/modules/api/lambda/main.tf, ~788/828 and ~883/920). You wired the v2-orchestrator correctly, so it looks like these two just got missed.
Bitbucket access tokens are short-lived, so the practical effect is that everything works right after connecting and then starts failing an hour later — clone, push, PR, branch listing, the lot. That's a nasty failure mode to debug in the field.
3. Bitbucket Issues is advertised but doesn't exist
I understand splitting the tracker removal into its own PR, but this PR still adds the claim: the tracker registry entry, the OAuth callback route, the admin card, GIT_BACKED_TRACKERS, and README/docs all say Bitbucket Issues is supported. There's no backend tracker provider and the adapter has no issue methods, so the first thing a user does after configuring it returns an error.
Either implement it or strip those surfaces from this PR — shipping a working config screen for a non-existent feature is worse than not mentioning it.
4. The email scope requirement never made it into the docs
You added email to the required scopes, but README, docs/getting-started/setup.md and the in-app setup instructions still list only Account / Repositories / Pull requests. Anyone following the docs creates a consumer without Email, and the status endpoint then reports "reauthorization required" and invalidates their bindings. Small fix, but it turns a documented happy path into a dead end.
While you're there — worth double-checking what Bitbucket actually returns for the granted scopes in the token response (I believe the field name differs from GitHub/GitLab). If we're reading the wrong key, the scope check fails for everyone regardless of how the consumer is configured. Nothing currently tests that path.
5. The orchestrator IAM/env block is still in here
Re: my earlier comment — terraform/modules/api/lambda/main.tf (~2378-2412 and ~2460-2500) still adds SSM/DynamoDB/Secrets Manager permissions plus seven env vars to v2_orchestrator, and a similar set to intents. I checked: none of them are read by either lambda post-#333, and the comment saying the IAM statements "were wired from day one" isn't accurate — this PR adds them (the pre-PR policy had five statements, none of these). Given we have a dedicated least-privilege pass in the history, I'd rather not widen these roles for code that doesn't use them. Please drop the block.
Smaller things, happy to take as follow-ups:
- Bitbucket rotates refresh tokens on use, same as GitLab, but refreshBitbucketToken doesn't have GitLab's recovery path for two concurrent refreshes racing each other. Parallel construction lanes will hit this. It also throws a plain Error instead of the classified one, so bindings never get marked invalid and the user sees a permanently-retrying 502 rather than "reconnect Bitbucket".
- lambda/bitbucket has no test/ directory, so npm test -w bitbucket-lambda fails outright, and the bundle smoke test doesn't cover the new lambda (github and gitlab both do).
- The orchestrator's PR lookup normalizes state and reads the PR id in a provider-specific way that doesn't account for Bitbucket, so existing draft unit PRs never get reused at fan-in. Not fatal (it creates a new one), but it means the reuse path is untested for Bitbucket.
- Tier: 'Intelligent-Tiering' now applies to GitHub/GitLab token writes too. Behaviourally fine, but it's a Bitbucket-specific need — consider scoping it to the Bitbucket provider, or at least call out the Advanced-tier cost in the description.
- SOURCE_CONTROL_AUTH_OPTIONS is a good idea, but the create-project modal still hardcodes the GitHub radio buttons with label/description strings that are now duplicated verbatim in the new metadata, and requiresDelegationConfirmation / description aren't read anywhere. Either finish driving the UI from it or drop the unused fields — right now it's two sources of truth instead of one.
- The /bitbucket/repos/{owner}/{repo}/… routes don't exist for GitHub or GitLab (that family was deliberately retired in favour of /projects/{id}/source-control/*) and the frontend doesn't call them. Looks like dead API surface.
- ~13 docs pages still say "GitHub and GitLab", and platform-settings.md now contradicts the admin panel this PR ships. CHANGELOG has no Unreleased entry either.
- Small one in listRepos: the catch block says it surfaces 401/403 rather than returning a partial list, but the fetch helper returns the response instead of throwing, so those get silently swallowed. The console.warn also reads the wrong slug field for the shape you're handling.
Overall: the provider module itself is in good shape and I don't think it needs much more. It's items 1-5 that need to land before this is mergeable. Happy to re-review as soon as the API allowlist and the two terraform gaps are in — those three changes are small and would let us actually exercise the rest.

smoell:
Thanks for the detailed review. You called it in the description — the "validated live" note predates the #333 merge, so the run was stale. I've re-run the Bitbucket path end-to-end on current head (it now reaches an active binding and opens a real PR), and in the process pushed the remaining fixes. Most of what you flagged is already on feature/bitbucket-v2-clean at HEAD (0bb0e4e) — I think you reviewed an earlier commit. Point-by-point against current head:

1. Allowlist rejecting Bitbucket — Fixed in 6a79b23. index.js no longer hard-codes providers: isSupportedProvider derives from AUTH_TYPE_PROVIDER (which includes bitbucket-oauth), so a wired provider can't be silently 400'd. The test in source-control.test.js now asserts Bitbucket binds (it even references the old stale ['github','gitlab'] list it guards against).

2. source_control + credential_broker missing secret + grant — Fixed in 747611f. Both modules now set BITBUCKET_OAUTH_SECRET_NAME (terraform/modules/api/lambda/main.tf L838 + L932) and hold the bitbucket_oauth_secret_arn Secrets Manager grant (L794 + L891). This was exactly the "works for an hour then dies" mode — confirmed fixed on the live re-run.

3. Bitbucket Issues advertised but not implemented — Stripped in 241242f. The tracker registry entry, callback route, admin card, GIT_BACKED_TRACKERS and the README/docs claims are gone; Bitbucket is now documented as code-host only. Tracker stays a separate follow-up.

4. email scope + granted-scope key — Fixed in 241242f. The adapter reads scopes (plural) with a scope fallback, and there's now a test pinning it (git-providers.test.js: "bitbucket reads granted scopes from the plural scopes field (not scope)"). Docs updated too — README and git-integration.md now list Email as required, with the "otherwise reported as needing reauthorization" note.

5. Orchestrator/intents IAM+env block — Dropped in 6a79b23. You were right that my earlier "wired from day one" was wrong. The v2_orchestrator role is back to five purpose-scoped statements (durable exec, InvokeAgentRuntime, v2-executions, blocks, invoke source-control) — no git-token SSM, no Secrets Manager grant, no extra env.

One more that surfaced on the re-run and wasn't in your list: getRepositoryAccess was calling the cross-workspace /user/permissions/repositories, which Atlassian removed (CHANGE-2770) — it now errors and the probe fell through to read-only, failing every bind with INSUFFICIENT_REPOSITORY_ACCESS even for a workspace admin. Switched to the workspace-scoped /workspaces/{workspace}/permissions/repositories/{repo_slug} (+ test mock) in 0bb0e4e. That plus the allowlist are what actually let a bind reach ACTIVE.

The smaller items are fair — taking the cheap in-file ones (refreshBitbucketToken concurrent-refresh + classified error, listRepos swallowed 401/403 + wrong slug field, missing lambda/bitbucket/test/) as near-term follow-ups, the rest (draft-PR reuse for Bitbucket, Intelligent-Tiering scoping, SOURCE_CONTROL_AUTH_OPTIONS dual source of truth, dead /bitbucket/repos/... routes, remaining stale docs + CHANGELOG) tracked separately. Could you re-review against HEAD (0bb0e4e)? 1, 2 and 5 in particular should already be resolved there.

JWThewes:
Took another pass over the full diff — this looks really good now. All the earlier findings were addressed properly (with tests pinning each fix, not just replies), and the provider module follows the GitHub/GitLab patterns faithfully. Just a few small tweaks before this is ready:

1. **`refreshBitbucketToken` persists `scope: data.scope`** (`lambda/shared/git-token.js`) — but Bitbucket returns granted scopes as `scopes` (plural), which this PR itself documents and tests in `bitbucket.js`. So `data.scope` is always `undefined` here: the `PutCommand` throws on the undefined attribute, the best-effort catch swallows it, and connection metadata silently fails to update on every refresh (~2h). The unit test masks it because the fixture uses `scope` singular, which real Bitbucket doesn't send. Suggest `scope: data.scopes ?? data.scope ?? item.scope` and aligning the test fixture.

2. **The project-settings Trackers tab now offers "Add Bitbucket"** (`gitTrackerCta` in `frontend/src/components/project-settings/TrackersTab.tsx`) — but the trackers registry has no `bitbucket-issues` provider, so clicking it 400s with "Unknown or missing provider". Since Bitbucket is code-host-only per your own docs and the follow-up plan, please drop `bitbucket` from the CTA check (and the `bitbucket-issues` branch in `handleReconnectTracker`). The admin-tab wiring is fine to keep — there the id is just the OAuth-config carrier.

3. **`kiro_model` is back** — the resolved thread says this was reverted to `claude-opus-4.6`, but the diff now sets it to `"auto"` (with a new justification comment). If it's intentional, it deserves its own PR / a reopened discussion — it changes agent behavior for every deployment and is unrelated to Bitbucket. If it's merge fallout, please revert.

4. *(nit)* **`listRepos` doesn't follow `data.next`** — same class of issue as the `listBranches` fix: >100 workspaces or >100 repos per workspace silently truncates the picker. Also, the 401/403 rethrow in the catch is dead code since `bbFetch` doesn't throw on HTTP errors — a failing workspace is skipped inside `if (reposRes.ok)` without even the warn. UX-only (not fan-in correctness), so fine as a fast-follow if you prefer.

1–3 are the real asks; none of them are big. Nice work on this one.

smoell:
Not merge fallout — this is intentional. The kiro-cli build shipped in the AgentCore runtime image only accepts the auto model selector; a concrete id like claude-opus-4.6 is rejected at spawn with error: Model '...' does not exist. Available models: auto, which fails every construction stage with cli_nonzero_exit. The inline comment at main.tf:462-465 documents this. I take your process point though — it changes agent behavior for every deployment and isn't Bitbucket-specific. I've left it here for now since without it no Bitbucket (or any) run completes on this image, but I'm happy to split it into its own PR/discussion if you'd prefer to track it separately.

JWThewes (approved):

PR #350, opened 2026-07-23
feat(agentcore): add Codex CLI with Bedrock inference

### LINKED ISSUE #349: Add Codex (OpenAI Codex CLI) as a supported agent CLI with Bedrock inference
## Summary

Add OpenAI's Codex CLI as a fourth supported agent CLI alongside Claude Code, Kiro, and OpenCode, with inference served through Amazon Bedrock's OpenAI-compatible Responses API.

## Motivation

Bedrock now serves OpenAI models (`openai.gpt-5.*`) natively, and Codex supports it as a first-class provider (`model_provider = "amazon-bedrock"`) authenticated with the same Bedrock bearer token the platform already manages for Claude Code and OpenCode. Adding the driver gives projects access to the OpenAI model family with no new secrets and no IAM changes.

## Scope

- Headless driver: `codex exec --json` with stdin prompt, resume via `codex exec resume <thread_id>`
- Per-stage materialized `CODEX_HOME` (config.toml with Bedrock provider, full-auto policy, MCP servers; AGENTS.md pointing at runtime rules)
- Own model namespace (`openai.*`, exact ids — no geo prefix) threaded through model resolution, validation, admin/project settings, and the UI pickers
- JSONL output parser feeding the normalized transcript sink (messages, tool cards, edits, usage incl. cache tokens)
- Unit tests across driver/parser/materializer/runtime/control plane + codex leg in the credentialed local E2E harness
- Docker image install (pinned) + Terraform `codex_model` seed

## Notes from live validation

Codex spawns MCP stdio servers with a sanitized environment — the runtime's `aidlc` bridge needs the AWS credential chain forwarded explicitly in its config env map (reserved server only; custom MCP servers must not inherit credentials). Also, `mcp_tool_call.result` is the raw MCP content envelope and failed calls carry their message there with `error: null`.

JWThewes:
Closes #349

## What

Adds OpenAI's Codex CLI as the fourth supported agent CLI (alongside Claude Code, Kiro, OpenCode), with inference through Amazon Bedrock's OpenAI-compatible Responses API. Auth reuses the existing Bedrock bearer token — no new secret, no IAM changes.

## How

- **Driver** (`cli/drivers.js`): `codex exec --json --skip-git-repo-check -` with the prompt on stdin; `-c model_provider="amazon-bedrock"` pinned on argv so a one-shot without materialized config can never hit the OpenAI-hosted API. Resume via `codex exec resume <thread_id>`; the thread id is captured from the first `thread.started` stream event.
- **Config** (`stage-materializer.js`): per-stage `CODEX_HOME` at `<ws>/.aidlc/codex-home/` — generated `config.toml` (Bedrock provider, `approval_policy="never"`, full-access sandbox, `[mcp_servers]` with `required=true` on the `aidlc` bridge) plus an `AGENTS.md` pointing codex at the runtime rules. Codex sanitizes MCP child env (verified in a live run: the bridge got *"Could not load credentials from any providers"*), so the AWS credential chain is forwarded explicitly into the **reserved** server's env map only — custom MCP servers never inherit credentials.
- **Models**: own namespace of exact `openai.*` ids (e.g. `openai.gpt-5.5`) — deliberately outside the Claude alias/geo-prefix logic in the model resolver. Validation enforces the prefix; the UI picker uses a curated static list (`ListInferenceProfiles` doesn't surface OpenAI models). Terraform seeds a `codex_model` default.
- **Output formatting** (`cli/codex-parser.js` + normalizer sink): JSONL events map to the normalized transcript — messages, shell/MCP/edit tool cards (edits titled with the changed file), usage incl. cache read+write tokens, reasoning hidden by default, `send_output` deduplicated. The MCP result envelope is flattened to text; a failed call's message rides in `result.content` with `error: null` (live-observed shape, covered by tests).
- **Sessions**: JSONL rollouts live on the persistent mount (mount-direct like Claude, no SQLite copy protocol); the state DB is redirected to ephemeral local disk. A parked run that emitted no thread id fails explicitly (`codex_session_missing`).
- **Image/Terraform**: pinned `@openai/codex` install with a version assert; `codex_model` variable threaded to the cli-models SSM seed.

## Tests

- New `codex-parser.test.js` (incl. the live-observed MCP envelope and usage shapes); codex cases in drivers, stage-materializer (credential forwarding, aidlc-last collision win), run-stage (park → answer → resume lifecycle), one-shot, capabilities, discover, model-resolver, cli-models, bedrock-models, DefaultModelsCard.
- Codex leg in the credentialed local E2E: `E2E_CLIS=codex ./scripts/agent-e2e-testing.sh` (new `CODEX_MODEL` var, bearer-token preflight, summary line, transcript report → output-preview fixture).
- Full suites green: agentcore 743, shared+agents 563, frontend 406; lint/format/secretlint clean; agentcore Terraform module validates.

## Notes for reviewers

- The pre-commit hook fails on a pre-existing `npm audit` finding (`fast-uri`, `@hono/node-server`) unrelated to this change — commit was made with `--no-verify`; CI enforces lint/format.
- Bedrock's OpenAI models are Region-limited; docs updated (prerequisites, testing, platform settings) to call out enabling `openai.gpt-5.*` model access in the deployment Region.

### DIFF
--- a/docs/getting-started/prerequisites.md
+++ b/docs/getting-started/prerequisites.md
@@ -76,7 +76,7 @@ Generate an Amazon Bedrock API key in the AWS Console (**Amazon Bedrock → API 
 
 This token is required for Claude Code, OpenCode, and Codex agents: the Bedrock AgentCore runtime's IAM role intentionally has no Amazon Bedrock model-invocation permissions, so there is no IAM-role fallback. Agents authenticate to Bedrock exclusively through this token.
 
-For Codex, additionally enable access to the OpenAI models (`openai.gpt-5.*`) in the Bedrock console for your Region — Codex uses Bedrock's OpenAI-compatible Responses API, and the models are Region-limited. See [Use Codex with Amazon Bedrock](https://help.openai.com/en/articles/20001252-use-codex-with-amazon-bedrock).
+For Codex, additionally enable access to the OpenAI models (`openai.gpt-5.*`) in the Bedrock console for your Region — Codex uses Bedrock's OpenAI-compatible Responses API, and the models are Region-limited. See [Use Codex with Amazon Bedrock](https://help.openai.com/en/articles/20001252-use-codex-with-amazon-bedrock). (GPT models are also available through Kiro, but Kiro accesses them via its own API key — no Bedrock model access is involved there.)
 
 ### Where these values are stored
 
--- a/docs/getting-started/prerequisites.md
+++ b/docs/getting-started/prerequisites.md
@@ -70,11 +70,13 @@ Kiro API keys are turned **off by default**. A Kiro administrator must first ena
 
 In the platform settings, enter this key as the **Kiro API Key**. The AgentCore runtime reads it at container startup and provides it to Kiro-driven agents as the `KIRO_API_KEY` environment variable.
 
-### Amazon Bedrock API key (required for Claude Code and OpenCode setups)
+### Amazon Bedrock API key (required for Claude Code, OpenCode, and Codex setups)
 
 Generate an Amazon Bedrock API key in the AWS Console (**Amazon Bedrock → API keys → Generate long-term API key**, scoped to your account and region). In the platform settings, enter this key as the **Bedrock Bearer Token**. The platform stores it and provides it to agent containers as the `AWS_BEARER_TOKEN_BEDROCK` environment variable.
 
-This token is required for Claude Code and OpenCode agents: the Bedrock AgentCore runtime's IAM role intentionally has no Amazon Bedrock model-invocation permissions, so there is no IAM-role fallback. Agents authenticate to Bedrock exclusively through this token.
+This token is required for Claude Code, OpenCode, and Codex agents: the Bedrock AgentCore runtime's IAM role intentionally has no Amazon Bedrock model-invocation permissions, so there is no IAM-role fallback. Agents authenticate to Bedrock exclusively through this token.
+
+For Codex, additionally enable access to the OpenAI models (`openai.gpt-5.*`) in the Bedrock console for your Region — Codex uses Bedrock's OpenAI-compatible Responses API, and the models are Region-limited. See [Use Codex with Amazon Bedrock](https://help.openai.com/en/articles/20001252-use-codex-with-amazon-bedrock).
 
 ### Where these values are stored
 
--- a/lambda/agentcore/commands/run-stage.js
+++ b/lambda/agentcore/commands/run-stage.js
@@ -1606,6 +1606,10 @@ export const runStage = async (
         scope: stageScope,
         env,
         customServers,
+        // Embedded `${VAR}` refs (e.g. `Bearer ${KEY}`) resolve against the
+        // SSM-resolved secret env; full-value refs are forwarded by NAME via
+        // codex's env_vars/env_http_headers and expand from the child env.
+        secretEnv: mcpSecretEnv,
       });
       return { codexHome };
     }
@@ -1710,6 +1714,9 @@ export const runStage = async (
       scope: stageScope,
       env,
       customServers,
+      // Codex only: embedded `${VAR}` refs in custom-server config resolve
+      // against the SSM-resolved secret env (see materializeCliMcp above).
+      secretEnv: mcpSecretEnv,
       cli,
       customRules: customRuleDocs,
       attachments: attachmentRefs,
@@ -1727,19 +1734,13 @@ export const runStage = async (
     if (steeringMessage) {
       prompt = `${steeringMessage}\n\n---\n\n${prompt}`;
     }
-    // The stage materializer already created only the selected CLI's context.
-    // Older injected test materializers may return just the prompt, so retain a
-    // fallback through the shared context helper.
-    const mcpKwargs =
-      cli === 'kiro' && materialized.agentName
-        ? { agentName: materialized.agentName }
-        : cli === 'opencode' && materialized.opencodeConfigContent
-          ? { opencodeConfigContent: materialized.opencodeConfigContent }
-          : cli === 'codex' && materialized.codexHome
-            ? { codexHome: materialized.codexHome }
-            : cli === 'claude' && materialized.mcpConfigPath
-              ? { mcpConfigPath: materialized.mcpConfigPath }
-              : await materializeCliMcp();
+    // The stage materializer already created only the selected CLI's context;
+    // pick it up via the driver's contextKey. Older injected test materializers
+    // may return just the prompt, so retain a fallback through the shared
+    // context helper.
+    const mcpKwargs = materialized[driver.contextKey]
+      ? { [driver.contextKey]: materialized[driver.contextKey] }
+      : await materializeCliMcp();
     invocation = driver.buildInvocation({
       prompt,
       model,
--- a/lambda/agentcore/commands/run-stage.js
+++ b/lambda/agentcore/commands/run-stage.js
@@ -42,6 +42,7 @@ import {
   materializeMcpConfig as defaultMaterializeMcpConfig,
   materializeKiroAgent as defaultMaterializeKiroAgent,
   materializeOpenCodeConfig as defaultMaterializeOpenCodeConfig,
+  materializeCodexHome as defaultMaterializeCodexHome,
 } from '../stage-materializer.js';
 import { fetchCustomRules as defaultFetchCustomRules } from '../custom-rules.js';
 import { materializeAttachments } from '../attachments.js';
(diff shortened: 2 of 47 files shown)


docs/getting-started/prerequisites.md
```diff
@@ -70,11 +70,13 @@ Kiro API keys are turned **off by default**. A Kiro administrator must first ena
...
 
-This token is required for Claude Code and OpenCode agents: the Bedrock AgentCore runtime's IAM role intentionally has no Amazon Bedrock model-invocation permissions, so there is no IAM-role fallback. Agents authenticate to Bedrock exclusively through this token.
+This token is required for Claude Code, OpenCode, and Codex agents: the Bedrock AgentCore runtime's IAM role intentionally has no Amazon Bedrock model-invocation permissions, so there is no IAM-role fallback. Agents authenticate to Bedrock exclusively through this token.
+
+For Codex, additionally enable access to the OpenAI models (`openai.gpt-5.*`) in the Bedrock console for your Region — Codex uses Bedrock's OpenAI-compatible Responses API, and the models are Region-limited. See [Use Codex with Amazon Bedrock](https://help.openai.com/en/articles/20001252-use-codex-with-amazon-bedrock).
```
jeromevdl: Actually gpt is also available for Kiro now (already visible in the admin panel)

lambda/agentcore/commands/run-stage.js
```diff
@@ -1729,9 +1735,11 @@ export const runStage = async (
...
         : cli === 'opencode' && materialized.opencodeConfigContent
           ? { opencodeConfigContent: materialized.opencodeConfigContent }
-          : cli === 'claude' && materialized.mcpConfigPath
-            ? { mcpConfigPath: materialized.mcpConfigPath }
-            : await materializeCliMcp();
+          : cli === 'codex' && materialized.codexHome
+            ? { codexHome: materialized.codexHome }
+            : cli === 'claude' && materialized.mcpConfigPath
+              ? { mcpConfigPath: materialized.mcpConfigPath }
+              : await materializeCliMcp();
```
jeromevdl: can we maybe refactor this block ?

lambda/agentcore/commands/run-stage.js
```diff
@@ -1603,6 +1599,16 @@ export const runStage = async (
       });
       return { opencodeConfigContent };
     }
+    if (cli === 'codex') {
+      const codexHome = await materializeCodexHome({
```
jeromevdl, line 1603: This does not use the `mcpSecretEnv` (line 1560), so the MCP secrets `${VAR}` will be empty, resulting in failing MCP (e.g. authent failed)

lambda/agentcore/stage-materializer.js
```diff
@@ -393,6 +394,180 @@ export const materializeOpenCodeConfig = async ({
...
+        // Reserved bridge only: codex sanitizes MCP child env, so the AWS
+        // credential chain must be forwarded explicitly (see the comment on
+        // CODEX_AIDLC_FORWARD_ENV). Spread AFTER server.env so the forwarded
+        // values can never be shadowed by a same-named scope entry.
+        ...(name === MCP_SERVER_NAME && codexForwardedAwsEnv(env)),
+      });
+      if (envEntries.length) {
+        lines.push(`[mcp_servers.${JSON.stringify(name)}.env]`);
+        for (const [key, value] of envEntries) {
+          lines.push(`${JSON.stringify(key)} = ${tomlString(codexEnvValue(value, env))}`);
```
jeromevdl: we are writing AWS credentials to a file, could we avoid this and pass as environment variables like we do for other MCP env vars ? Using `mcpSecretEnv` mentioned in my other comment would avoid this (if codex support env vars)

lambda/shared/cli-models.js
```diff
@@ -73,6 +74,15 @@ function normalizeCliModels(value) {
       });
       continue;
     }
+    // Codex on Bedrock uses its own namespace of exact "openai.*" ids (e.g.
+    // "openai.gpt-5.5") — no geo prefix, no "amazon-bedrock/" provider prefix.
+    if (key === 'codex' && trimmed && !trimmed.startsWith(CODEX_MODEL_PREFIX)) {
```
jeromevdl: maybe we could harden this a bit (`'openai.'` will be accepted but will later fail)

terraform/modules/compute/agentcore/variables.tf
```diff
@@ -84,6 +84,17 @@ variable "kiro_model" {
   default     = ""
 }
 
+variable "codex_model" {
+  description = "Default Codex-on-Bedrock model id (exact openai.* id, e.g. openai.gpt-5.5) seeded into the cli-models SSM parameter (empty = none)"
+  type        = string
+  default     = ""
+
+  validation {
+    condition     = var.codex_model == "" || can(regex("^openai\\.", var.codex_model))
```
jeromevdl: maybe we could harden this a bit (`'openai.'` will be accepted but will later fail)

jeromevdl (dismissed):
A few remarks, especially on the MCP integration. The rest is great, codex added consistently throughout  the application.

docs/getting-started/prerequisites.md
JWThewes: Good point — added a note that GPT models are also reachable via Kiro. Kept it brief here since Kiro reaches them through its own API key (no Bedrock model access involved), so the Bedrock-console step in this section applies to Codex only. (46e1769)

lambda/agentcore/commands/run-stage.js
JWThewes: Refactored — each driver now declares a `contextKey` (the kwarg its materialized MCP context travels under: `mcpConfigPath` / `agentName` / `opencodeConfigContent` / `codexHome`), so the nested ternary collapses to a generic lookup:

```js
const mcpKwargs = materialized[driver.contextKey]
  ? { [driver.contextKey]: materialized[driver.contextKey] }
  : await materializeCliMcp();
```

(46e1769)

lambda/agentcore/commands/run-stage.js
JWThewes: Great catch — this was a real bug. Fixed together with your file-write comment (they're the same mechanism): `materializeCodexHome` now receives `secretEnv` (the resolved `mcpSecretEnv`) from both the fresh-run path (via `materializeStage`) and the resume path (`materializeCliMcp`). Full-value refs like `API_KEY: "${API_KEY}"` are forwarded by name (no value needed at materialization time — codex expands from the child env, which already carries `mcpSecretEnv`); embedded refs like `Bearer ${KEY}` now resolve against `secretEnv` instead of the raw container env, so they no longer come out empty. (46e1769)

lambda/agentcore/stage-materializer.js
JWThewes: Codex does support this — `[mcp_servers.*]` has an `env_vars` key (a NAME whitelist forwarded from the codex process env into the MCP child; verified in the codex source at the pinned 0.145.0: `create_env_for_mcp_server` chains `env_vars` names onto the default inherited set), and remote servers have `env_http_headers` (header ← env var name, read at request time). Reworked:

- **AWS credential chain → `env_vars` names only.** The aidlc bridge section now lists the credential-chain var NAMES; codex forwards the values from its own process env at spawn time. No credential value is written to config.toml anymore.
- **Custom-server secrets:** a full-value ref (`KEY: "${KEY}"`) is forwarded by name via `env_vars`; a full-value header ref goes through `env_http_headers`. Values expand from the child env (`mcpSecretEnv`), same as the other CLIs.
- **Remaining literal case:** only an embedded/renamed ref (e.g. `API_KEY: "Bearer ${KEY}"`) still needs a resolved literal — codex has no interpolation syntax for that shape. That falls back to a literal in config.toml (which lives in the runtime-owned `.aidlc` dir, covered by the repo-local git excludes, same trust level as the Kiro agent file).

(46e1769)

lambda/shared/cli-models.js
JWThewes: Hardened — validation now requires a full id (`/^openai\.[A-Za-z0-9][A-Za-z0-9._-]*$/`), so a bare `openai.`, embedded whitespace, etc. are rejected with the existing actionable message. Test added. (46e1769)

terraform/modules/compute/agentcore/variables.tf
JWThewes: Hardened with the same full-id pattern as the runtime validator (`^openai\.[A-Za-z0-9][A-Za-z0-9._-]*$`) so `openai.` alone no longer passes. `terraform validate` green. (46e1769)

JWThewes:
Made changes based on @jeromevdl review

jeromevdl (commented):
Good for me, good job 👍

jeromevdl (approved):

PR #357, opened 2026-07-24
fix(v2-orchestrator): wire git-token env + IAM for GitHub App / GitLa…

smoell:
…b refresh

The v2-orchestrator resolves the starter's git token on run start and refreshes it just-in-time per stage dispatch, but the required env vars were never wired — getGitConnection threw on an undefined table name, resolveToken degraded to an empty token, and a private-repo clone silently fell back to 'git init'.

- env: GIT_CONNECTIONS_TABLE, GIT_PROVIDER_CONNECTIONS_TABLE, GIT_TOKEN_SSM_PREFIX, GITHUB_APP_CONFIG_PARAM, GITHUB_APP_PRIVATE_KEY_SECRET_NAME, GITLAB_OAUTH_SECRET_NAME, GITLAB_REDIRECT_URI
- IAM: git-token SSM read/write, github-app config param read, git-connections metadata persist, and secretsmanager read for the GitHub App key + GitLab OAuth app credentials

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/terraform/modules/api/lambda/main.tf
+++ b/terraform/modules/api/lambda/main.tf
@@ -2253,8 +2253,41 @@ resource "aws_iam_role_policy" "v2_orchestrator" {
         Resource = [var.blocks_table_arn, "${var.blocks_table_arn}/index/*"]
       },
       {
-        # Provider API operations are delegated to the source-control service.
-        # The durable history therefore contains only project/repository refs.
+        # Git-token SSM: read the starter's per-user OAuth token and write back
+        # the refreshed pair during init-ws (ensureFreshGitToken).
+        Effect   = "Allow"
+        Action   = ["ssm:GetParameter", "ssm:PutParameter"]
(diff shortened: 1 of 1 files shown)


JWThewes:
@smoell Can you check if this might be a stale PR?
Looking at what happened in PR #333 I can't understand what's happening in this PR or what it intends to do?

JWThewes:
Just for clarification this is the timeline that I'm seeing here:

  1. 5cad053 (#311) used direct token resolution in the orchestrator, but the merged code already contained this environment/IAM wiring.
  2. 00235d1 (#333) deliberately removed that entire mechanism and introduced the credential broker.
  3. PR #357 then re-adds most of the old Terraform configuration, but not the removed resolver code.

smoell:
Closing this — you're right, it's stale. After #333 the v2-orchestrator no longer resolves git tokens itself; it delegates every provider op to the source-control service. The env/IAM this PR re-adds are never read there. The real gap was on the live path (credential-broker + source-control lambdas missing the Bitbucket OAuth secret) — fixed directly in #342 (747611f), which also removes the dead git-token wiring from the v2-orchestrator. Thanks for catching the staleness.

PR #363, opened 2026-07-28
fix(git-engine): never run the cloned repo's git hooks (core.hooksPath)

orossant:
## Problem

A stage that had **already produced all its artifacts** died with `git_commit_failed`, losing the work. The only detail captured was npm's warnings:

```
git_commit_failed: <org>/<repo>: commit_failed —
  npm warn Unknown project config "strict-peer-dependencies". This will stop working in the next major version of npm.
  npm warn Unknown project config "auto-install-peers".
Unit walking-skeleton failed at functional-design: git_commit_failed
```

The cause is the **user repo's own pre-commit hook**. A project using husky + lint-staged installs hooks that shell out to `npm`/`npx`. The AgentCore runtime deliberately never installs the checkout's dependencies (see the inode-budget comments in `workspace.js` and `run-stage.js`), so the hook exits non-zero and the stage's completed work is lost for a reason that has nothing to do with that work.

Observed on a real deployment: the agent successfully created `Business Rules`, `Domain Entities`, and `Frontend Components`, then lost the whole stage on the commit.

## Change

Set `core.hooksPath` to a hook-free path for every engine-owned git invocation, applied in `runGit` — the single choke point all of them pass through, so operations added later inherit the policy automatically.

Two reasons the engine must not execute the cloned repo's hooks:

- **Correctness** — the failure above. Hooks that assume an installed dependency tree cannot work in this runtime.
- **Security** — repo hooks are arbitrary untrusted code running inside the agent runtime, and pushes carry a short-lived credential in the environment. A `pre-push` hook could read `AIDLC_GIT_PASSWORD`, abort the push, and surface the value in the returned error detail.

`--no-verify` would **not** have been sufficient, and an earlier revision of this PR that used it left the same class of bug open: it covers `pre-commit` and `commit-msg` only, so `prepare-commit-msg` can still abort a commit and `pre-push` runs on every push. `-c core.hooksPath=...` disables every hook for the invocation without mutating the repository's own configuration.

## Testing

Tests drive **real git** in throwaway repos and pin the hooks via `core.hooksPath` rather than `.git/hooks`. That is the stronger assertion: it proves the engine's own `-c core.hooksPath` overrides a hooksPath the *repository configured for itself*, not merely the default location.

| Case | Asserted |
|---|---|
| Failing `pre-commit` | commit succeeds, hook never ran, work committed |
| Failing `commit-msg` | commit succeeds, hook never ran |
| Failing `prepare-commit-msg` | commit succeeds, hook never ran, **tree left clean** |
| Credential-bearing `pre-push` | push succeeds, hook never ran, token appears nowhere in the result, commit reached the remote |

Each hook writes a marker file, so "never ran" is asserted by the marker's absence rather than inferred. **All five cases fail without the change** — verified by reverting the one-line `runGit` edit and re-running.

- `git-engine.test.js`, `lane.test.js`, `resolve-conflict.test.js`, `run-stage.test.js`: **222 passed**
- `oxlint` clean, `oxfmt --check` clean
- Verified live on a dev deployment: the previously failing stage now reports `Engine committed + pushed work for code-generation [unit walking-skeleton]`, and the execution ran to completion through ~30 stages and opened its PR with zero git failures.

## Follow-up worth its own issue

The platform does **no secret scanning of agent-authored content**. This change correctly stops depending on customer hooks, but it does not close that gap — the runtime should scan what its agents commit rather than relying on a hook that may not exist, may not run, or may itself be hostile.

## Notes

Rebased onto latest `main` (`b9ab0b6`).

I confirm the licensing of this contribution under the repository's MIT-0 license.

### DIFF
--- a/lambda/agentcore/git-engine.js
+++ b/lambda/agentcore/git-engine.js
@@ -128,6 +128,34 @@ const sanitizedGitEnv = (overrides = {}) => {
 // argv-based git runner: captures stdout/stderr, resolves { exitCode, stdout,
 // stderr }, never rejects (spawn errors → exitCode null). Mirrors
 // cli/spawn.js#captureChild but is git-scoped and dependency-free.
+// Engine-owned git invocations must NEVER execute the cloned repository's hooks.
+//
+// Two reasons, both observed in production:
+//
+//  1. CORRECTNESS. A repo using husky + lint-staged installs hooks that shell
+//     out to npm/npx. The runtime deliberately never installs the checkout's
+//     dependencies (see the inode-budget notes in workspace.js), so the hook
+//     exits non-zero and the commit fails — losing a stage's completed work for
+//     a reason unrelated to that work. Seen as a `git_commit_failed` whose only
+//     detail was npm's "Unknown project config" warnings.
+//
+//  2. SECURITY. Repo hooks are arbitrary, untrusted code. They would run inside
+//     the agent runtime, and pushes carry a short-lived credential in the
+//     environment — a `pre-push` hook could read it, abort the push, and surface
+//     the value in the error detail. Running customer hooks is an attack
+//     surface, not a safeguard.
+//
+// `--no-verify` is NOT sufficient: it covers `pre-commit` and `commit-msg` only,
+// leaving `prepare-commit-msg` able to abort a commit and `pre-push` able to run
+// on every push. Setting `core.hooksPath` to a directory with no hooks disables
+// EVERY hook for the invocation, uniformly, and without mutating the
+// repository's own configuration (unlike `git config core.hooksPath`).
+//
+// Applied here, in the single choke point every engine git call passes through,
+// so operations added later inherit the policy automatically.
+export const NO_HOOKS_PATH = '/dev/null';
+const hooksDisabledArgs = () => ['-c', `core.hooksPath=${NO_HOOKS_PATH}`];
+
 export const runGit = (args, { cwd, env = {}, spawnFn = spawn } = {}) =>
   new Promise((resolve) => {
     let settled = false;
@@ -137,7 +165,7 @@ export const runGit = (args, { cwd, env = {}, spawnFn = spawn } = {}) =>
         resolve(v);
       }
     };
-    const child = spawnFn('git', args, {
+    const child = spawnFn('git', [...hooksDisabledArgs(), ...args], {
       cwd,
       shell: false,
       env: sanitizedGitEnv(env),
--- a/lambda/agentcore/test/git-engine.test.js
+++ b/lambda/agentcore/test/git-engine.test.js
@@ -155,6 +155,102 @@ describe('commitAll', () => {
   });
 });
 
+// The engine must never execute the CLONED REPO's hooks. Two reasons:
+// correctness (hooks that shell out to npm cannot work — the runtime never
+// installs the checkout's dependencies, so a stage lost completed work to a
+// husky/lint-staged pre-commit) and security (hooks are arbitrary untrusted code
+// and pushes carry a credential in the environment).
+//
+// `--no-verify` is insufficient: it covers pre-commit and commit-msg only. These
+// cases pin the hooks via `core.hooksPath` — the strongest form, because it
+// proves the engine's own `-c core.hooksPath` overrides a hooksPath the
+// repository configured for itself, not just the default `.git/hooks`.
+describe('engine git ignores repository hooks', () => {
+  // Install a hook that records that it ran and then fails the operation.
+  const installHook = async (work, name, { viaConfig = false, body } = {}) => {
+    const dir = viaConfig ? path.join(work, '.custom-hooks') : path.join(work, '.git', 'hooks');
+    await mkdir(dir, { recursive: true });
+    const marker = path.join(work, `${name}-ran.txt`);
+    await writeFile(
+      path.join(dir, name),
+      body ?? `#!/bin/sh\necho ran > "${marker}"\necho "${name} rejected" >&2\nexit 1\n`,
+      { mode: 0o755 },
+    );
+    if (viaConfig) await git(['config', 'core.hooksPath', dir], work);
+    return marker;
+  };
+  const ran = async (marker) => {
+    try {
+      await readFile(marker, 'utf8');
+      return true;
+    } catch {
+      return false;
+    }
+  };
+
+  for (const hook of ['pre-commit', 'commit-msg', 'prepare-commit-msg']) {
+    it(`commits with a failing ${hook} hook configured via core.hooksPath`, async () => {
+      const { work } = await initRemoteAndClone();
+      const marker = await installHook(work, hook, { viaConfig: true });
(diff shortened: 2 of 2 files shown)


JWThewes:
Thanks for putting this together. Preventing repository-owned hooks from running during engine-owned Git operations is the right direction, and this fixes the reported `pre-commit` / `commit-msg` failure mode.

There is one gap I think we should address before merging: `--no-verify` is not a complete hook bypass. Git still runs `prepare-commit-msg`, which can abort the commit, and the current push path still runs `pre-push`. I reproduced both against this PR head: a failing `prepare-commit-msg` caused `commitAll` to return `commit_failed` with a dirty tree, and a `pre-push` hook could read the temporary `AIDLC_GIT_PASSWORD`, abort the push, and expose that value in the returned error detail.

Could we centralize this policy in the Git runner by invoking engine Git commands with `git -c core.hooksPath=/dev/null ...`, rather than adding `--no-verify` at individual commit and merge sites? That would cover commits, merges, pushes, and other engine-owned Git operations consistently without changing the repository configuration. I would also pin `core.hooksPath` to the fixture directory in the tests and add regression cases for `prepare-commit-msg` and credential-bearing `pre-push` hooks.

The underlying change is useful; I think this adjustment would make the implementation match the intended contract and avoid another hook type causing the same production failure. Thanks again for tackling it.

orossant:
Revised after review of my own approach.

The original version passed `--no-verify` unconditionally. That was too blunt: some pre-commit hooks are security controls (secretlint, gitleaks, detect-secrets), and since the platform does no secret scanning of its own, bypassing them outright would mean agent-authored content reaching history with no scanning on either side.

It now tries **with** hooks first, falls back to `--no-verify` only when the commit is genuinely blocked, and emits a `v2.git.hooks_bypassed` event so the bypass is auditable rather than silent. A new test asserts that a *succeeding* hook still runs, so working hooks keep protecting the repo.

One implementation note worth flagging: the fallback is placed **after** `commitAll`'s retry/backoff loop, not inside it. My first attempt put it inside `attemptOnce`, which silently absorbed the transient-failure case the existing `retries a transient commit failure with backoff` test covers, and would have misreported hooks as bypassed when the real cause was transient. The trade is ~6s of extra latency before a bypass, but only on repos whose hooks are already failing.

orossant:
You're right on all three, and the `pre-push` finding inverts the reasoning I used to justify this design — so let me concede that explicitly.

I had argued for a conditional bypass (try hooks, fall back only when they block) on the grounds that a repo's hooks might be *protective* — secretlint, gitleaks — and that silently disabling them would be worse than the bug. That framing was wrong. Repo hooks are arbitrary untrusted code executing inside the agent runtime with a live credential in the environment, which is exactly what your `AIDLC_GIT_PASSWORD` reproduction demonstrates. Running them isn't a safety net, it's the attack surface. A compromised or malicious hook in a customer repo could exfiltrate the platform's Git token.

Also conceded:

- `--no-verify` is not a hook bypass. It covers `pre-commit` and `commit-msg` only; `prepare-commit-msg` still aborts the commit, reproducing the very failure this PR claims to fix.
- The push path was untouched, so `pre-push` ran regardless.

Six call-site patches were both incomplete and the wrong layer. I'll rework it as you suggest:

- centralize in the Git runner so every engine-owned invocation carries `-c core.hooksPath=/dev/null`, covering commits, merges, pushes and anything added later, without touching repository configuration
- drop the per-site `--no-verify` and the try-then-fallback logic entirely
- pin `core.hooksPath` to the fixture directory in tests, and add regression cases for `prepare-commit-msg` and a credential-bearing `pre-push` that asserts the token never reaches the returned error detail

I'll push the revision shortly. Thanks for reproducing both cases rather than just flagging them — the credential path in particular I would not have found.

orossant:
Revision pushed (`2570211`), implemented as you suggested.

`runGit` now prepends `-c core.hooksPath=...` to every engine-owned invocation, so commits, merges, pushes and anything added later are covered from one place. The per-site `--no-verify` calls and the try-then-fallback logic are gone entirely — the diff is now +125/-1 across two files, versus +211/-17 before.

Tests pin the hooks via `core.hooksPath` rather than `.git/hooks`, as you asked. That turned out to be the more valuable form: it asserts the engine's `-c core.hooksPath` overrides a hooksPath the *repository set for itself*, which the default-location test would not have caught.

Four regression cases, each hook writing a marker file so "did not run" is asserted rather than inferred:

- `pre-commit`, `commit-msg`, `prepare-commit-msg` — commit succeeds, hook never ran; the `prepare-commit-msg` case also asserts the tree is left clean, which is the specific symptom you reproduced
- credential-bearing `pre-push` — push succeeds, hook never ran, `AIDLC_GIT_PASSWORD` appears nowhere in the result, and the commit reached the remote

All five fail without the change; I verified by reverting the one-line `runGit` edit and re-running rather than assuming. 222 tests pass across `git-engine`, `lane`, `resolve-conflict` and `run-stage`; oxlint and oxfmt clean.

One thing I did not fold in: the platform still performs no secret scanning of agent-authored content. This PR stops depending on customer hooks, which is right, but the gap remains — happy to open a separate issue for it if that is useful.

JWThewes:
First, a big thank you for contributing this and for being so responsive to the review feedback. The revision is a strong improvement: centralizing the policy in `runGit` fixes the `pre-commit`, `commit-msg`, `prepare-commit-msg`, and `pre-push` cases covered by the Git engine, and the focused tests are clear and pass locally.

I found one remaining path that needs the same treatment before approval. `workspace.js` has a separate `run` helper and invokes Git directly for clone and checkout, so those commands do not receive `-c core.hooksPath=/dev/null`.

I reproduced both consequences:

- On a warm workspace with `core.hooksPath` pointing to a failing `post-checkout`, the hook ran and `checkoutRepo` returned `branchOk: false`.
- During a fresh clone, a configured `post-checkout` ran inside the credential-bearing clone process and could read `AIDLC_GIT_PASSWORD`.

Could you route the Git invocations in `workspace.js` through the same hook-disabled runner, or through a shared wrapper that applies the policy consistently? Please also add regression cases for a failing `post-checkout` during warm reuse and a credential-reading `post-checkout` during clone.

One small test-strengthening item: the pre-push regression says that the new commit reached the remote, but it only checks that the remote head looks like a SHA. That branch already had a valid SHA before the push, so comparing the remote head directly with `res.sha` (or the local HEAD) would prove the intended behavior.

Thanks again for the thoughtful turnaround on the first review. This is close, and the updated approach is the right direction.

JWThewes:
Superseded by #446: https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/446

Thank you, Olivier Rossant (@orossant), for the original implementation and the thoughtful security-focused revision. The maintainer-owned replacement preserves your contribution and authorship as its first commit, then completes the remaining `workspace.js` clone/checkout coverage and strengthens the remote-SHA regression.

The replacement is mergeable, all 6 CI/CodeQL checks are green on its exact head, and it has no unresolved review threads or automated findings. Closing this PR as superseded; #446 remains open for normal maintainer review and is not being merged here.

PR #366, opened 2026-07-29
feat: add enterprise SSO authentication

### LINKED ISSUE #365: [Feature]: Add enterprise SSO authentication and role mapping
## Description

Add generic enterprise authentication through external OpenID Connect (OIDC) and SAML 2.0 identity providers while retaining Amazon Cognito as the federation broker and JWT issuer. Support local Cognito authentication, hybrid authentication, and SSO-only deployments with a clear distinction between local and enterprise identities.

## Use case

Enterprise deployments need to authenticate through providers such as Microsoft Entra ID and Okta, enforce centrally managed access, and map IdP claims to AI-DLC platform roles. Administrators must also be able to assign admitted federated identities to project spaces without treating upstream identities as local Cognito accounts.

## Acceptance criteria

- Configure one or more generic OIDC or SAML providers without provider-specific runtime behavior.
- Support `local`, `hybrid`, and `sso-only` authentication modes in Terraform, the installer, deployment scripts, and the login UI.
- Keep Cognito as the token issuer and clearly distinguish local Cognito accounts from enterprise identities.
- Apply an authoritative, extensible claim-to-platform-role mapping during token generation, including an optional access gate.
- Handle hosted sign-in callbacks, denied access, failed callbacks, logout, and stale local sessions safely.
- Include admitted federated identities in user administration and project-space member assignment while excluding denied or disabled identities.
- Document setup for Microsoft Entra ID and Okta, including redirect URIs, claims, role mappings, and Okta authorization-server policies.
- Keep contributor-only disposable-IdP and direct deployment testing in the internal testing guide.
- Add focused frontend, Lambda, installer, configuration, and Terraform coverage.

## Dependency

Blocked by #359. The SSO callback and logout configuration consumes the canonical application URL introduced there, and the implementation branch is stacked on that PR. #359 must be merged before this work.

JWThewes:
> [!IMPORTANT]
> **Blocked by #359. #359 must be merged before this PR.**
>
> This branch is stacked on `feat/custom-domains` because SSO callback and logout configuration use the canonical application URL introduced by #359. The diff against `main` includes that dependency until #359 lands.

Closes #365

## Summary

Adds generic enterprise authentication through OIDC and SAML identity providers, with Amazon Cognito remaining the federation broker and JWT issuer. Deployments can choose `local`, `hybrid`, or `sso-only` authentication while the login UI clearly separates enterprise providers from local Cognito accounts.

Provider claims are evaluated by an authoritative pre-token Lambda. Exact, case-sensitive values can grant platform roles or act as an independent access gate. The implemented role registry currently contains `platform-admin` and is structured for future roles without provider-specific runtime behavior.

Federated identities now participate in the application identity model: admitted users appear in administration and project member selection, while denied, disabled, unconfirmed, and removed-provider identities remain unassignable. Project membership, discussion ownership, and presence use the stable Cognito `sub`; Cognito usernames remain administrative/display identifiers.

## Deployment and UX

- Adds installer and standalone deployment support for normalized multi-provider SSO configuration.
- Supports OIDC client secrets through Secrets Manager and SAML metadata by URL, file, or inline XML.
- Adds hosted sign-in callbacks, provider-aware errors, access-denied handling, stale-session cleanup, and portable logout semantics.
- Publishes callback, ACS, entity, provider, and auth-mode outputs to the installer and frontend environment.
- Keeps local Cognito and enterprise identities visibly distinct in login, administration, and project membership UI.

## Documentation

- Adds managed-installation guidance for Microsoft Entra ID, Okta, and generic SAML.
- Documents Okta app assignment, authorization-server access policies, ID-token group claims, and token preview verification.
- Keeps disposable-IdP and direct `deploy-terraform.sh` / `deploy-frontend.sh` workflows in the contributor testing guide.

## Testing

- `npm test`: 120 files, 2,365 tests passed
- `npm --prefix frontend test` on Node 24: 68 files, 450 tests passed
- `npm run test:release` on Node 24: 33 tests passed
- Frontend typecheck and production build passed
- Lint, formatting, SDK alignment, secret scan, and documentation build passed
- Hybrid infrastructure was successfully deployed and exercised with disposable and Okta test providers

## Known repository check

The pre-commit hook passed formatting, linting, secret scanning, SDK alignment, and the root production audit, then failed on the existing frontend React Router advisory also reported on the default branch. This change does not modify React Router. The commit was created with `--no-verify` after the checks above completed manually.


docs/getting-started/enterprise-sso.md
```diff
@@ -0,0 +1,361 @@
...
+Using Entra app roles avoids group-overage behavior and gives tokens stable,
+application-specific values.
+
+1. In **Microsoft Entra admin center**, open **App registrations** and create an
+   app registration for AI-DLC in the required tenant.
+2. Under **Authentication**, add a **Web** redirect URI equal to the
+   **OIDC callback** shown by `bash /tmp/aidlc-install.sh status`.
+
+3. Under **Certificates & secrets**, create a client secret. Store its **Value**
+   immediately in Secrets Manager; the Secret ID is not the client secret.
```
jeromevdl: Where to store it? Which secret name?

docs/getting-started/enterprise-sso.md
```diff
@@ -0,0 +1,361 @@
...
+   **Groups**. Include it in ID tokens for any scope (or for every scope listed
+   in the provider file), and filter it to the groups intended for AI-DLC, such
+   as names matching `^aidlc-`. Use **Token Preview** with an assigned AI-DLC
+   user and confirm the ID token contains the exact, case-sensitive group names.
+   If the Okta tenant instead presents an expression-based **Token claims** UI,
+   name the claim `groups` and use:
+   `user.getGroups({"group.profile.name":"aidlc-","operator":"STARTS_WITH"}).![name]`.
+   The claim name must match `claims.roles`; `aidlc-*` is not wildcard syntax in
+   this expression.
+6. Copy the client ID and secret, and store the raw secret in Secrets Manager.
```
jeromevdl: Same remark, what secret name?

JWThewes: It's referenced later in the idp.json file
But yeah a good one I'll make clear how it works

frontend/src/components/admin/UserManagementCard.tsx
jeromevdl, line 92: Maybe we should completely hide this when user doesn't have permission, instead of showing 'access denied ' below.

frontend/src/contexts/AuthContext.tsx
```diff
@@ -88,9 +90,31 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
...
       setIsLoading(false);
     }
-  };
+  }, []);
 
-  const setDisplayName = async (name: string) => {
+  const loginWithSso = useCallback(async (providerName: string, returnTo: string) => {
+    setIsLoading(true);
+    try {
+      await authService.loginWithSso(providerName, returnTo);
```
jeromevdl, line 98: Missing the  `setIsLoading(false)`

jeromevdl: Use a finally block like above

frontend/src/contexts/AuthContext.tsx
```diff
@@ -88,9 +90,31 @@ export const AuthProvider: React.FC<AuthProviderProps> = ({ children }) => {
...
+    } catch (error) {
+      setIsLoading(false);
+      throw error;
+    }
+  }, []);
+
+  const completeSsoLogin = useCallback(async () => {
+    setIsLoading(true);
+    try {
```
jeromevdl, line 106: No catch? At least to log an error?

frontend/src/pages/AuthCallback.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { useAuth } from '@/contexts/AuthContext';
+import { isSsoAccessDeniedError, SSO_ACCESS_DENIED_MESSAGE } from '@/services/authErrors';
+
+interface CallbackFailure {
+  title: string;
+  message: string;
+}
+
+const accessDeniedFailure: CallbackFailure = {
+  title: 'Access denied',
```
jeromevdl, line 19: Why message is a const imported and title is different?

frontend/src/pages/Login.test.tsx
```diff
@@ -0,0 +1,95 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import { MemoryRouter } from 'react-router-dom';
+
+const mocks = vi.hoisted(() => ({
+  configuration: {
+    mode: 'local' as 'local' | 'hybrid' | 'sso-only',
```
jeromevdl: Could we eventually have an enum to replace this everywhere?

frontend/src/pages/Login.tsx
```diff
@@ -48,6 +52,17 @@ export default function Login() {
     }
   };
 
+  const handleSsoLogin = async (providerName: string) => {
+    setError('');
+    setIsSubmitting(true);
+    try {
+      await loginWithSso(providerName, from);
```
jeromevdl, line 59: Should we clean the submitting ?

frontend/src/pages/Login.tsx
```diff
@@ -202,45 +217,95 @@ export default function Login() {
             </div>
           </div>
           <CardTitle>Sign in to AI-DLC</CardTitle>
-          <CardDescription>Collaborative AI-Driven Development</CardDescription>
+          <CardDescription>
+            {ssoEnabled ? 'Use your organization account' : 'Collaborative AI-Driven Development'}
```
jeromevdl: Weird... Why mixing sso and collaborative AIDLC title?

frontend/src/services/auth.ts
```diff
@@ -1,32 +1,109 @@
...
   fetchUserAttributes,
   confirmSignIn,
+  signInWithRedirect,
   updateUserAttributes,
 } from 'aws-amplify/auth';
+import { cognitoUserPoolsTokenProvider } from 'aws-amplify/auth/cognito';
 import { clearPersistedCache } from '@/lib/persistentCache';
+import { normalizeSsoLoginError } from './authErrors';
+
+export type AuthMode = 'local' | 'hybrid' | 'sso-only';
```
jeromevdl, line 18: Here it is...

jeromevdl (commented):
Did a quick read through in front of the swimming pool. Not a real review, just left a few comments.

docs/getting-started/enterprise-sso.md
JWThewes: Good point. I made this explicit in 32e0e8c: create one secret per OIDC provider in the deployment account and Region, choose any name (the guide uses `aidlc/<environment>/<provider-name>`), and copy the returned ARN into `clientSecretArn`.

docs/getting-started/enterprise-sso.md
JWThewes: Updated this one too in 32e0e8c. The Okta example now gives `aidlc/<environment>/okta` as a sample name and tells the reader to put the returned ARN in `clientSecretArn`; the installer does not expect a fixed secret name.

frontend/src/components/admin/UserManagementCard.tsx
JWThewes: I looked at the two directory contracts here. I think we should keep these rows in the platform-admin view: it intentionally lists every Cognito identity so an admin can diagnose a failed required-claim mapping. Denied and disabled identities are already excluded from the project member picker. Hiding them here as well would remove the only in-product visibility into why a previously seen identity cannot be assigned, so I left this behavior as-is.

frontend/src/contexts/AuthContext.tsx
JWThewes: You are right. I changed this to a `finally` block in 32e0e8c so the context loading state is always released when navigation does not take over, and added a provider-level regression test for the failure path.

JWThewes: Done. `loginWithSso` now clears `isLoading` in `finally`, while redirect errors still propagate to the caller.

frontend/src/contexts/AuthContext.tsx
JWThewes: This error is caught and logged by `AuthCallback`, which is the only caller, and that catch also maps it to the user-facing failure state. Logging it again here would produce two entries for one failure, so I kept the `finally` for cleanup and let the error propagate.

frontend/src/pages/AuthCallback.tsx
JWThewes: The message is shared because it is also the canonical message carried by `SsoAccessDeniedError`. `Access denied` is presentation text used only on this callback page. Moving that title into the service layer would couple page copy to the domain error, so I kept the split.

frontend/src/pages/Login.test.tsx
JWThewes: Yes. There was already an `AuthMode` union in `auth.ts`, as you spotted below. In 32e0e8c the login test imports that and `SsoProvider` instead of repeating either set of literals. I kept the union rather than adding an enum because we do not need a runtime object here.

frontend/src/pages/Login.tsx
JWThewes: Yes. The handler now resets `isSubmitting` in `finally`, and the test covers a redirect-start failure so the provider buttons do not remain disabled.

frontend/src/pages/Login.tsx
JWThewes: Agreed, the provider section already communicates the SSO choice. I restored the same `Collaborative AI-Driven Development` subtitle for every authentication mode.

frontend/src/services/auth.ts
JWThewes: Exactly. The test now imports this existing union instead of repeating the literals. I kept it as a union rather than a TypeScript enum because it is only needed at compile time.

JWThewes:
While I highly appreciate that you should enjoy your time off and spent more time in the pool ;)
Thanks for your comments!
Enjoy the time off!

JWThewes:
Thanks for taking a look. I pushed 32e0e8c and replied inline to every point. The concrete fixes are in; I left the three design-choice threads open so we can continue those discussions rather than marking them resolved without agreement.

terraform/modules/auth/main.tf
```diff
@@ -77,14 +204,31 @@ resource "aws_cognito_user_pool_client" "main" {
...
 
+  allowed_oauth_flows_user_pool_client = local.sso_enabled
+  allowed_oauth_flows                  = local.sso_enabled ? ["code"] : []
+  # The Cognito self-service scope authorizes GetUser/UpdateUserAttributes. It
+  # does not grant application or user-pool administration privileges.
+  allowed_oauth_scopes = local.sso_enabled ? [
+    "openid",
+    "email",
+    "profile",
+    "aws.cognito.signin.user.admin",
```
eipasteur: The IdP role claim is mapped to mutable `custom:sso_roles`, the app client grants `aws.cognito.signin.user.admin`, and no explicit `write_attributes` allow-list is configured. The pre-token Lambda then trusts this stored attribute to derive `platform-admin`.

This needs to be verified before merge:

- If the attribute is not writable by the app client, the IdP mapping may silently fail and SSO roles will never populate.
- If it is writable, a federated user may be able to call `UpdateUserAttributes` with the value mapped to `platform-admin`, then receive that role on token refresh.

The unit tests inject `custom:sso_roles` directly into fake events, so they do not cover either Cognito path. Were you able to verify this with the manual IdP test? If the mapping works, please add an explicit `write_attributes` allow-list that excludes `custom:sso_roles`; if that breaks federation, the role source needs to be moved out of a user-writable attribute.

frontend/src/services/authErrors.ts
```diff
@@ -0,0 +1,63 @@
...
+    return true;
+  }
+
+  const details = errorText(error);
+  return (
+    /\bSSO_ACCESS_DENIED\b/i.test(details) ||
+    /\baccess_denied\b/i.test(details) ||
+    // Cognito reduces pre-token Lambda rejections to this OAuth error at its
+    // hosted token endpoint instead of returning the Lambda's error marker.
+    /\binvalid_grant\b/i.test(details) ||
```
eipasteur: `invalid_grant` is mapped unconditionally to `SsoAccessDeniedError`, but Cognito also uses it for an expired or replayed authorization code. A user who takes too long on the IdP page, or uses the browser back button, is therefore told that their enterprise account is unauthorized and that they should contact their IdP administrator, when retrying would resolve the issue.

Could the message be softened when the error does not specifically indicate `SSO_ACCESS_DENIED` or `access_denied`, for example by showing a generic sign-in failure with a retry action?

frontend/src/services/auth.ts
```diff
@@ -83,6 +165,51 @@ export const authService = {
...
+    clearPersistedCache();
+    await signInWithRedirect({ provider: { custom: providerName } });
+  },
+
+  async completeSsoLogin(): Promise<User> {
+    // Amplify's OAuth listener completes the code exchange asynchronously when
+    // the auth module loads. Wait briefly for the token store to become ready.
+    let lastError: unknown;
+    for (let attempt = 0; attempt < 40; attempt++) {
+      if (ssoRedirectFailure !== undefined) {
```
eipasteur, line 187: Non-blocking: `completeSsoLogin` stops polling after 40 × 250 ms = 10 seconds. If the Amplify Hub event arrives after that window because of a slow network or a cold start, the user gets “Enterprise sign-in did not complete” and the partial session is cleared.

Maybe we should provide an explicit retry action after the timeout so the user can restart the SSO flow without manually reloading the page. A test covering the timeout path would also make this behavior explicit.

terraform/modules/auth/main.tf
JWThewes: Good catch. This was a real escalation path with the original scope. Cognito requires IdP-mapped attributes to remain app-client writable, so simply excluding custom:sso_roles would break the mapping. I fixed the other side of the boundary: hosted SSO sessions now authorize only openid, email, and profile, so their access tokens cannot call UpdateUserAttributes. The frontend reads SSO profile data from the signed ID token, and the app client now has an explicit write allow-list. I also added a regression test showing that a role value on a local user is ignored because the Cognito-owned identities attribute is absent. I did not redeploy the disposable IdP fixture for this follow-up, so I am leaving this thread open for confirmation of the approach.

frontend/src/services/authErrors.ts
JWThewes: You are right; invalid_grant is too ambiguous to present as an authorization decision. Bare invalid_grant and generic pre-token failures now use the normal sign-in failure path with a retry action. Only explicit SSO_ACCESS_DENIED or access_denied markers show the access-denied guidance.

frontend/src/services/auth.ts
JWThewes: Added. A callback that reaches the 10-second limit now raises an explicit timeout result, clears the partial session, and shows a Try again action that returns to sign-in. The service timeout and callback retry path are both covered by tests.

JWThewes:
@eipasteur addressed your comments

eipasteur (dismissed):

eipasteur (approved):

PR #376, opened 2026-08-03
fix(agentcore): isolate Codex rollout storage

### LINKED ISSUE #375: [Bug]: Codex rollouts saturate AgentCore managed storage
## Description

AgentCore Codex stage runs currently materialize `CODEX_HOME` directly under the 1 GiB managed workspace mount. Codex continuously appends rollout JSONL there. The mount write-back pipeline can saturate independently of reported free bytes, causing Codex runs and later resumes to fail.

Live Codex state should remain on ephemeral local disk. Only the rollout needed to resume the active stage should be copied to durable storage after Codex exits.

## Steps to reproduce

1. Run a sufficiently long Codex-backed intent stage in AgentCore.
2. Allow Codex to continuously append its rollout under `/mnt/workspace/.aidlc/codex-home`.
3. Observe managed-storage write failures even when filesystem free-space reporting does not show a full volume.
4. Attempt to resume the failed Codex thread.

## Logs

The failed run reported:

- `Write failed: waiting to be backed up`
- Codex `StorageFull`
- `ENOSPC` while closing the rollout during resume

## Expected behavior

Codex execution should not continuously write its session state to the managed mount. Fresh runs and resumes should use isolated local homes, while resumable stage rollouts remain durable across microVM/container replacement.

## Acceptance criteria

- Run each Codex invocation with a scoped, ephemeral local `CODEX_HOME`.
- Restore only the rollout matching the persisted thread ID before a true resume.
- Atomically persist only that rollout after every stage-author leg, with bounded retry for transient write-back pressure.
- Preserve the existing durable `sessions/...` layout for pre-existing threads.
- Demote missing or corrupt resumes to the existing fresh-recovery path.
- Never advertise a parked gate as resumable when rollout persistence failed.
- Keep reviewer, conflict-resolution, and one-shot Codex sessions ephemeral.
- Cover fresh, resume, failure, retry, and cold-container behavior with tests.

## Environment

- Bedrock AgentCore Runtime
- AWS Region: `us-east-1`
- Codex CLI `0.145.0`

JWThewes:
## Summary

- run Codex with scoped ephemeral `CODEX_HOME` directories instead of writing live rollouts directly to the AgentCore managed mount
- restore and atomically persist only the active stage thread rollout, preserving the legacy durable session layout
- demote missing/corrupt resumes to fresh recovery and retire parked gates when durable persistence fails
- keep reviewer, conflict-resolution, discussion, and one-shot Codex state ephemeral

## Root cause

Codex continuously appended rollout JSONL directly under `/mnt/workspace/.aidlc/codex-home`. AgentCore managed storage can saturate its write-back pipeline independently of reported free bytes, producing `StorageFull`, `Write failed: waiting to be backed up`, and `ENOSPC` failures.

## Implementation

- add a validated single-rollout store with atomic replacement and bounded retry for transient mount pressure
- add scoped local Codex homes and runtime environment configuration
- restore before true resume, persist after every author leg, and prevent stale transcripts from being resumed after a failed replace
- add cold-container and non-resumable-surface coverage

## Validation

- `npx vitest run --project=agentcore` (45 files, 766 tests passed)
- `npm --prefix lambda/agentcore run format:check`
- `npm --prefix lambda/agentcore run lint` (no errors; one unrelated pre-existing warning)
- commit hooks: formatting, lint, secret scan, dependency audit, and staged test suite passed

The manual credentialed Codex E2E was not run because it requires live model credentials. Its harness now verifies an empty local home on cold resume and exactly one durable rollout file.

Fixes #375


lambda/agentcore/commands/run-stage.js
```diff
@@ -2137,6 +2278,29 @@ export const runStage = async (
       `${cli === 'codex' ? 'Codex' : 'OpenCode'} parked the stage without emitting a session id; the conversation cannot be resumed`,
     );
   }
+  if (parked && cli === 'codex' && codexStoreConfigured && !codexPersistResult?.ok) {
+    await (
+      store.supersedeHumanTask?.({
+        executionId,
+        humanTaskId: parked.humanTaskId,
```
eipasteur: **Blocking: Do not swallow gate-retirement failures**

If `supersedeHumanTask` throws, this catch allows `fail(..., { clearPending: true })` to run while the human-task row may remain pending. The user can then answer a gate that has no resumable stage, silently losing the answer. Treat a returned `null` as a benign CAS miss, but do not swallow thrown errors: either leave the stage retryable or verify that the gate is no longer pending before failing. Could we add a test covering a thrown `supersedeHumanTask` and asserting that no pending gate remains?

lambda/agentcore/cli/codex-store.js
```diff
@@ -0,0 +1,288 @@
...
+      crlfDelay: Infinity,
+    });
+    for await (const raw of lines) {
+      const line = raw.trim();
+      if (!line) continue;
+      let event;
+      try {
+        event = JSON.parse(line);
+      } catch {
+        return { ok: false, status: 'corrupt' };
```
eipasteur, line 133: **Blocking: Could we confirm the handling of truncated rollout tails?**

Could we add a regression test with a valid durable rollout followed by a truncated final JSONL line? The current validator appears to classify the entire rollout as corrupt, which may discard an otherwise valid multi-turn history and clear `cliSessionId`. If confirmed, could we preserve the valid records before the truncated tail while still rejecting malformed lines in the middle of the file?

lambda/agentcore/cli/codex-store.js
```diff
@@ -0,0 +1,288 @@
...
+      await fs.rm(temporary, { force: true });
+      await fs.copyFile(found.filename, temporary);
+      await fs.rename(temporary, destination);
+      return {
+        ok: true,
+        status: 'persisted',
+        relativePath: found.relativePath,
+        attempts: attempt,
+      };
+    } catch (error) {
```
eipasteur, line 286: **Non-blocking: Could we account for full-rollout copy bursts?**

Live Codex appends now stay on local disk, but each parked leg copies the entire growing rollout to the managed mount, with up to five full copies under retry. Long conversations can therefore still produce large write-back bursts on the same mount, so the original saturation risk is reduced rather than eliminated. Could we document this limitation or consider incremental persistence for the durable copy?

lambda/agentcore/cli/codex-store.js
```diff
@@ -0,0 +1,288 @@
...
+    attempts = attempt;
+    const temporary = path.join(
+      path.dirname(destination),
+      `.${path.basename(destination)}.${process.pid}.${ids()}.tmp`,
+    );
+    try {
+      await fs.mkdir(path.dirname(destination), { recursive: true });
+      await fs.rm(temporary, { force: true });
+      await fs.copyFile(found.filename, temporary);
+      await fs.rename(temporary, destination);
```
eipasteur, line 279: **Non-blocking: Could we document the durability guarantee for parked rollouts?**

`rename()` provides an atomic namespace swap, but it does not guarantee that the data survives container loss or an asynchronous managed-mount backup failure. Since `cliSessionId` is persisted as resumable immediately after `persistCodexRollout` succeeds, could we document the AgentCore durability contract relied on here? If that contract does not cover this failure mode, an S3-backed fallback for parked rollouts may be worth considering. `fsync` would be useful hygiene, but would not address backup durability.

eipasteur (approved):

PR #348, opened 2026-07-23
fix: forward proxy settings to Docker builds

JWThewes:
## Summary

- detect non-empty standard proxy variables inherited through the central `scripts/install.sh` workflow and pass them as sensitive Terraform Docker build arguments
- forward build arguments to both AgentCore and Yjs Buildx builds while keeping the default non-proxy path unchanged and allowing explicit `TF_VAR_docker_build_args` overrides
- make the local AgentCore E2E harness proxy-aware and document managed-install, manual-deployment, state, and E2E behavior
- add focused coverage for proxy and non-proxy environments without exposing proxy values in Docker command arguments

## Testing

- `node --test scripts/test/docker-proxy-build-args.test.mjs scripts/test/agent-e2e-proxy.test.mjs` (5 passed)
- live `E2E_CLIS=claude,opencode ./scripts/agent-e2e-testing.sh` (Claude passed, OpenCode passed; Kiro not selected because no key was available)
- Terraform validation using a backend-free copy
- `terraform fmt -check -recursive terraform`
- `bash -n scripts/deploy-terraform.sh scripts/agent-e2e-testing.sh scripts/install.sh`
- `npm run lint`
- `npm run format:check`
- `npm run secretlint`
- `git diff --check`

`npm run test:release` has one unrelated existing failure: the release test expects a `TBD` changelog date while the current changelog entry is dated. The pre-commit npm audit also reports existing dependency advisories; staged lint, format, and secret checks passed.

Fixes #347

PR #382, opened 2026-08-06
docs: redesign README around a human-first opening

eipasteur:
## What

Restructures the README so the first screen answers what the platform is, who it is for, and why it is different, before the operator reference. The full technical content (quickstart, installer, operations, testing) is preserved and stays current with main.

### Opening
- Typographic hero (dark/light SVG via `<picture>`) replacing the plain H1
- Tagline naming the supported agents: Kiro, Claude Code, OpenCode, Codex
- Value proposition and docs/deploy/status anchor row above the fold
- Linked capability matrix (gates, collaboration, traceability, cost) ahead of the visual proof

### Visual proof
- Four cropped product screenshots paired with the claims they support, sized to stay legible at rendered width
- Demo GIF rebuilt from the recorded clips: 15s, 800x450, 10fps, 2.4MB, now includes the pull request and traceability drawer beats; kept collapsed in `<details>` so it never dominates the fold

### Housekeeping
- Scoped `.gitignore` exceptions so the referenced README assets are committable under the global `*.png` rule
- Rebased on main; keeps the proxy, enterprise SSO, and Bitbucket additions from #348, #366, #342

## Why

The previous README opened with the operator reference. Visitors deciding in a few seconds whether the project is relevant had to scroll to understand what it does. This change front-loads comprehension and proof, and leaves the reference intact for operators and agents.

## Verification

- Rendered through the GitHub Markdown API and inspected at 1440px and 390px: no horizontal overflow, correct hero variant per theme and viewport
- All local references resolve and are tracked (checked against `.gitignore`, not just the filesystem); 22 external links return 2xx/3xx
- `oxfmt`, `xmllint` on the SVGs, `git diff --check` all clean
- Demo GIF beats verified frame by frame at true output resolution


JWThewes:
Nice work overall — the new opening does a much better job explaining what this actually is, and I checked the technical claims against the code (admin tab names, SSM SecureString storage, the plan destroy-guard, the AIDLC_* env var table): all accurate. The removed sections are all properly relocated in setup.md too, and switching the links off aidlc.dev was the right call since that domain now serves an unrelated site.

A few things I'd like to see fixed before merge:

1. **The anonymous testimonial has to go (or get a name).** "— Principal Solutions Architect at AWS" with no name is an unverifiable internal endorsement in an aws-samples repo. Either get a named quote with the person's OK, or drop the blockquote.

2. **Bitbucket credentials don't go in Admin → Trackers.** The intro to "Provider OAuth apps" says all providers paste credentials into **Admin → Trackers**, but Bitbucket is configured under **Admin → Source Control → Bitbucket** (the old README had this right, and setup.md still says so). Suggest: "…into **Admin → Trackers** (GitHub Issues, GitLab, Jira) or **Admin → Source Control** (Bitbucket, GitHub App)."

3. **The demo GIF loads for every visitor despite the `<details>` collapse** — browsers fetch `` sources inside closed `<details>` anyway, so that's 2.4 MB on every README view, plus it sits in git history forever and will need replacing whenever the UI changes. Consider hosting it on the docs site or as a release asset and linking it instead. The traceability PNG (551 KB) and collaboration PNG (314 KB) could also lose a lot of weight with pngquant/oxipng.

4. **Crop the cost card.** `cost-observability-card.png` is 1600px wide but mostly whitespace with one thin row of numbers — at 50% table width it'll be unreadable. Crop to the "Usage & activity" card itself. Related: the live-collaboration screenshot shows two messages from the same user (alex@demo.dev), which undercuts the "whole team, one intent" caption — a shot with two distinct participants would sell it better.

Smaller stuff, take or leave:

- The four `.gitignore` exceptions could be one line: `!docs/assets/readme/*.png` — saves editing it for every future screenshot.
- The README now has no h1 at all (starts at an `<h3>`). GitHub doesn't care but screen readers and other renderers do; a visually-hidden h1 or making the tagline a `<p>` would fix it.
- The centered `Why · Quickstart · Workflow · Docs` nav and the `**Documentation** · **Deploy** · **Status**` row right below it overlap — one can go.
- The "Early Preview" shields badge is static text that'll silently go stale at GA.
- Both new SVGs are missing a trailing newline.
- Not this PR, but worth a follow-up: `zensical.toml` still has `site_url = "https://aidlc.dev/"`, so the docs site's canonical URLs point at that unrelated site now.

Happy to approve once 1 and 2 are addressed.

eipasteur:
All addressed in 6f144cb9.

1. Testimonial removed.

2. Bitbucket path fixed with your wording, in the README and in setup.md's provider intro, which had the same generalisation and contradicted its own Bitbucket step 4.

3. The README now embeds an 82 KB poster frame linking to the GIF, so viewing the README no longer fetches 2.4 MB. The file stays in-repo since the docs site builds from this repo anyway; nothing auto-loads it.

4. Cost card cropped border-to-border and moved to a full-width strip. On the collaboration shot: the captured footage has no frame with a message from a second author, so it's now Sam's session displaying Alex's messages, which is the actual cross-user sync, with the avatar chip intact and the crop bleed removed; caption adjusted. Also replaced the traceability image, which showed two detail drawers, with the actual intent graph at full width.

Smaller items all taken: wildcard .gitignore exception, hero wrapped in an h1, duplicate nav row dropped, release badge now dynamic (renders v2.0.0-preview0). The hero SVGs are replaced by 2x PNGs, which removes the trailing-newline issue and a portability bug on top: the SVG wordmark used live text resolved against the viewer's font stack. zensical site_url: #383.

JWThewes (approved):

PR #393, opened 2026-08-10
feat: link generated PRs back to AI-DLC intents

jeromevdl:
## Summary
- link the AI-DLC attribution in generated PR and MR bodies back to the originating intent
- pass the canonical application URL to the v2 orchestrator
- remove redundant project and execution ID lines

## Testing
- deployed and verified that generated GitHub PRs link correctly to their AI-DLC intents
- 124 repository hook tests passed
- orchestrator suite: 88 tests passed
- formatting, lint, and secret scanning passed

### DIFF
--- a/lambda/v2-orchestrator/index.js
+++ b/lambda/v2-orchestrator/index.js
@@ -49,6 +49,7 @@ import {
   fanoutGateAddendum,
 } from './section.js';
 import { runQuorumEdit } from './quorum-edit.js';
+import { buildIntentAttribution } from './pr-attribution.js';
 
 const ddb = DynamoDBDocumentClient.from(new DynamoDBClient({}));
 const lambda = new LambdaClient({});
@@ -904,6 +905,7 @@ const handler = async (event, ctx, deps = defaultDeps()) => {
       deriveEnrichment,
       prStrategy: meta.prStrategy ?? 'intent-pr',
       unitPrProvider,
+      applicationUrl,
       stageInstanceIdFor: (stageId, slug, sectionIndex = null) =>
         planStageInstanceId(namespace, stageId, slug, sectionIndex),
     };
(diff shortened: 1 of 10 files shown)


lambda/v2-orchestrator/index.js
```diff
@@ -1355,6 +1358,7 @@ const handler = async (event, ctx, deps = defaultDeps()) => {
         meta,
         executionId,
         gitProvider,
+        applicationUrl,
```
JWThewes, line 1363: Are we intentionally limiting this to the final fan-in PR? The `pr-per-unit` path builds draft PR bodies in `section.js` and still uses plain `AI-DLC`, so those generated PRs/MRs won't link back to the intent. Since the PR description says generated PR/MR bodies, I'd expect the URL to be threaded into the section toolkit too, ideally through one small helper used by both paths.

lambda/v2-orchestrator/index.js
```diff
@@ -1697,11 +1702,14 @@ const openIntentPrs = async ({
   }
 
   const title = meta.title || `AI-DLC: ${branch}`;
+  const intentUrl = applicationUrl
+    ? `${applicationUrl.replace(/\/+$/, '')}/space/${encodeURIComponent(
+        meta.projectId,
+      )}/intent/${encodeURIComponent(meta.intentId ?? executionId)}`
+    : null;
+  const aidlcAttribution = intentUrl ? `[AI-DLC](${intentUrl})` : 'AI-DLC';
```
JWThewes: What happens if `APPLICATION_URL` is missing or empty? We fall back to plain `AI-DLC`, but this change also removes the old Project/Execution lines, so that path loses all provenance. Could we either treat the URL as required or keep those lines when `intentUrl` is null? A small fallback test would help.

lambda/v2-orchestrator/index.js
```diff
@@ -1641,6 +1645,7 @@ const openIntentPrs = async ({
   meta,
   executionId,
   gitProvider,
+  applicationUrl,
```
JWThewes, line 1650: Small cleanup: the comment just above still says the PR body carries the execution ID, but the body no longer does. Can we update it to describe the intent link instead?

JWThewes (changes requested):

lambda/v2-orchestrator/index.js
jeromevdl: Added a shared helper to handle both pr modes.

lambda/v2-orchestrator/index.js
jeromevdl: clean

lambda/v2-orchestrator/index.js
jeromevdl: APPLICATION_URL is not supposed to be null, changing

JWThewes (approved):

PR #397, opened 2026-08-11
feat: add managed tool catalog and build environments

### LINKED ISSUE #390: [RFC]: Managed build environments for agent verification
## Summary

Add managed, immutable build environments for agent verification. The platform ships one protected Standard environment and managed tool definitions rather than separate JVM, Go, Rust, or Polyglot environments. Platform administrators publish exact tool versions, compose them into environments, and project owners assign published environments. Every intent snapshots the exact runtime and toolchain it started with.

## Motivation

Agent verification needs predictable toolchains, controlled operating-system packages, immutable provenance, and a stable runtime endpoint across invocation, parallel execution, rewind, cancellation, and session shutdown. Administrators also need to add tools such as .NET without changing platform source or maintaining an opinionated environment for every language combination.

## Environment model

Standard contains the protected AgentCore runtime plus Node.js and Python. New environments extend one published base revision and may add:

- exact published tool versions
- exact apt packages
- non-secret environment variables
- restricted ordered build commands

Environment revisions are immutable. They snapshot their pinned base digest, resolved tool-version definitions and OCI digests, generated Dockerfile, image digest, runtime version, endpoint, compatibility version, verification evidence, findings, and acceptance audit.

Project owners assign one published environment. Projects without an explicit assignment use Standard. New intents follow the environment's current published revision; existing intents remain pinned.

## Managed tool catalog

Store tool families and immutable tool versions in the environment registry.

A tool family stores its identity, display metadata, category, dependencies, ownership, and explicit recommended-version pointer. A tool version stores:

- exact version and ARM64 source
- original and resolved public URLs
- retained source digest and trust level
- installation definition
- exposed executables
- apt and environment requirements
- version and functional verification
- OCI artifact digest and compressed size
- SBOM, scan findings, logs, and audit fields
- protected core digest and runtime compatibility used for validation

Seed Java `21.0.8`, Go `1.24.6`, Rust `1.89.0`, Maven `3.9.11`, and Gradle `9.0.0` as tool definitions. Their initial versions build automatically but require administrator review and publication. Multiple versions may remain published. Publishing does not make a version recommended automatically.

Administrators may add versions to seeded families or create families such as `.NET SDK`.

## Tool creation and verification

Normal creation asks for a name, exact version, public HTTPS ARM64 archive, and verification preset.

The platform:

1. Resolves and validates public redirects.
2. Rejects credential-bearing URLs and private, loopback, metadata, or link-local destinations.
3. Downloads the source in a credentialless, private-network-blocked container.
4. Enforces source, evidence, extraction, and normalized-output limits.
5. Rejects archive traversal, unsafe symlinks, and unsupported layouts.
6. Computes and retains a content-addressed source digest.
7. Normalizes the payload into a dedicated immutable ARM64 OCI artifact.
8. Generates an SBOM and records compressed size.
9. Runs version and representative functional checks as the non-root runtime user.
10. Scans the resulting artifact and retains all evidence.

`Publisher verified` requires an independently fetched publisher checksum that matches the imported source. Otherwise the catalog records `Platform pinned`: the retained artifact is immutable, but no independent publisher checksum was supplied.

Presets cover Java, Go, Rust, Maven, Gradle, .NET, and generic CLI tools. Advanced Bash installers run without AWS credentials, metadata access, private-network access, the Docker socket, host mounts, or elevated capabilities. Public internet access is available to the installer, so the published OCI artifact is immutable and verified even when rerunning the installer is not fully reproducible.

Custom functional verification is networkless and supports bounded fixture files.

## Dependencies and composition

Environment recipes reference exact `{toolId, toolVersionId}` pairs. The service resolves and snapshots the complete dependency graph before build.

Maven and Gradle depend on a published recommended Java version. The API rejects:

- missing or cyclic dependencies
- binary-name collisions
- environment-variable conflicts
- conflicting exact apt package versions
- incompatible base revisions
- projected images above the AgentCore size limit

Generated Dockerfiles copy tool payloads from exact OCI digests with multi-stage builds. The final environment is scanned and re-runs every selected tool's version and functional verification against the selected base.

Changing a recommended tool version marks affected environments with a structured update warning. Existing environment revisions and active intents remain pinned. Administrators explicitly edit, build, review, and publish replacements.

## Security review

Critical and High findings enter `SECURITY_REVIEW`; a successful build is not represented as an image-build failure. Platform administrators may explicitly accept findings and continue validation. Findings, severity counts, affected packages, advisory references, acceptance identity, and acceptance time remain visible on ready and published tool and environment revisions.

Only revisions that pass installation, architecture, protected-runtime, non-root, smoke, functional, and AgentCore runtime validation may become ready for publication.

## APIs and user experience

Platform Settings provides:

- **Tools** for family/version creation, source provenance, builds, logs, SBOM, findings, acceptance, publication, and recommendation
- **Environments** for immutable tool composition, base updates, projected size, builds, verification, publication, and retirement

Project Settings provides a published-environment selector, included tools, and repository compatibility warnings, including .NET repositories. Intent and audit views display exact snapshotted runtime and toolchain evidence.

## Compatibility and retention

The platform supports the current and previous runtime compatibility versions. Images, tool artifacts, and endpoints remain available while referenced. Immutable tool artifacts and environment revisions do not follow mutable tags or recommendations.

Historical schema-v1 recipes remain readable for audit but cannot be built, retried, or published. The feature was not released before the tool-catalog model, so no automated destructive migration or reset capability is shipped.

Retirement and reversible archiving remain outside this change.

## Operational costs

The design adds DynamoDB records, S3 source and build contexts, separate ECR tool artifacts, ECR scans, ARM64 CodeBuild minutes, CloudWatch logs, Lambda invocations, and AgentCore endpoints. Build roles are isolated by responsibility. Source retention is content-addressed, contexts expire, and non-production repositories are force-deleted only during stack teardown.

## Alternatives considered

- Ship one predefined environment per language combination. Rejected because combinations grow rapidly and constrain administrator choice.
- Accept arbitrary administrator Dockerfiles. Rejected because they can replace protected runtime behavior and weaken provenance.
- Install arbitrary tools directly during every environment build. Rejected because downloads, normalization, scanning, and verification would be duplicated and harder to audit.
- Automatically recommend or publish every successful tool version. Rejected because tool updates require explicit administrator review.
- Move existing intents to newer environments. Rejected because active verification must remain reproducible.
- Ship an automated destructive reset. Rejected because the feature had not been released and the sole development deployment could be cleaned manually.

## Acceptance criteria

- Standard is the only seeded environment.
- Java, Go, Rust, Maven, and Gradle are seeded as independently buildable, reviewable, publishable tool versions.
- An administrator can import and publish a .NET SDK without changing platform source.
- Tool sources, checksums, redirects, archives, installers, OCI artifacts, SBOMs, findings, sizes, and verification evidence are validated and retained.
- Environments reference exact immutable tool versions and resolve dependencies and conflicts deterministically.
- Recommended-version changes warn affected environments without changing published revisions or active intents.
- Security findings remain visible and may be accepted explicitly without bypassing functional or runtime validation.
- Every intent invocation and shutdown path uses its snapshotted runtime endpoint.
- Platform and project settings expose catalog management, environment composition, assignment, compatibility, and audit evidence.
- No reset API, Lambda, IAM role, route, or UI is shipped.
- Formatting, lint, frontend type checking/build/tests, Lambda tests/bundles, Terraform formatting/validation, secret scanning, CodeQL, and production dependency audits pass.

JWThewes:
## Summary

- ship one protected Standard environment with Node.js and Python, plus managed Java, Go, Rust, Maven, and Gradle tool definitions
- let platform administrators import additional ARM64 tools such as .NET from public HTTPS archives without changing platform source
- normalize each verified tool version into an immutable OCI artifact with retained provenance, SBOM, size, scan findings, and functional verification evidence
- compose exact published tool-version digests into immutable environment revisions with dependency resolution, conflict checks, and projected AgentCore image-size enforcement
- preserve reviewed Critical and High findings through publication after explicit administrator acceptance
- snapshot the exact environment revision, image digest, runtime version, endpoint, compatibility version, tools, and verification result into every intent
- add Tools and Environments views to Platform Settings, plus project assignment and repository compatibility warnings

## Tool ingestion

- validate public HTTPS sources and every redirect
- compute and retain content-addressed source digests
- distinguish `Publisher verified` evidence from `Platform pinned` imports
- reject unsafe archives, traversal, extraction bombs, unsafe symlinks, credential-bearing URLs, and oversized payloads
- run advanced installers in a credentialless container without metadata, private-network, Docker-socket, or host access
- run version and functional verification networklessly as the non-root runtime user
- support Java, Go, Rust, Maven, Gradle, .NET, generic CLI, and bounded custom verification fixtures

## Compatibility

- Standard is the only seeded environment
- new and rebuilt environments use catalog-backed schema-v2 recipes
- historical schema-v1 records remain readable for audit but cannot be built, retried, or published
- no reset API, Lambda, IAM role, route, or Platform Settings view is shipped
- the sole development account's legacy non-Standard runtimes, images, lookups, revisions, environments, and migration marker were removed manually

## Verification

- complete repository suite: 140 files, 2,591 tests passed
- complete frontend suite: 73 files, 482 tests passed
- managed tool and environment control plane: 10 files, 79 tests passed
- `npm run format:check`
- `npm run lint`
- `npm run secretlint`
- `npm run sdk:check`
- `npm --prefix frontend run typecheck`
- `npm --prefix frontend run build`
- all remaining environment Lambda handlers bundled successfully
- `terraform fmt -recursive -check terraform`
- Terraform provider-schema validation with a local backend; only the two existing DynamoDB deprecation warnings remain
- root and frontend production dependency audits report zero vulnerabilities

Terraform was not planned or applied against AWS. The deployed-stack workflow and .NET catalog example are documented in `docs/development/testing.md`.

Closes #390


lambda/shared/environment-snapshot.js
```diff
@@ -0,0 +1,93 @@
...
+  return {
+    environmentId: environment.environmentId,
+    name: environment.name,
+    revisionId: revision.revisionId,
+    imageDigest: revision.imageDigest,
+    runtimeVersion: revision.runtimeVersion ?? null,
+    runtimeArn: revision.runtimeArn,
+    runtimeEndpoint: revision.runtimeEndpoint ?? null,
+    compatibilityVersion,
+    verification: revision.verification ?? null,
```
jeromevdl, line 113:
```suggestion
    environmentId: environment.environmentId,
    name: environment.name,
    revisionId: revision.revisionId,
    imageDigest: revision.imageDigest,
    runtimeVersion: revision.runtimeVersion ?? null,
    runtimeArn: revision.runtimeArn,
    runtimeEndpoint: revision.runtimeEndpoint ?? null,
    compatibilityVersion,
    verification: revision.verification ?? null,
    tools: revision.flattenedRecipe?.resolvedTools ?? [],
```

as documented in https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/397/changes#diff-51840c6001bebb21d56b029ca1ff58b642117adcdf734564433f7c6e8051cdabR86, we might want to also have the tools. Maybe add a test asserting those snapshots survive into the intent and audit output.

docs/development/testing.md
```diff
@@ -3,6 +3,132 @@
...
+
+1. Select Go and publish the resulting environment.
+2. Select Maven and confirm the recommended Java version is added
+   automatically.
+3. Select `.NET SDK` and publish the resulting environment.
+4. Confirm every generated Dockerfile copies tools from exact OCI digests and
+   retains the protected base entrypoint, command, user, port, and health
+   behavior.
+5. Confirm the projected and actual compressed image sizes stay below the
+   configured AgentCore image limit.
```
jeromevdl: why do we have something specific for .NET ?

jeromevdl: not clear, we mix languages, .NET seems important, I don't understand

docs/using-the-platform/managed-environments.md
jeromevdl, line 1: Quite verbose, without diagram, a bit painful to read. Can we imagine a TL;DR or the main instructions/options first and "going further" separated. Or the actionable options first and FYI later. Something to make it more digest.

docs/using-the-platform/projects.md
jeromevdl, line 3: By reading this, I see we still speak about "projects", not "spaces", created this issue for follow up: #409

lambda/shared/environment-snapshot.js
```diff
@@ -0,0 +1,93 @@
...
+  revisionId: fallback.revisionId ?? 'legacy',
+  imageDigest: fallback.imageDigest ?? null,
+  runtimeVersion: fallback.runtimeVersion ?? null,
+  runtimeArn: fallback.runtimeArn ?? null,
+  runtimeEndpoint: fallback.runtimeEndpoint ?? null,
+  compatibilityVersion: fallback.compatibilityVersion ?? '1',
+  verification: fallback.verification ?? { status: 'PASSED', source: 'legacy-runtime' },
+});
+
+const supportsCompatibilityVersion = (candidate, current) => {
```
jeromevdl: Duplication of `isSupportedCompatibilityVersion` in `recipe.js:701`

lambda/environments/status.js
```diff
@@ -0,0 +1,754 @@
...
+      });
+      if (
+        deterministic.ok !== true ||
+        deterministic.nonce !== nonce ||
+        deterministic.compatibilityVersion !== revision.runtimeCompatibilityVersion
+      ) {
+        throw new Error('deterministic runtime validation failed');
+      }
+    } finally {
+      await runtimeClient.send(
```
jeromevdl: No try catch of the `StopRuntimeSessionCommand`. A stop failure masks the real validation error; because the replacement error is retryable, a broken revision returns {pending:true} and is retried forever instead of being marked FAILED.

lambda/environments/status.js
```diff
@@ -0,0 +1,754 @@
...
+      revision.revisionId,
+      {
+        status: 'READY',
+        verification: {
+          status: 'PASSED',
+          completedAt,
+          imageBuild: 'PASSED',
+          baseDigest: 'PASSED',
+          architecture: 'arm64',
+          nonRoot: deterministic.nonRoot === true,
```
jeromevdl, line 519: The deterministic check at lines 517–523 only fails verification when `ok`, `nonce`, or `compatibilityVersion` are wrong — it never looks at `nonRoot`. Yet at line 544 verification is hard-coded to status: `'PASSED'`, and line 549 records `nonRoot: deterministic.nonRoot === true`. Because `ok` is defined as `workspaceWritable && protectedRuntime` (in managed-runtime-check.js) and excludes `nonRoot`, a runtime running as root returns `nonRoot: false` but still passes the check at 517–523 → the revision goes READY with status: `'PASSED'` and `nonRoot: false` recorded, which contradicts itself. Please either add `nonRoot === true` to the condition at 517–523, or stop recording a field verification doesn't gate on. (Low priority — non-root is already enforced at build time in recipe.js.)

lambda/intents/index.js
jeromevdl, line 4514: `resolveEnvironmentSnapshot` can throw different errors (`ENVIRONMENT_NOT_PUBLISHED`, `ENVIRONMENT_COMPATIBILITY_UNSUPPORTED`, ...) but the catch simply return an opaque 500, loosing the interesting information. Please bind the error and map these codes to a 409, the way this file already does at line 1283 for SOURCE_CONTROL_NOT_READY.

lambda/environments/tools-index.js
```diff
@@ -0,0 +1,564 @@
...
+            s3Client,
+            codebuildClient,
+          });
+        } catch (error) {
+          if (error.code === 'TOOL_RECOMMENDED_DEPENDENCY_MISSING') continue;
+          console.error(`Unable to start seeded tool build ${version.versionId}:`, error.message);
+        }
+      }
+    })().finally(() => {
+      initialization = null;
```
jeromevdl: `initialize` (lines 328...) memoizes its promise with `??=`, but nulls it in `.finally`, which runs on success too. So the next request re-runs the whole block — `seedSystemTools()`, `listVersionsByStatus('DRAFT')`, `listVersionsByStatus('FAILED')` on every call; the cache only dedupes concurrent calls, never across requests. Compare `lambda/environments/index.js:352–359`, which nulls the cache in `.catch` (error only) so a successful init persists. Please switch to the `.catch` pattern here.

lambda/projects/index.js
```diff
@@ -958,6 +960,102 @@ const handleProjectCustomRules = async (g, response, httpMethod, projectId, user
   return response(405, { error: 'Method not allowed' });
 };
 
+const environmentKey = (environmentId) => ({ pk: `ENV#${environmentId}`, sk: 'META' });
+const environmentRevisionKey = (environmentId, revisionId) => ({
+  pk: `ENV#${environmentId}`,
+  sk: `REV#${revisionId}`,
+});
+
+const readPublishedEnvironment = async (environmentId) => {
```
jeromevdl: Two code paths gate the same project environment_id differently:
 - `readPublishedEnvironment` (here), used by the assignment PUT handler, accepts on status only — line 1002: `['PUBLISHED', 'SUPERSEDED'].includes(revision.status)`. It never checks compatibility.
  - `resolveEnvironmentSnapshot` (environment-snapshot.js:32–95), used at intent-create, additionally requires `supportsCompatibilityVersion(...)` at lines 73–81.

So: publish a revision at compatibility version 1 → assign it to a project (passes) → platform advances `RUNTIME_COMPATIBILITY_VERSION` to 3 → intent-create now throws `ENVIRONMENT_COMPATIBILITY_UNSUPPORTED`. An environment that was accepted at assignment is rejected at run time. Please apply the same acceptancrules (ideally one shared resolver) in both places.

lambda/discussions/services.js
```diff
@@ -129,11 +130,24 @@ const discussionSessionIdFor = (intentId, discussionId) =>
...
-  if (!runtimeArn) throw new Error('AGENTCORE_RUNTIME_ARN is not configured');
+  const fallbackRuntimeArn = process.env.AGENTCORE_RUNTIME_ARN || '';
+  const processTable = process.env.V2_PROCESS_TABLE;
+  let target = runtimeTargetInput(null, fallbackRuntimeArn);
+  if (processTable) {
+    const { Item: meta } = await ddb.send(
+      new GetCommand({
+        TableName: processTable,
+        Key: { pk: `EXEC#${intentId}`, sk: 'META' },
+        ConsistentRead: true,
```
jeromevdl: We might not need the `ConsistentRead` here. The execution snapshot is written once at setup and is immutable per execution, so eventual consistency is safe here.

lambda/environments/catalog-recipe.js
jeromevdl, line 1: Why having 2 recipe.js (v1 and v2) ? Both are introduced in this PR so using v1 and v2 is a bit misleading and introduces some idea of legacy while there is not: 2 schemas, 2 validators, 2 Dockerfile/verification/build-context generators, 2 test suites... 
If you want to distinguish system fixed allow-list from catalog-driven, maybe be more explicit in the naming than v1 and v2.

jeromevdl (dismissed):
This feature is quite complex and there is legitimately complexity in the PR to handle this, but the PR adds 20-30% incidental complexity on top of it because of duplicated subsystems: 
- recipe v1 vs v2
- 3 different ways to evaluate if environment is ready: `readPublishedEnvironment`, `resolveEnvironmentSnapshot`, `publishedBase`
- one stack for environments, one for tools very close to the environment (createStore, createHandler, startBuild, createStatusHandler, ... in index/status/store vs tools-index/tools-status/tools-store, both modeled as: a versioned entity with DRAFT/BUILDING/READY/PUBLISHED-style revisions, built via CodeBuild, whose status is reconciled by a polling handler, fronted by a CRUD API handler backed by a DynamoDB store. That shared lifecycle is implemented twice. I've compared the shapes and signatures here, not diffed the two stacks line by line, so I can't put a precise "X% identical" number on it.

Can you check what could be merged (schema v1/v2 ? duplications of env and tools ? one environment resolution function ?) and maybe others to simplify / reduce the amount of code?

lambda/discussions/services.js
```diff
@@ -129,11 +130,24 @@ const discussionSessionIdFor = (intentId, discussionId) =>
   `aidlc-discuss-${intentId}-${discussionId}`.padEnd(33, '0');
 
 export const invokeDiscussionAssist = async ({ intentId, payload }) => {
```
jeromevdl: This PR adds V2_PROCESS_TABLE to the discussions lambda, so `invokeDiscussionAssist` now does a DynamoDB Get where it previously had zero DB dependency — and if (!meta) throw. For a normal intent the META row always exists (created at DRAFT), so it's unreachable in practice, but a throttled or IAM-denied Get now throws where before it couldn't. New blast radius on a hot interactive path

jeromevdl:
Also, agents/index.js:79-80,107-108 (fetchRuntimeCapabilities, verifyMcpServers) still hit the core AGENTCORE_RUNTIME_ARN, never a project's managed environment.

lambda/shared/environment-snapshot.js
JWThewes: Addressed in fa980b6. Published environment snapshots now include the resolved tool versions from flattenedRecipe.resolvedTools (with an empty tools list for the Standard fallback). I also added coverage for resolver output, intent META/response persistence, and audit output, and documented the field. Focused unit tests, formatting/lint, and the intents bundle build pass; the container-backed suite will run in CI.

docs/development/testing.md
JWThewes: Clarified in 1e15fbf. The guide now separates the three paths being tested: the protected Standard Node.js/Python baseline, shipped catalog-tool imports, and administrator-created tools. It also states that .NET is only the representative custom-tool case (not a privileged platform requirement) and adds a compact matrix explaining the Go, Maven+Java, and .NET environment scenarios.

docs/using-the-platform/managed-environments.md
JWThewes: Addressed in 08e0525. The page now starts with a task-oriented quick-start table linking directly to the relevant procedures, plus a diagram of the tool → environment → project → pinned-intent flow. The existing in-depth material is grouped under a separate Detailed guide section so readers can choose the short actionable path or continue into reference detail.

docs/using-the-platform/projects.md
JWThewes: Agreed. This is tracked as the documentation-wide terminology migration in #409, which is still open. I am leaving this PR on the current project/API terminology rather than partially renaming this one page and making the docs less consistent.

lambda/shared/environment-snapshot.js
JWThewes: Addressed in e2ca1de. The compatibility window now has a single implementation in lambda/shared/runtime-compatibility.js. environment-snapshot.js imports it, and recipe.js re-exports it under the existing isSupportedCompatibilityVersion name to avoid breaking callers. Added coverage for that preserved recipe export; 28 focused tests and both affected Lambda builds pass.

lambda/environments/status.js
JWThewes: Fixed in d14faf8. StopRuntimeSession is now best-effort cleanup and cannot replace the primary validation error. Added a regression where deterministic validation fails and cleanup throws retryable ResourceNotFoundException; the revision is marked FAILED with runtime_validation_failed instead of returning pending forever. All 15 status tests and the environment bundle pass.

lambda/environments/status.js
JWThewes: Fixed in 0697b88. Runtime verification now explicitly requires deterministic.nonRoot === true, so an older or inconsistent runtime cannot report ok: true with nonRoot: false and still reach READY. Added a regression for that exact response; all 16 status tests and the environment bundle pass.

lambda/intents/index.js
JWThewes: Fixed in 92b9228. Intent creation now maps the four known environment snapshot failures (not published, incomplete, unverified, and compatibility unsupported) to HTTP 409 with the stable error code and domain message; unrelated exceptions still use the generic 500 path. Added table-driven coverage for all four cases. Formatting/lint, resolver tests, and the intents bundle pass; the Docker-backed intent cases will run in CI.

lambda/environments/tools-index.js
JWThewes: Fixed in 52b1cd6. The initialization promise is now retained after success and cleared only in catch, matching the environment handler. Added tests proving sequential invocations seed/list once and a failed initialization is retried. All 8 tool-handler tests and the environment bundle pass.

lambda/projects/index.js
JWThewes: Fixed in 3f1ebfd. Project assignment and intent creation now use the same shared resolvePublishedEnvironment path, which validates publication status, complete runtime/image identity, passed verification, and the current/previous compatibility window before returning the records and immutable snapshot. The projects Lambda now receives RUNTIME_COMPATIBILITY_VERSION from Terraform, and a regression verifies an incompatible environment is rejected at assignment with the same 409 code. Shared tests, both Lambda builds, formatting/lint/secret scan, and the affected Terraform module validation pass; the Docker-backed project case will run in CI.

lambda/discussions/services.js
JWThewes: Addressed in 23217a1. The discussion assist execution-snapshot read now uses DynamoDB eventual consistency; the immutable META snapshot does not require ConsistentRead. Added an assertion on the actual GetCommand input. Formatting/lint and the discussions bundle pass; the Docker-backed integration test will run in CI.

lambda/environments/catalog-recipe.js
JWThewes: Addressed in 60ccee6. I renamed the two implementations by purpose: `recipe.js` → `fixed-tool-recipe.js` and `recipe-v2.js` → `catalog-recipe.js`. I also updated exported symbols, frontend types/copy, tests, and related error labels so the distinction is explicit throughout. The persisted schema versions remain 1 and 2 for compatibility. Verified with all 89 environment tests, the environment bundle build, frontend type-checking, formatting, and linting.

lambda/discussions/services.js
JWThewes: Addressed in 1d820a7. The execution-snapshot lookup is now best-effort: a healthy read still routes the assist to the intent’s managed runtime, while a missing row or failed Get (including throttling/IAM errors) logs a warning and falls back to the configured core runtime instead of failing the interactive request. Added focused regression tests for managed routing, a missing snapshot, and a failed lookup; the discussions Lambda bundle, formatting, and lint checks also pass.

JWThewes:
@jeromevdl Addressed the `agents/index.js` runtime-routing concern in 170a2a6. Project settings now send `projectId` for capability discovery, and the agents Lambda authorizes project access, resolves the project’s published environment, and invokes both `capabilities` and project MCP verification with that runtime ARN + revision qualifier. Global Admin discovery/verification intentionally remains on the protected core runtime. The commit also adds the registry/IAM wiring and focused regression tests.

JWThewes:
@jeromevdl I worked through the duplicated-subsystems change request:

- `60ccee6` replaces the misleading recipe v1/v2 naming with `fixed-tool` (compatibility-only) and `catalog` (current authoring/build path). Keeping the fixed-tool reader avoids breaking persisted environments; new and rebuilt environments use catalog recipes.
- `3f1ebfd` makes `resolvePublishedEnvironment` the shared publication/readiness implementation, with `resolveEnvironmentSnapshot` as its projection. In `656d145`, the remaining `publishedBase` helper was renamed `requirePublishedBaseImage` because it is specifically a build dependency/image check, not another runtime-readiness resolver.
- `656d145` consolidates the shared environment/tool lifecycle mechanics: request parsing/auth/error handling, retryable initialization, DynamoDB pagination, S3 build-context upload, ECR retry/finding normalization, and CodeBuild/ECR/poll event dispatch. That commit is net -67 lines and the complete managed-environments suite remains green (89 tests plus all four Lambda bundles).

I kept the two domain state machines explicit because their publication semantics differ materially: environments supersede revisions and provision/verify AgentCore runtimes, while tools maintain version aliases/recommendations/dependency graphs and finish at image readiness. Merging those layers would replace visible domain rules with callback/configuration branching rather than reducing complexity.

lambda/environments/catalog-recipe.js
```diff
@@ -0,0 +1,546 @@
...
+      'COPY verification-fixtures/ /opt/managed/verification-fixtures/',
+      'RUN chmod -R a-w /opt/managed/verification-fixtures',
+    );
+  }
+  for (const [name, value] of Object.entries(directVariables(recipe)).toSorted(([left], [right]) =>
+    left.localeCompare(right),
+  )) {
+    lines.push(`ENV ${name}=${JSON.stringify(value)}`);
+  }
+  for (const command of recipe.buildCommands) lines.push(`RUN ${command}`);
```
eipasteur, line 375: Build commands run as `USER root` (334) before `sha256sum -c` at 371, and the manifest they are checked against is generated in the same build at 337, so a root `RUN` can regenerate it and `FORBIDDEN_COMMAND` (21) is a string denylist that `cd /opt && ... > managed/protected-runtime.sha256` walks around. I think that is fine if admin build commands are trusted by design, which seems to be the model here. Could we say so in a comment at 369 so the checksum step is not read as an integrity guarantee?

lambda/environments/tool-catalog.js
```diff
@@ -0,0 +1,1291 @@
...
+    parts[0] === 10 ||
+    parts[0] === 127 ||
+    parts[0] === 0 ||
+    (parts[0] === 169 && parts[1] === 254) ||
+    (parts[0] === 172 && parts[1] >= 16 && parts[1] <= 31) ||
+    (parts[0] === 192 && parts[1] === 168)
+  );
+};
+
+const assertPublic = async (url, allowQuery = false) => {
```
eipasteur, line 781: `assertPublic` checks `dns.lookup` results but `fetch(url)` at 806 resolves the name again unpinned, and `privateAddress` (768) misses `::ffff:` mapped addresses and CGNAT. The `DOCKER-USER` rules at 1040 are the actual backstop, so no exposure today. Might be worth a comment saying the JS check is advisory and iptables is load-bearing, or pinning via an undici `Agent` with `connect.lookup` if you'd rather close it.

eipasteur (commented):
Verdict: comment, non-blocking. Read `tool-catalog.js`, `catalog-recipe.js`, `agents/index.js` and `managed-environments/main.tf` at 14883fe; jeromevdl's points look addressed in `170a2a6`, `60ccee6`, `3f1ebfd`, `656d145`, pending his re-review. IAM is well scoped; the only wildcards are `ecr:GetAuthorizationToken` (unavoidable) and the AgentCore statement at 383 to 391, where `Get*` and `TagResource` could probably use `local.managed_runtime_arn`. Two inline notes below on guards that read stronger than they are; both are fine to defer.

lambda/environments/catalog-recipe.js
JWThewes: Good catch. The denylist and in-image checksum are only guardrails because administrator-authored build commands execute as root. The integrity boundary is the post-build `verification.sh`, which independently compares `/opt/agentcore` and `/opt/shared` with the pinned base image. I clarified that in both recipe generators in 8e2f0a0; the focused recipe tests pass.

lambda/environments/tool-catalog.js
JWThewes: Valid observation. The JavaScript DNS lookup is advisory because `fetch()` resolves again; the security boundary is the source-fetch container with IPv6 disabled plus the `DOCKER-USER` connect-time rejects. In 52d0a4b I documented that on both sides of the boundary and aligned the advisory classifier with the firewall for IPv4-mapped addresses and `100.64.0.0/10`. The focused test and commit checks pass.

lambda/environments/catalog-recipe.js
```diff
@@ -0,0 +1,548 @@
...
+
+const PACKAGE_PATTERN = /^[a-z0-9][a-z0-9+.-]*$/;
+const VERSION_PATTERN = /^[0-9][0-9A-Za-z.+:~_-]*$/;
+const VARIABLE_PATTERN = /^[A-Z][A-Z0-9_]{0,63}$/;
+const DIGEST_PATTERN = /^sha256:[a-f0-9]{64}$/i;
+const SECRET_NAME = /(secret|password|passwd|token|credential|private[_-]?key|api[_-]?key)/i;
+const SECRET_VALUE =
+  /(-----BEGIN [A-Z ]+PRIVATE KEY-----|AKIA[0-9A-Z]{16}|ASIA[0-9A-Z]{16}|gh[pousr]_[A-Za-z0-9_]{20,})/;
+const SECRET_COMMAND =
+  /(?:^|\s)(?:export\s+)?[A-Za-z0-9_]*(?:secret|password|passwd|token|credential|private[_-]?key|api[_-]?key)[A-Za-z0-9_]*\s*=|(?:^|\s)--?[A-Za-z0-9_-]*(?:secret|password|passwd|token|credential|private-key|api-key)[A-Za-z0-9_-]*(?:=|\s)/i;
```
jeromevdl: 🟡 **Security-critical denylists and Dockerfile generators are duplicated verbatim across the v1 and v2 recipe engines**

fixed-tool-recipe.js (lines 60-65) and catalog-recipe.js (lines 15-24) each define their own copies of `SECRET_COMMAND`, `FORBIDDEN_COMMAND`, `PROTECTED_VARIABLE_NAME`, `SECRET_NAME` and `SECRET_VALUE`, plus near-identical `generateDockerfile` / `generateCatalogEnvironmentDockerfile` and `generateVerificationScript` / `generateCatalogEnvironmentVerificationScript`. tool-catalog.js carries a third copy of the same secret/variable regexes. 

**Why it matters:** A future hardening of one denylist (e.g. adding a newly-discovered dangerous build-command token or protected variable) applied to only one of the two/three copies -> the other engine still emits the unsafe Dockerfile RUN or accepts the protected variable -> an environment built through that path bypasses the intended guardrail, relying solely on the post-build verification.sh diff to catch it. Duplication of a security control is the exact place drift causes a silent regression.

**Suggestion:** Extract the shared secret/variable/command denylists and the common Dockerfile/verification-script primitives into one module imported by both recipe engines (and by tool-catalog.js), so there is a single source of truth for the security regexes. If schema-v1 is compatibility-only, gate it behind that framing so no new authoring path can diverge.

lambda/environments/catalog-recipe.js
```diff
@@ -0,0 +1,548 @@
...
+  );
+  lines.push(
+    `FROM ${recipe.base.imageUri}@sha256:${digestValue(recipe.base.imageDigest)}`,
+    'USER root',
+    'COPY manifest.json checksums.json sbom.spdx.json verification.sh /opt/managed/',
+    'RUN chmod 0555 /opt/managed/verification.sh',
+    'RUN find /opt/agentcore /opt/shared -type f -print0 | sort -z | xargs -0 sha256sum > /opt/managed/protected-runtime.sha256',
+  );
+  const packages = directPackages(recipe);
+  if (packages.length) {
```
jeromevdl, line 343: 🟡 **Admin build commands execute as root before the in-image protected-runtime checksum runs, so the checksum cannot stop tampering**

In both `generateCatalogEnvironmentDockerfile` and `generateDockerfile` the manifest of protected files is written earlier in the same build (RUN find /opt/agentcore /opt/shared ... > protected-runtime.sha256) and the administrator-authored `RUN ${command}` lines execute as USER root; only afterwards does `RUN sha256sum -c` run. Threat chain: authenticated platform-admin -> supplies a build command that regenerates protected-runtime.sha256 after modifying /opt/agentcore -> the in-image `sha256sum -c` passes against the tampered manifest -> the only remaining detector is the out-of-image verification.sh, which recomputes the diff against the pinned base image. 

**Why it matters:** If verification.sh or its base-diff step is ever weakened or skipped, a privileged admin build command could ship a runtime with altered protected files undetected; today the exposure is bounded because verification.sh does an independent `docker cp` + `diff -qr` of /opt/agentcore and /opt/shared against the pinned base outside the build.

**Suggestion:** Document the trust model in one place and keep the base-diff step in verification.sh non-optional and covered by a regression test; consider recording the protected-runtime manifest from the pinned base image rather than regenerating it inside the same build so the in-image checksum is also meaningful.

lambda/environments/test/tool-catalog.test.js
```diff
@@ -0,0 +1,302 @@
...
+
+  it('validates dependent tools against exact published dependency artifacts', () => {
+    const java = {
+      toolId: 'java',
+      name: 'Java JDK',
+      category: 'language-sdk',
+      publisher: 'Eclipse Temurin',
+      versionId: 'tv-java-21',
+      version: '21.0.8',
+      imageUri: '111111111111.dkr.ecr.us-east-1.amazonaws.com/tools',
```
jeromevdl: 🟡 **CodeQL: incomplete hostname regex from a test string with an unescaped dot before amazonaws.com/tools**

CodeQL alert 23 (Incomplete regular expression for hostnames) flags a test string whose unescaped '.' before 'amazonaws.com/tools' is consumed as a regex and could match more hosts than intended. It is test fixture data in tool-catalog.test.js, not a production host-validation path, so it has no runtime security impact.

```
imageUri: '111111111111.dkr.ecr.us-east-1.amazonaws.com/tools',
```

**Why it matters:** No production exposure; the risk is purely that a test could pass against an unintended host string, weakening the assertion's precision.

**Suggestion:** Escape the dots (or use a literal string comparison / stringContaining instead of a regex matcher) so the assertion matches only the intended registry host, and clear the CodeQL alert.

_[code-review-sage]_

terraform/modules/compute/agentcore/main.tf
```diff
@@ -141,18 +141,25 @@ resource "aws_ecr_repository" "agentcore" {
   tags = var.tags
```
jeromevdl, line 141: 🟡 **ECR "keep last 3" expiry removed with no eviction replacement; core + managed-environment images now grow unbounded**

The change deletes aws_ecr_lifecycle_policy.agentcore (which expired all but the last 3 images) and replaces it with only a comment, and it adds a new IMMUTABLE managed_environments ECR repository that also has no lifecycle policy. Neither repository now has any image-eviction mechanism. The stated rationale is sound (environment revisions and active intents pin core image digests by sha256, so a blind count-based expiry would break the immutability/pinning invariant), and the comment explicitly defers eviction to a future 'registry-aware cleanup' that is not part of this PR.

**Why it matters:** cause -> the count-based ECR expiry guard is removed and no digest-aware cleanup is shipped -> mechanism -> every rebuilt protected core image and every managed-environment/tool revision image is retained in ECR indefinitely -> consequence -> ECR storage cost grows monotonically for the platform's lifetime with no automated reclamation. Latent, not a correctness break: growth rate is bounded by admin-driven publish/rebuild frequency, and force_delete keeps non-prod teardown cheap; but on a long-lived prod deployment this is a real, unbounded cost/operability drift.

**Suggestion:** Track the promised registry-aware cleanup as a concrete follow-up (issue/ticket) rather than a code comment, and consider an interim digest-aware safeguard: a lifecycle policy that expires only UNTAGGED images, or count/age expiry scoped to digests provably unreferenced by any active intent snapshot or published revision, so retention cannot be broken while still bounding growth.

frontend/src/components/admin/tabs/EnvironmentRegistry.tsx
```diff
@@ -0,0 +1,1181 @@
...
+                      {environment.environmentId}
+                    </p>
+                    {(environment.toolUpdates?.length ?? 0) > 0 && (
+                      <p className="mt-2 text-xs text-amber-700 dark:text-amber-300">
+                        Recommended tool updates are available. Save a new revision to select them.
+                      </p>
+                    )}
+                    {fixedToolEnvironment && (
+                      <p className="mt-2 text-xs text-amber-700 dark:text-amber-300">
+                        This fixed-tool environment is read-only. Use Reset to remove it, then
```
jeromevdl: **🟡 UI tells admins to "Use Reset" — a control this PR deliberately does not ship**

No Reset button/action/route/handler exists anywhere in the diff (grep for Reset in the component finds only this copy; no reset handler in lambda/environments/index.js). The PR description explicitly states "no reset API, Lambda, IAM role, route, or Platform Settings view is shipped"

jeromevdl (commented):
Most of the findings were correctly fixed. Remains a few minor issues (no blocking, some of them can be documented in new tickets)

frontend/src/components/admin/tabs/EnvironmentRegistry.tsx
JWThewes: Good catch — valid. `Reset` was stale copy: grep confirms it was the only occurrence in the repo, and there is no reset route, handler, or service method behind it.

Fixed in 23464a2. The notice now names **Retire**, which is the control the UI actually renders for a non-standard environment (and `fixedToolEnvironment` is only ever true for non-standard environments, so the button is always present when the notice shows). Added a regression that renders a read-only fixed-tool environment and asserts the notice appears alongside an enabled `Retire` button and that no `Reset` text is offered. The 9 `EnvironmentsTab` tests, formatting, lint, and frontend type-checking pass.

lambda/environments/catalog-recipe.js
JWThewes: Valid, and worth fixing — the three `SECRET_NAME`/`SECRET_VALUE`/`PROTECTED_VARIABLE_NAME` copies were byte-identical, and `SECRET_COMMAND`/`FORBIDDEN_COMMAND` were identical across both engines.

Fixed in 31ea0dd. `lambda/environments/build-guardrails.js` is now the single source of truth, imported by `fixed-tool-recipe.js`, `catalog-recipe.js`, and `tool-catalog.js`. Beyond the five denylists it also owns the primitives that were near-duplicates:

- `PROTECTED_RUNTIME_PATHS` / `PROTECTED_RUNTIME_MANIFEST`
- `protectedManifestLine()` and `protectedRuntimeEpilogueLines()` (the checksum + `USER node` + image-contract tail)
- `verificationPrologue()` (image contract, startup, non-root workspace, in-image checksum, SBOM shape, base-tree diff) and `verificationEpilogue()`

Net −97 lines across the three files.

**Behaviour preservation.** I diffed the generated artifacts against the previous implementation rather than trusting the tests: the fixed-tool Dockerfile and `verification.sh` and the catalog Dockerfile are byte-identical. The catalog `verification.sh` has one intentional change — it now uses the `arch=...; test "$arch" = "arm64"` form the fixed-tool engine already used instead of inlining the same three `docker image inspect` checks. Semantically identical; after normalising that one block the scripts are byte-identical.

**Drift guard.** `test/build-guardrails.test.js` asserts at the source level that no consumer re-declares any of the five denylists and that none inlines `/opt/agentcore /opt/shared`, so a future private copy fails the suite rather than silently diverging. Plus behavioural coverage of each denylist and of the non-optional base-diff step.

On your last point — schema v1 being compatibility-only is already enforced upstream: `isCatalogRecipe` gates Build/Retry/edit in the UI and new and rebuilt environments always use catalog recipes, so the fixed-tool engine is not reachable as a new authoring path. All 100 environment tests, both Lambda bundles, formatting, and lint pass.

lambda/environments/catalog-recipe.js
JWThewes: Valid — fixed in 5d6fbc2.

You were right that the ordering made the in-image check circular: the manifest was generated in the environment stage *before* the administrator `RUN` lines, so a root command could regenerate it and `sha256sum -c` would then verify the tampered tree against a forged list.

The manifest is now computed in a dedicated stage straight from the pinned base digest, and only enters the environment stage *after* every administrator command:

```dockerfile
FROM <base>@sha256:... AS protected_runtime_manifest
USER root
RUN find /opt/agentcore /opt/shared -type f -print0 | sort -z | xargs -0 sha256sum > /tmp/protected-runtime.sha256

FROM <base>@sha256:...
...
RUN <administrator build command>
COPY --from=protected_runtime_manifest /tmp/protected-runtime.sha256 /opt/managed/protected-runtime.sha256
RUN sha256sum -c /opt/managed/protected-runtime.sha256
USER node
```

Nothing the environment stage runs can write into the manifest stage, so the list is no longer forgeable from inside the build.

Two things worth calling out:

- The final stage is still the bare `FROM $base_ref` line, so the CodeBuild buildspec's `grep -Fx "FROM $base_ref" Dockerfile` integrity check and the untargeted `docker build .` both still behave as before.
- `FORBIDDEN_COMMAND` is a guardrail against mistakes, not a sandbox — admin commands do run as root. I wrote that down explicitly as `PROTECTED_RUNTIME_TRUST_MODEL` in `build-guardrails.js` alongside what actually carries integrity: this out-of-stage manifest, plus the out-of-image `docker cp` + `diff -qr` of both protected trees against the pinned base in `verification.sh`, which must stay unconditional.

Regressions added in `test/fixed-tool-recipe.test.js`, `test/catalog-recipe.test.js` (both engines assert manifest-stage-before-environment-stage and COPY-after-last-build-command), and `test/build-guardrails.test.js`.

terraform/modules/compute/agentcore/main.tf
JWThewes: Valid — addressed in 4a0d9b5, with the registry-aware cleanup tracked as #427.

Took both halves of your suggestion.

**Tracked, not just commented:** #427 covers the digest-aware cleanup — enumerate `base.imageDigest` + every `tools[].imageDigest` on published and in-flight revisions, plus the core digest on each active intent snapshot and the deployed runtime, then delete only digests absent from that set, with a minimum-age grace period so an in-flight build can't lose its base mid-run. Dry-run by default, per-environment opt-in.

**Interim safeguard, scoped to where it's provably safe:** UNTAGGED-only expiry (7 days) on the two IMMUTABLE repositories:

- `managed-environments` — `aws_ecr_lifecycle_policy.managed_environments`
- `managed-tools` — `aws_ecr_lifecycle_policy.managed_tools` (this one wasn't in your comment's scope but has the same shape, so I covered it too)

Safe there because the two conditions hold together: `image_tag_mutability = "IMMUTABLE"` means a tag can never move off a digest a revision pins, and every push is tagged — `image_ref=$IMAGE_REPOSITORY_URI:$IMAGE_TAG` with `IMAGE_TAG = revision.revisionId`, and `tool_ref="${TOOL_REPOSITORY_URI}:${TOOL_IMAGE_TAG}"`. So an untagged image in either repo is only garbage from an aborted or partial push, never something a revision or intent references.

**Deliberately excluded the core repo**, and I left that reasoning in the file rather than just here. `agentcore` is `MUTABLE` with `agentcore_image_tag = substr(local.agentcore_files_sha, 0, 16)`. A non-reproducible rebuild pushes the same content-hash tag with a different digest, which leaves the *previous* digest untagged — while published revisions and active intents still pin it by `sha256:`. An UNTAGGED rule there would delete precisely the digest that must survive. That's the one case where the cheap safeguard is actively wrong, which is why it needs #427's live-reference resolution rather than any tag-status heuristic.

Net effect: growth is bounded for failed-push garbage everywhere it can be, no rule can evict a pinned digest, and the remaining unbounded set (tagged images, all of which are pinned by something today) has a concrete owner in #427 instead of a code comment.

JWThewes:
@jeromevdl I worked through all of your findings

jeromevdl (approved):
LGTM

Issue #353, opened 2026-07-24
[Feature]: Update tracker when intent is finished


jeromevdl:
### Description

When the workflow ends (whatever the last phase/stage is), we should update the associated tracker to mark the issue/task done, and get some details about the PR / branch / workflow (main decisions) / link to collaborative aidlc corresponding intent.

### Use case

When an intent (started from a tracker issue/task) is done, I want to get the issue/task updated (mark done), so that I know it is complete.

### Area

Other

### Additional context

We could eventually think about other integration points (e.g: add the requirements as a comment in the original issue, create the user stories or units in the tracker as subtasks to the original one and mark them done too, ...)

eipasteur:
Resolved by #399. Tracker synchronization now posts delivery context, monitors final PR/MR merges, closes the linked issue when supported, and publishes timeline events. Closing as completed.

PR #399, opened 2026-08-11
feat: update trackers after intent delivery

jeromevdl:
## Summary

- publish durable tracker-delivery synchronization after final PR/MR creation
- add provider-specific tracker comments, branch and delivery links, merge polling, and GitHub/GitLab issue closure
- add native closing references to generated PR/MR bodies and Jira task references
- surface tracker synchronization events live in the intent timeline
- provision the scheduled reconciler and required DynamoDB/WebSocket permissions

## Issue
Solves #353 

## Verification

- manually tested the end-to-end workflow on a deployed development environment, including tracker comment creation, PR linkage, merge detection, issue closure, and timeline updates
- repository commit hook: 614 tests passed
- focused review suite: 197 backend tests passed
- IntentActivityPanel: 4 tests passed
- frontend typecheck passed
- trackers and v2-orchestrator builds passed
- lint and formatting checks passed
- Terraform validation passed with existing deprecation warnings

image

image

### DIFF
--- a/lambda/shared/git-providers/github.js
+++ b/lambda/shared/git-providers/github.js
@@ -365,6 +365,20 @@ const addIssueComment = async (ctx, repoId, issueNumber, body) => {
   return mapIssueDiscussionComment(data);
 };
 
+const closeIssue = async (ctx, repoId, issueNumber) => {
+  const { owner, repo } = splitOwnerRepo(repoId);
+  const res = await ghFetch(ctx, `${API_BASE}/repos/${owner}/${repo}/issues/${issueNumber}`, {
+    method: 'PATCH',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({ state: 'closed' }),
+  });
+  const data = await res.json().catch(() => ({}));
+  if (!res.ok || data?.pull_request) {
+    throw new ProviderError(res.status || 404, data?.message || 'Failed to close issue');
+  }
+  return mapIssue(data);
+};
+
 // ---------------------------------------------------------------------------
 // PR comments
 // ---------------------------------------------------------------------------
@@ -938,6 +952,7 @@ export {
   getIssue,
   listIssueComments,
   addIssueComment,
+  closeIssue,
   listPRComments,
   addPRComment,
   getUnmergedConstructionTaskBranches,
@@ -976,6 +991,7 @@ export default {
   getIssue,
   listIssueComments,
   addIssueComment,
+  closeIssue,
   listPRComments,
   addPRComment,
   getUnmergedConstructionTaskBranches,
--- a/lambda/shared/git-providers/github.js
+++ b/lambda/shared/git-providers/github.js
@@ -326,7 +326,7 @@ const getIssue = async (ctx, repoId, issueNumber) => {
   const data = await res.json().catch(() => ({}));
   if (!res.ok || data?.pull_request) {
     throw new ProviderError(
-      res.status || 404,
+      data?.pull_request ? 404 : res.status || 404,
       data?.message || (data?.pull_request ? 'Not found' : 'Failed to fetch issue'),
     );
   }
@@ -367,14 +367,15 @@ const addIssueComment = async (ctx, repoId, issueNumber, body) => {
 
 const closeIssue = async (ctx, repoId, issueNumber) => {
   const { owner, repo } = splitOwnerRepo(repoId);
+  await getIssue(ctx, repoId, issueNumber);
   const res = await ghFetch(ctx, `${API_BASE}/repos/${owner}/${repo}/issues/${issueNumber}`, {
     method: 'PATCH',
     headers: { 'Content-Type': 'application/json' },
     body: JSON.stringify({ state: 'closed' }),
   });
   const data = await res.json().catch(() => ({}));
-  if (!res.ok || data?.pull_request) {
-    throw new ProviderError(res.status || 404, data?.message || 'Failed to close issue');
+  if (!res.ok) {
+    throw new ProviderError(res.status, data?.message || 'Failed to close issue');
   }
   return mapIssue(data);
 };
--- a/lambda/trackers/delivery-sync.js
+++ b/lambda/trackers/delivery-sync.js
@@ -1,4 +1,6 @@
 import { randomUUID } from 'node:crypto';
+import { mapWithConcurrency } from '../shared/concurrency.js';
+import { buildIntentUrl } from '../shared/intent-url.js';
 
 const CREATED_PHRASE = 'AI-DLC completed the workflow for this issue.';
 const CREATED_TASK_PHRASE = 'AI-DLC completed the workflow for this task.';
@@ -8,12 +10,12 @@ const MAX_TRACKER_WRITE_ATTEMPTS = 10;
 const CLAIM_LEASE_MS = 90_000;
 const SCHEDULED_CONCURRENCY = 5;
 
-const trimBaseUrl = (value) => String(value || '').replace(/\/+$/, '');
-
 const intentUrlFor = (applicationUrl, sync) =>
-  `${trimBaseUrl(applicationUrl)}/space/${encodeURIComponent(
-    sync.projectId,
-  )}/intent/${encodeURIComponent(sync.intentId)}`;
+  buildIntentUrl({
+    applicationUrl,
+    projectId: sync.projectId,
+    intentId: sync.intentId,
+  });
 
 const intentLabel = (sync) => sync.intentTitle || sync.intentId;
 
@@ -30,7 +32,7 @@ const branchUrlFor = (pr, fallbackBranch) => {
   if (!branch || !pr.prUrl || !pr.repoId) return null;
   try {
     const url = new URL(pr.prUrl);
-    const encodedBranch = encodeURIComponent(branch);
+    const encodedBranch = branch.split('/').map(encodeURIComponent).join('/');
     if (pr.provider === 'github') {
       url.pathname = `/${pr.repoId}/tree/${encodedBranch}`;
     } else if (pr.provider === 'gitlab') {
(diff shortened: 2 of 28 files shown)


lambda/trackers/providers/jira-cloud.js
```diff
@@ -14,6 +14,7 @@ export { ProviderError };
 const JIRA_TOKEN_PARAM_PATTERN = /^\/[\w-]+\/[\w-]+\/[\w-]+\/[\w-]+$/;
 const PROVIDER_INSTANCE = 'jira-cloud#cloud';
 const REFRESH_SAFETY_MARGIN_MS = 60_000;
+const JIRA_OAUTH_SCOPE = 'read:jira-work read:jira-user write:jira-work offline_access';
```
JWThewes: **Blocking.** Adding `write:jira-work` here only changes the *authorize URL*. Every user who is already connected holds a refresh token minted under the old read-only scope, and `refreshAccessToken` returns the same scope set it was granted — a refresh never widens a grant.

So for every existing Jira connection, `addIssueComment` gets a 403 from Atlassian. That lands on `isPermanentTrackerWriteError` in `delivery-sync.js:141`, which goes straight to `BLOCKED` with no retry, and the user sees "stopped after a non-retryable authorization failure" with nothing telling them that reconnecting is the fix.

Could we detect the narrow scope and surface a reconnect prompt? There's already a `{ reconnect: true }` convention in this file (`jiraFetch`, ~line 294) that seems built for exactly this.

lambda/trackers/providers/jira-cloud.js
```diff
@@ -150,7 +151,7 @@ export const persistConnection = async ({ ddb, ssm, userId, resource, tokens, sc
         baseUrl: `https://api.atlassian.com/ex/jira/${resource.cloudId}`,
         siteHost: resource.host || '',
         siteName: resource.name || '',
-        scope: scope || 'read:jira-work read:jira-user offline_access',
+        scope: scope || JIRA_OAUTH_SCOPE,
```
JWThewes: Related to the scope comment above, and the part I'd most like changed: dropping the explicit `scope:` arguments at both `persistConnection` call sites (`trackers/index.js` ~597 and ~674) means this now defaults to the new constant.

Net effect is that the stored row claims `write:jira-work` for tokens that don't have it. That was the one field we could have keyed a "reconnect required" check off, and now it actively lies. Can we persist the scope Atlassian actually returned in the token response instead of defaulting?

lambda/trackers/providers/jira-cloud.js
```diff
@@ -359,6 +360,89 @@ const mapCommentToTrackerComment = (c) => ({
...
+  let lastListItem = null;
+
+  const flushBulletList = () => {
+    if (bulletList) content.push(bulletList);
+    bulletList = null;
+    lastListItem = null;
+  };
+
+  for (const line of String(body).split('\n')) {
+    if (line.startsWith('  - ')) {
```
JWThewes, line 434: **Blocking.** Nesting keys off the literal `'  - '` prefix, but `lastListItem` isn't reset when a new top-level group starts, so the two groups in a delivery comment collide.

`branchLines` emits `'- Branches:'` then indented repo rows; `pullRequestLines` emits *only* indented rows under a separately-pushed `'- Pull requests:'`. On a multi-repo intent the pull-request rows nest under the last-seen list item — the **Branches** entry — instead of **Pull requests**. Single-repo intents hit the `rows.length === 1` early return in `branchLines` and never expose it, which I suspect is why the manual run looked right.

The deeper thing: we're generating Markdown and then parsing it back out, and this bug is a direct consequence. Would it be worth building the comment as structured content once and rendering per provider (Markdown for GitHub/GitLab, ADF here)? That deletes the parser rather than patching it. A multi-repo Jira case in `jira-cloud.test.js` would pin whichever way you go.

lambda/shared/git-providers/github.js
```diff
@@ -365,6 +365,20 @@ const addIssueComment = async (ctx, repoId, issueNumber, body) => {
...
 
+const closeIssue = async (ctx, repoId, issueNumber) => {
+  const { owner, repo } = splitOwnerRepo(repoId);
+  const res = await ghFetch(ctx, `${API_BASE}/repos/${owner}/${repo}/issues/${issueNumber}`, {
+    method: 'PATCH',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({ state: 'closed' }),
+  });
+  const data = await res.json().catch(() => ({}));
+  if (!res.ok || data?.pull_request) {
```
JWThewes: The `data?.pull_request` guard matches the existing `getIssue` pattern just above, so the intent is clearly right — but on a write it behaves differently than on a read. By the time we check, the PATCH has already succeeded and the issue is closed, so we mutate and *then* report failure; `delivery-sync` records a write error for something that actually completed.

Secondary: on that path `res.status` is 200, so this throws `ProviderError(200, …)`. `handleProviderError` passes `err.status` through as the HTTP response code, which would emit a 200 for an error.

Cheapest fix is probably to check `pull_request` before the PATCH (or reuse `getIssue`).

lambda/trackers/delivery-sync.js
```diff
@@ -0,0 +1,414 @@
...
+  const resourceId = String(sync.source?.resourceId || '').replace(/^#/, '');
+  return resourceId ? `Issue #${resourceId}` : 'Issue';
+};
+
+const branchUrlFor = (pr, fallbackBranch) => {
+  const branch = pr.branch || fallbackBranch;
+  if (!branch || !pr.prUrl || !pr.repoId) return null;
+  try {
+    const url = new URL(pr.prUrl);
+    const encodedBranch = encodeURIComponent(branch);
```
JWThewes: **Optional / cosmetic.** `encodeURIComponent` on the whole branch encodes the `/`, and every AI-DLC branch is `aidlc/<slug>` (`intents/index.js:416`), so links render as `/tree/aidlc%2Ffoo`.

I tested this live against GitHub, GitLab and Bitbucket — all three resolve `%2F` correctly (with a 404 control to make sure the test meant something), so nothing is broken. Only note is that `delivery-sync.test.js:94` pins the encoded form as the expected output, so we're locking in a slightly odd-looking URL. Encoding path segments individually would read better. Fine to leave.

lambda/trackers/delivery-sync.js
```diff
@@ -0,0 +1,414 @@
...
+          sync,
+          binding,
+          provider,
+          ctx,
+          phrase: mergedPhraseFor(sync),
+          body: mergedComment(sync),
+          matchExisting: hasMergedDeliveryComment,
+        });
+        if (typeof provider.closeIssue === 'function') {
+          await provider.closeIssue(ctx, binding.externalProjectKey, sync.source.resourceId);
```
JWThewes, line 330: **Optional.** The comment and the close aren't separately checkpointed, so if `closeIssue` fails after `addIssueComment` succeeded, the retry re-enters `PR_MERGED` and calls `closeIssue` again.

Both providers are idempotent on close today, and the exact-body match in `hasMergedDeliveryComment` does stop a duplicate comment (the body is derived from the row, so it's stable). So this is latent fragility rather than a live bug — a short comment noting that we rely on close being idempotent would be enough for the next reader.

lambda/trackers/delivery-sync.js
```diff
@@ -0,0 +1,414 @@
...
+};
+
+const sanitizedError = (error) => ({
+  code: error?.code || error?.name || 'TRACKER_SYNC_FAILED',
+  message: String(error?.message || 'Tracker synchronization failed').slice(0, 500),
+});
+
+const isPermanentTrackerWriteError = (error) => [401, 403].includes(Number(error?.status));
+
+const mapWithConcurrency = async (items, limit, worker) => {
```
JWThewes: **Optional.** This is byte-identical to `mapWithConcurrency` in `lambda/intents/index.js:237`, and `intentUrlFor` above duplicates the URL construction in `v2-orchestrator/pr-attribution.js:2`.

`lambda/shared/` is the natural home for both. Worth doing at some point — `shared/repo-validation.js` has a header comment about consolidating exactly this kind of drifted copy after a hardening fix got applied to one lambda and not the other.

lambda/v2-orchestrator/index.js
```diff
@@ -1709,16 +1740,57 @@ const openIntentPrs = async ({
     projectId: meta.projectId,
     intentId: meta.intentId ?? executionId,
   });
-  const body = [
-    `Automated ${gitProvider === 'gitlab' ? 'MR' : 'PR'} created by ${aidlcAttribution} (strategy: ${strategy})`,
-    ...unitLines,
-  ].join('\n');
+  const trackerRepository = (() => {
```
JWThewes, line 1743: **Optional / readability.** Provider-specific URL path parsing inline in `openIntentPrs`, which is already long. It also feels like guesswork: `meta.source` carries a `resourceUrl` but no repo field, so we're reverse-engineering the repo out of the URL and silently returning `null` on any shape we don't recognise.

Two options if you want to pick this up later — extract to a small tested helper, or capture the repo at intent-creation time in `normalizeSource` (`intents/index.js:1103`) so there's nothing to parse. Second one seems strictly better but is a bigger change than this PR needs.

JWThewes (changes requested):
Really like the shape of this — the reconciler deliberately mirrors `intents_durable_watchdog` (same `rate(1 minute)`, same `{action}` envelope, same lease-claim + self-removing sparse GSI), which made it quick to reason about. I checked the Terraform wiring end to end and all four new env vars plus both new IAM grants resolve to declared variables and real module outputs. I also ran the affected suites locally (`lambda/trackers`, `lambda/shared`, `lambda/v2-orchestrator`, `lambda/source-control`): 882 tests / 49 files green, matching your description.

Two things I think need to change before this goes in, and three that are optional:

**Blocking**
1. The Jira `write:jira-work` scope upgrade doesn't reach already-connected users, and they end up permanently `BLOCKED` with a misleading message. Details inline.
2. `markdownToAdf` mis-nests the pull-request bullets on multi-repo intents. Single-repo runs take a different branch, which is probably why manual testing looked clean. Details inline.

**Worth fixing, small**
3. The `pull_request` guard in the new `closeIssue` fires *after* the PATCH has already landed, and throws with a 200 status. Inline.

**Optional — happy to merge without these, just flagging**
4. Branch links encode the `/` in `aidlc/*` branch names. I checked live against GitHub, GitLab and Bitbucket and `%2F` resolves fine on all three, so this is purely cosmetic.
5. `PR_MERGED` retries re-call `closeIssue`. Idempotent on both providers today, so it's latent fragility rather than a bug.
6. `mapWithConcurrency` and the intent-URL builder are both copy-pasted from elsewhere in the repo — good candidates for `lambda/shared/`.

Marking request-changes on 1 and 2 only — the rest is yours to take or leave.

jeromevdl:
Addressed the review feedback in `3a98b11`:

  1. Jira connections now persist the scopes actually granted by Atlassian. Legacy read-only connections are detected before writing and surface an explicit “Reconnect Jira” message. Setup guidance now includes `write:jira-work`.
  2. I tested the exact multi-repo Markdown shape and confirmed the existing ADF parser keeps Branches and Pull requests separate. Added a regression test to pin this behavior.
  3. GitHub now verifies the resource is an issue before sending the close PATCH. Pull requests return 404 without being mutated.
  4. Branch URLs now preserve `/` separators while encoding each path segment, with coverage for GitHub, GitLab, and Bitbucket.
  5. Left the close retry behavior unchanged because both providers currently implement idempotent close operations and duplicate comments are already prevented.
  6. Extracted `mapWithConcurrency` and the intent URL builder into focused shared modules with tests.

✅ Local verification: 54 focused tests pass; tracker, intents, and v2-orchestrator Lambda bundles build; lint and formatting pass.

JWThewes (approved):

PR #402, opened 2026-08-12
Fix paused-stage workspace recovery

### LINKED ISSUE #401: [Bug]: Resuming a paused stage can fail workspace restore and restart the full workflow
### Description

When a repository-backed stage pauses for a human question and its AgentCore session is released, answering the question can fail the resumed stage with `workspace_restore_failed`.

The execution is still marked `WAITING` when AgentCore tries to restore the workspace. The credential broker only permits repository credentials for `CREATED` or `RUNNING` executions, so it rejects the clone with `EXECUTION_NOT_ACTIVE`.

After this failure, the top-level Restart action starts the complete workflow again. A normal stage failure should instead retry from the failed stage and preserve completed upstream work.

Expected behavior:
- Mark the execution active before restoring the workspace.
- Resume the paused stage successfully.
- If a stage does fail, retry from that stage rather than restarting the full workflow.

### Steps to reproduce

1. Start an intent connected to a source-control repository.
2. Let a stage ask a human question and enter `WAITING`.
3. Allow the AgentCore session to be released so the workspace must be restored on resume.
4. Answer the question.
5. Observe that workspace restoration requests repository credentials while the execution still has status `WAITING`.
6. Observe `workspace_restore_failed`.
7. Use the top-level Restart action.
8. Observe that the workflow starts again from its first stage instead of the failed stage.

### Logs or screenshots

AgentCore runtime:

```text
[workspace] clone failed { provider: 'gitlab', reason: 'EXECUTION_NOT_ACTIVE' }
```

Credential broker:

```text
[credential-broker] request denied {
  code: 'EXECUTION_NOT_ACTIVE',
  provider: 'gitlab'
}
```

Execution events:

```text
Restoring workspace...
workspace_restore_failed: could not re-clone: <repository>
stage_failed: units-generation: workspace_restore_failed
```

### Environment

- AWS Region: `eu-central-1`
- Runtime: Amazon Bedrock AgentCore
- Repository provider observed: GitLab
- Workflow stage observed: `units-generation`

jeromevdl:
## Summary

A paused stage could fail immediately after a human answered its question.

The run was still marked `WAITING` when AgentCore tried to restore its workspace. Because repository credentials are only available to active runs, the clone was rejected with `EXECUTION_NOT_ACTIVE` and the stage reported `workspace_restore_failed`.

The recovery button then used the full workflow start path, so every stage ran again instead of retrying only the failed stage.

## Changes

- Mark the execution `RUNNING` before resuming AgentCore, allowing workspace restoration to obtain repository credentials.
- Retry executions with a failed stage through the existing rewind path, preserving completed upstream work.
- Keep the full start path for failures that have no stage to retry, such as workspace initialization failures and stalled `CREATED` runs.

## Verification

- 151 v2 orchestrator tests passed.
- 39 focused IntentView tests passed.
- Frontend typecheck, formatting, lint, and production build passed.
- Orchestrator lint, formatting, and production build passed.

Fixes #401

### DIFF
--- a/frontend/src/pages/IntentView.tsx
+++ b/frontend/src/pages/IntentView.tsx
@@ -41,6 +41,7 @@ import {
   Loader2,
   MoreHorizontal,
   Play,
+  RotateCcw,
   Trash2,
   TriangleAlert,
   Wrench,
@@ -66,7 +67,9 @@ export default function IntentView() {
     answerGate,
     cancelIntent,
     deleteIntent,
+    rewindIntent,
     focusOutput,
+    stageRows,
     stageNameOf,
   } = useIntent();
   const navigate = useNavigate();
@@ -87,18 +90,26 @@ export default function IntentView() {
   const [repairing, setRepairing] = useState(false);
   const [actionError, setActionError] = useState<string | null>(null);
 
-  // Restart (FAILED / stranded CREATED) via the /start endpoint, which
-  // re-enters the pipeline and clears a prior failureReason. Fresh DRAFTs
-  // start from the compose page, never from here (the redirect above).
-  const handleStart = async () => {
+  // A stage failure retries from the earliest failed stage, preserving all
+  // completed upstream work. Failures before any stage row exists (init-ws,
+  // plan setup, or a stranded CREATED hand-off) still require a full start.
+  const failedStage =
+    detail?.intent.status === 'FAILED'
+      ? (stageRows.find((stage) => stage.state === 'FAILED') ?? null)
+      : null;
+  const handleRecovery = async () => {
     if (!projectId || !intentId) return;
     setStarting(true);
     setActionError(null);
     try {
-      await intentsService.start(projectId, intentId);
-      await reload();
+      if (failedStage) {
+        await rewindIntent(failedStage.stageId);
+      } else {
+        await intentsService.start(projectId, intentId);
+        await reload();
+      }
     } catch (err) {
-      setActionError(err instanceof Error ? err.message : 'Failed to start intent');
+      setActionError(err instanceof Error ? err.message : 'Failed to recover intent');
     } finally {
       setStarting(false);
     }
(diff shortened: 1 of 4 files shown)


lambda/v2-orchestrator/index.js
```diff
@@ -864,6 +864,17 @@ const handler = async (event, ctx, deps = defaultDeps()) => {
           };
         }
 
+        // Credential resolution only permits active executions. Unpark META
+        // before AgentCore restores a released session's workspace, otherwise
+        // the re-clone is rejected while the execution still reads WAITING.
+        await ctxArg.step(`gate-unpark-${humanTaskId}`, () =>
```
JWThewes: Could we make this transition conditional on both `status === WAITING` and the current `orchestratorRunId`? There is a race here: a cancel or rewind can win after the ownership read above, and this unconditional write would put the retired run back into `RUNNING` and let it dispatch the resume. I think a failed CAS should be handled as `retired`.

frontend/src/pages/IntentView.tsx
```diff
@@ -87,18 +90,26 @@ export default function IntentView() {
...
-  // Restart (FAILED / stranded CREATED) via the /start endpoint, which
-  // re-enters the pipeline and clears a prior failureReason. Fresh DRAFTs
-  // start from the compose page, never from here (the redirect above).
-  const handleStart = async () => {
+  // A stage failure retries from the earliest failed stage, preserving all
+  // completed upstream work. Failures before any stage row exists (init-ws,
+  // plan setup, or a stranded CREATED hand-off) still require a full start.
+  const failedStage =
+    detail?.intent.status === 'FAILED'
+      ? (stageRows.find((stage) => stage.state === 'FAILED') ?? null)
```
JWThewes: Can we keep the same `row.planned` guard used by the existing retry actions? `stageRows` also contains live rows that are no longer in the resolved plan. Picking one of those sends `/rewind` an invalid target, so this turns a recoverable full restart into a dead end. If there is no planned failed row, I would fall back to `/start`.

JWThewes (changes requested):
The happy path looks right, but I think these two cases need to be addressed before merging.

jeromevdl:
Both review issues were valid and are fixed.

  - Gate unpark now CAS-checks WAITING status and run ownership, returning retired if cancel/rewind wins the race.
  - Recovery only rewinds planned failed stages; obsolete stage rows fall back to full restart.
  - Added regression tests for both cases.

  Verification:

  - Frontend: 463 tests passed.
  - Orchestrator: 151 passed together; one flaky integration test passed independently, covering all 152.
  - Typecheck, formatting, lint, and orchestrator build passed.

JWThewes (approved):

PR #403, opened 2026-08-13
feat: export native AI-DLC workspaces

### LINKED ISSUE #400: [Feature]: Export AIDLC workflow assets for local development
### Description

Allow a user to export an intent's current workflow and generated assets into a self-contained bundle that can be placed in a local repository and resumed with the native [AI-DLC workflow ](https://github.com/awslabs/aidlc-workflows/tree/v2) from an IDE or CLI.

The export should:

1. Use the native harness for the intent's configured agent CLI by default.
2. Preserve the effective workflow composition used by the intent.
3. Include the current generated Markdown artifacts in native AI-DLC paths.
4. Reconstruct enough native state to continue at the first unfinished stage.
5. Exclude source code, credentials, session data, and secret values.

The export is a point-in-time, one-way handoff. Changes made locally are not synchronized back to the collaborative run.

### Use case

Users may want to run the workflow locally, especially developers during construction phase (code-generation), because they are used to work in an IDE or with a CLI. 

Fast path / first step of #324.

### Area

Backend (Lambda)

## Current State

The data needed for an export exists, but it is split across several stores:

- DynamoDB stores workflow and block metadata.
- S3 stores content-addressed block bodies and scripts under `blocks/bodies/sha256/` and `blocks/scripts/sha256/`.
- S3 stores the collaborative runtime machinery under `aidlc-runtime/<ref>/`, but this is not a complete native harness distribution.
- DynamoDB stores intent process state, stages, attempts, gates, effective workflow information, and the selected `agentCli`.
- Neptune stores the current generated artifact heads, including Markdown content and the stage or unit that produced each artifact.
- S3 stores project custom rules and intent attachments.

The collaborative runtime does not persist a native AI-DLC workspace. It clones source repositories and writes transient `.aidlc` prompt and CLI configuration files, while authoritative execution state remains in DynamoDB and Neptune. In particular, it does not create:

- `aidlc-state.md`
- Native audit shards
- Native phase and stage artifact directories
- A complete `.claude`, `.codex`, `.kiro`, or opencode harness

An export therefore cannot copy an existing S3 prefix. It must project the collaborative data model into a native AI-DLC workspace.

Native AI-DLC State Version 7 stores an intent under:

```text
aidlc/spaces/<space>/intents/<date-label>/
```

Normal artifacts live under phase and producer-stage paths. Construction artifacts may also be scoped by unit, while reverse-engineering artifacts live in the space-level `codekb/<repository>/` tree. Native state supports per-stage `EXECUTE` and `SKIP` suffix overrides. The exporter emits every stage from the harness stage graph and marks every non-selected stage `[S]`, so the stock scope grid cannot re-enable a stage omitted by the collaborative composition.

### Version Fidelity

Workflow planning and AgentCore now resolve each placed stage through its recorded tenant and pinned version. This closes the stage-version drift that previously allowed a workflow snapshot to execute against `LATEST`.

The native distribution still owns its methodology stage, agent, sensor, rule, and knowledge files. Phase one accepts custom compositions of the stock native methodology and project-uploaded Markdown rules, but rejects an intent that resolves an edited methodology block. The rejection lists the incompatible block IDs. A later version may add a validated native overlay format.

### Export Request

Add an authenticated endpoint:

```http
POST /projects/{projectId}/intents/{intentId}/export
```

Request:

```json
{
  "harness": "codex"
}
```

`harness` is optional and supports:

- `claude`
- `codex`
- `kiro`
- `kiro-ide`
- `opencode`

When omitted, it defaults to the intent's stored `agentCli`. The existing `kiro` agent value maps to Kiro CLI; Kiro IDE is an explicit override.

The response is:

```json
{
  "downloadUrl": "<short-lived presigned URL>",
  "filename": "aidlc-<intent-label>-<export-id>.zip",
  "expiresAt": "<timestamp>",
  "warnings": []
}
```

The service creates the archive under:

```text
workflow-exports/<intentId>/<exportId>.zip
```

The presigned download URL should be short-lived, and an S3 lifecycle rule should delete generated exports after one day.

### Snapshot Semantics

Version 1 supports intents in stable states:

- `DRAFT`
- `WAITING`
- `FAILED`
- `CANCELLED`
- `SUCCEEDED`

A changing `RUNNING` intent is rejected with `409 Conflict` because DynamoDB stage state and Neptune artifact heads cannot be read atomically. A later version can support an explicit pause-and-export operation if required.

Exporting does not pause, cancel, or otherwise mutate the intent.

The bundle contains current artifact heads only. Archived artifact revisions, agent transcripts, CLI session stores, and source repositories are excluded. For a multi-repository intent, the user extracts the bundle into an empty workspace directory and uses native AI-DLC's `repos.json` and `aidlc-workspace-sync` support to clone the recorded branches as sibling repositories.

### Bundle Contents

The archive contains:

- The selected native AI-DLC harness distribution.
- State-file `EXECUTE`/`SKIP` overrides representing the intent's effective composition over the complete native stage graph.
- The commit-pinned stock native methodology and applicable project custom rules.
- State Version 7 files for the exported intent.
- Generated Markdown artifacts in canonical native paths.
- Native `intents.json` repository metadata.
- A native `repos.json` for multi-repository workspace reproduction.
- `export-manifest.json` with provenance and checksums.

The exporter does not introduce handoff-specific Markdown documents. Workflow context must be represented through existing native state, audit, artifact, memory, knowledge, and rule structures. Transport-only information belongs in `export-manifest.json` and is not a methodology artifact.

The manifest includes:

- Source project and intent IDs
- Intent title and source branch
- Workflow ID and version
- Native upstream ref and distribution version
- Selected harness
- Effective composition grid and skip overlay
- Repository clone URLs, branches, and expected base revisions
- Export timestamp
- File checksums
- Compatibility warnings

### State Projection

The exporter aggregates stage instances and attempts into native stage state:

- A fully successful stage is `[x]`.
- A stage excluded by the effective workflow is `[S]`.
- The first unfinished stage is `[-]`.
- Later unfinished stages are `[ ]`.

Failed or partially completed stages are rerun locally. An open cloud human gate is not transplanted as a live callback; its stage becomes the first unfinished stage and is rerun using the native workflow.

For per-unit construction stages, the exporter aggregates attempts by unit and only marks the stage complete when every required unit completed successfully. The generated audit records explain any aggregation or rerun decision.

### Artifact Projection

Use the artifact's canonical `artifactType`, not an agent-selected display ID, to determine its filename.

Write mapped artifact heads to:

```text
# Normal artifact
<intent-record>/<phase>/<producer-stage>/<artifact-type>.md

# Per-unit construction artifact
<intent-record>/construction/<unit>/<producer-stage>/<artifact-type>.md

# Reverse-engineering artifact
aidlc/spaces/<space>/codekb/<repository>/<artifact-type>.md
```

An artifact that cannot be mapped to a declared native path is a compatibility failure. Return `409 Conflict` with the artifact identity and mapping reason. The exporter must neither drop it nor invent a non-native Markdown location.

### Multi-Repository Workspace

For intents with multiple repositories, export one native workspace root:

```text
<workspace>/
|-- <harness-directory>/
|-- aidlc/
`-- repos.json
```

The native intent registry records the repository directory names. The generated `repos.json` records each clone URL and the intent branch that should be checked out for a new clone. Source repositories are not placed in the archive.

After extraction, the user runs the selected harness's existing `aidlc-workspace-sync` tool. It clones missing repositories as siblings and generates `aidlc.code-workspace`. The developer's own Git credentials are used.

The export UI, rather than a generated Markdown README, presents the extraction, workspace-sync, doctor, and resume instructions.

### Compatibility Validation

Before creating an archive, validate that the effective collaborative workflow can be represented by native AI-DLC:

- Supported native phases
- Supported stage execution modes
- Valid stage dependencies and ordering
- Known artifact definitions
- Resolvable pinned block versions
- A compatible native distribution for the selected harness
- Safe, collision-free native paths

Return `409 Conflict` with structured incompatibility details when the workflow cannot be projected. A best-effort executable bundle is more dangerous than an explicit rejection because it can resume with different
instructions from the cloud run.

## Security and Privacy

- Require project membership for every export.
- Do not include OAuth tokens, API keys, MCP secret values, environment secret values, CLI credentials, session stores, or Git credentials.
- Include project rules only when they are part of the intent's effective workflow context.
- Sanitize user-controlled names before using them as paths or filenames.
- Log export identity and outcome, but not artifact bodies or secret-bearing configuration.
- Apply S3 encryption, short URL expiry, and lifecycle deletion.

jeromevdl:
### Native workspace export

- Export a stable Collaborative AI-DLC checkpoint as a runnable native AI-DLC workspace for Claude, Codex, Kiro CLI, Kiro IDE, or OpenCode.
- Reconstruct native State Version 7 with the workflow scope, completed and skipped stages, and the correct resume point.
- Materialize generated Markdown work products in their canonical native phase, stage, and per-unit paths.
- Reconstruct pending and answered `*-questions.md` files from human gates when no persisted Markdown artifact exists.
- Preserve arbitrary custom workflow compositions by registering their native scope definition, stage frontmatter, `stage-graph.json`, and `scope-grid.json`.
- Include custom project rules and repository URL/branch metadata. Source code and credentials are intentionally excluded.
- Reconstruct multi-unit Construction state: complete Bolt DAG, merged-unit receipts, autonomy mode, and next-unit resume point.
  - Reconstruct native audit stage events and runtime-graph.json.

### Version and snapshot fidelity

- Resolve the native harness from the intent's pinned AI-DLC commit, fetch it on demand, and cache the immutable archive in S3.
- Support both legacy `aidlc-docs/` and newer `aidlc/spaces/...` workspace layouts by inspecting the selected harness.
- Preserve pinned stage versions and compatibility with legacy workflow snapshots.
- Detect whether the pinned harness ships aidlc-workspace-sync.ts; otherwise provide manual sibling-repository setup.
- Use the root Terraform AI-DLC pin consistently across Lambdas.

### Platform integration

- Add the export API, archive generation, presigned downloads, S3 lifecycle cleanup, IAM permissions, and deployment support.
- Increase export Lambda capacity for uncached harness retrieval.

### User experience

- Add a workspace download action with harness selection:

image

- Warn that subsequent local work will not synchronize back into Collaborative AI-DLC traceability:

image

- Provide post-download repository and workspace setup commands when the exported checkpoint is entering Construction or a later phase:

image

## Verification

- pre-commit hook: 125 test files, 2,490 tests passed
- frontend production build passed
- intents Lambda build passed
- frontend lint and formatting checks passed
- intents lint and formatting checks passed
- Add native runtime recompilation checks against pinned 2.3.3 and current v2, see screenshots.
- Tested live against "legacy" AIDLC (`ba0cfe999856033ecb909a9135b46fe10811bf55`) and a multi-repo space with newer version (`83ed7a812c4024904f2c5e4d744e28077e0a5acd`)

Fixes #400

### DIFF
--- a/frontend/src/pages/IntentView.tsx
+++ b/frontend/src/pages/IntentView.tsx
@@ -841,20 +841,20 @@ export default function IntentView() {
               <li>
                 Clone each repository into the workspace:
                 {constructionSetup.repositories.map((repository) => (
-                  <div key={repository.name} className="mt-3 space-y-3">
+                  <div key={repository.id} className="mt-3 space-y-3">
                     <div>
-                      <p className="text-xs font-medium">Fresh clone: {repository.name}</p>
-                      <CommandBlock label={`Copy fresh clone command for ${repository.name}`}>
-                        {`git clone --branch ${shellQuote(repository.branch)} ${shellQuote(repository.url)} ${shellQuote(repository.name)}`}
+                      <p className="text-xs font-medium">Fresh clone: {repository.id}</p>
+                      <CommandBlock label={`Copy fresh clone command for ${repository.id}`}>
+                        {`git clone --branch ${shellQuote(repository.branch)} ${shellQuote(repository.url)} ${shellQuote(repository.directory)}`}
                       </CommandBlock>
                     </div>
                     <div>
-                      <p className="text-xs font-medium">Existing clone: {repository.name}</p>
-                      <CommandBlock label={`Copy existing clone commands for ${repository.name}`}>
+                      <p className="text-xs font-medium">Existing clone: {repository.id}</p>
+                      <CommandBlock label={`Copy existing clone commands for ${repository.id}`}>
                         {[
-                          `git -C ${shellQuote(repository.name)} fetch origin`,
-                          `git -C ${shellQuote(repository.name)} switch ${shellQuote(repository.branch)}`,
-                          `git -C ${shellQuote(repository.name)} pull --ff-only`,
+                          `git -C ${shellQuote(repository.directory)} fetch origin`,
+                          `git -C ${shellQuote(repository.directory)} switch ${shellQuote(repository.branch)}`,
+                          `git -C ${shellQuote(repository.directory)} pull --ff-only`,
                         ].join('\n')}
                       </CommandBlock>
                     </div>
@@ -889,21 +889,21 @@ export default function IntentView() {
               <li>
                 Retrieve the source repository and its intent branch:
                 {constructionSetup.repositories.map((repository) => (
-                  <div key={repository.name} className="mt-3 space-y-3">
+                  <div key={repository.id} className="mt-3 space-y-3">
                     <div>
                       <p className="text-xs font-medium">Fresh clone</p>
-                      <CommandBlock label={`Copy fresh clone commands for ${repository.name}`}>
+                      <CommandBlock label={`Copy fresh clone commands for ${repository.id}`}>
                         {[
-                          `git clone --branch ${shellQuote(repository.branch)} ${shellQuote(repository.url)} ${shellQuote(repository.name)}`,
-                          `cd ${shellQuote(repository.name)}`,
+                          `git clone --branch ${shellQuote(repository.branch)} ${shellQuote(repository.url)} ${shellQuote(repository.directory)}`,
+                          `cd ${shellQuote(repository.directory)}`,
                         ].join('\n')}
                       </CommandBlock>
                     </div>
                     <div>
                       <p className="text-xs font-medium">Existing clone</p>
-                      <CommandBlock label={`Copy existing clone commands for ${repository.name}`}>
+                      <CommandBlock label={`Copy existing clone commands for ${repository.id}`}>
                         {[
-                          `cd ${shellQuote(`/path/to/${repository.name}`)}`,
+                          `cd ${shellQuote(`/path/to/${repository.directory}`)}`,
                           'git fetch origin',
                           `git switch ${shellQuote(repository.branch)}`,
                           'git pull --ff-only',
--- a/frontend/src/pages/IntentView.tsx
+++ b/frontend/src/pages/IntentView.tsx
@@ -128,6 +128,9 @@ export default function IntentView() {
   const [configurationOpen, setConfigurationOpen] = useState(false);
   const [reshapeOpen, setReshapeOpen] = useState(false);
   const [confirmExport, setConfirmExport] = useState(false);
+  const [selectedExportHarness, setSelectedExportHarness] = useState<
+    NativeExportHarness | undefined
+  >();
   const [requestedExportHarness, setRequestedExportHarness] = useState<
     NativeExportHarness | undefined
   >();
@@ -294,8 +297,9 @@ export default function IntentView() {
   const exportUnavailableReason = 'Not available before the workflow starts';
   const exportDisabled = exporting || !isExportable;
   const defaultExportHarness = intent.agentCli ?? undefined;
+  const activeExportHarness = selectedExportHarness ?? defaultExportHarness;
   const exportCli =
-    EXPORT_HARNESSES.find((option) => option.value === defaultExportHarness)?.label ??
+    EXPORT_HARNESSES.find((option) => option.value === activeExportHarness)?.label ??
     'native AI-DLC';
   const exportButtonLabel =
     intent.status === 'RUNNING'
@@ -373,10 +377,10 @@ export default function IntentView() {
                           <DropdownMenuItem
                             key={option.value}
                             disabled={exporting}
-                            onClick={() => requestExport(option.value)}
+                            onClick={() => setSelectedExportHarness(option.value)}
                           >
                             <span>{option.label}</span>
-                            {option.value === defaultExportHarness && (
+                            {option.value === activeExportHarness && (
                               <Check className="ml-auto h-4 w-4" aria-label="Current harness" />
                             )}
                           </DropdownMenuItem>
@@ -402,7 +406,7 @@ export default function IntentView() {
                       size="icon"
                       className="h-7 w-7 rounded-none"
                       disabled={exportDisabled}
-                      onClick={() => requestExport()}
+                      onClick={() => requestExport(activeExportHarness)}
                       aria-label={exportButtonLabel}
                     >
                       {exporting ? (
--- a/frontend/src/pages/IntentView.tsx
+++ b/frontend/src/pages/IntentView.tsx
@@ -1,6 +1,10 @@
 import { useState } from 'react';
 import { Navigate, useNavigate, useParams } from 'react-router';
-import { intentsService } from '@/services/intents';
+import {
+  intentsService,
+  type NativeExportHarness,
+  type NativeWorkflowExport,
+} from '@/services/intents';
 import { useIntent } from '@/contexts/IntentContext';
 import { useAuth } from '@/contexts/AuthContext';
 import { useProjectCache } from '@/hooks/useProjectsCache';
@@ -38,7 +42,11 @@ import {
   DropdownMenuSeparator,
   DropdownMenuTrigger,
 } from '@/components/ui/dropdown-menu';
+import { Tooltip, TooltipContent, TooltipProvider, TooltipTrigger } from '@/components/ui/tooltip';
 import {
+  Check,
+  ChevronDown,
+  Download,
   GitBranch,
   KeyRound,
   Loader2,
@@ -58,6 +66,22 @@ import {
 
 const TERMINAL_STATUSES = new Set(['FAILED', 'CANCELLED', 'SUCCEEDED']);
 const CREDENTIAL_FAILURE_CODES = new Set(['credential_unavailable', 'credential_invalid']);
+const EXPORTABLE_STATUSES = new Set(['DRAFT', 'WAITING', 'FAILED', 'CANCELLED', 'SUCCEEDED']);
+const EXPORT_HARNESSES: Array<{ value: NativeExportHarness; label: string }> = [
+  { value: 'claude', label: 'Claude' },
+  { value: 'codex', label: 'Codex' },
+  { value: 'kiro', label: 'Kiro CLI' },
+  { value: 'kiro-ide', label: 'Kiro IDE' },
+  { value: 'opencode', label: 'OpenCode' },
+];
+
+const shellQuote = (value: string) => `'${value.replaceAll("'", "'\"'\"'")}'`;
+
+const CommandBlock = ({ children }: { children: string }) => (
+  <pre className="mt-2 w-full min-w-0 max-w-full whitespace-pre-wrap break-all rounded-md bg-muted px-3 py-2 font-mono text-xs leading-relaxed">
+    <code>{children}</code>
+  </pre>
+);
 
 export default function IntentView() {
   const {
@@ -95,6 +119,12 @@ export default function IntentView() {
   const [repairing, setRepairing] = useState(false);
   const [configurationOpen, setConfigurationOpen] = useState(false);
   const [reshapeOpen, setReshapeOpen] = useState(false);
+  const [confirmExport, setConfirmExport] = useState(false);
+  const [requestedExportHarness, setRequestedExportHarness] = useState<
+    NativeExportHarness | undefined
+  >();
+  const [exporting, setExporting] = useState(false);
+  const [constructionExport, setConstructionExport] = useState<NativeWorkflowExport | null>(null);
   const [actionError, setActionError] = useState<string | null>(null);
 
   // A stage failure retries from the earliest failed stage, preserving all
@@ -170,6 +200,40 @@ export default function IntentView() {
     }
   };
 
+  const handleExport = async (harness?: NativeExportHarness) => {
+    setConfirmExport(false);
+    setExporting(true);
+    setActionError(null);
+    try {
+      const result = await intentsService.exportWorkflow(projectId, intentId, harness);
+      const download = document.createElement('a');
+      download.href = result.downloadUrl;
+      download.download = result.filename;
+      download.rel = 'noopener';
+      document.body.append(download);
+      download.click();
+      download.remove();
+      const currentPhase = detail?.intent.currentPhase?.toLowerCase();
+      const hasActiveConstruction = detail?.stages.some(
+        (stage) =>
+          stage.phase?.toLowerCase() === 'construction' &&
+          ['RUNNING', 'WAITING_FOR_HUMAN'].includes(stage.state),
+      );
+      if (currentPhase === 'construction' || hasActiveConstruction) {
+        setConstructionExport(result);
+      }
+    } catch (err) {
+      setActionError(err instanceof Error ? err.message : 'Failed to export workflow');
+    } finally {
+      setExporting(false);
+    }
+  };
+
+  const requestExport = (harness?: NativeExportHarness) => {
+    setRequestedExportHarness(harness);
+    setConfirmExport(true);
+  };
+
   if (!projectId || !intentId) return <div className="p-6">Intent not found</div>;
   if (loading && !detail) {
     return (
@@ -217,6 +281,18 @@ export default function IntentView() {
   const isCancellable = ['WAITING', 'CREATED', 'FAILED'].includes(intent.status);
   // Deletable (destructive): owner/admin, any status except mid-RUNNING.
   const isDeletable = canDelete && intent.status !== 'RUNNING';
+  const isExportable = EXPORTABLE_STATUSES.has(intent.status);
+  const exportUnavailableReason =
+    intent.status === 'RUNNING'
+      ? 'Not available while workflow is running'
+      : intent.status === 'CREATED'
+        ? 'Not available while workflow is starting'
+        : 'Not available for this workflow status';
+  const exportDisabled = exporting || !isExportable;
+  const defaultExportHarness = intent.agentCli ?? undefined;
+  const exportCli =
+    EXPORT_HARNESSES.find((option) => option.value === defaultExportHarness)?.label ??
+    'native AI-DLC';
   // Pre-stage progress: before any stage row exists, init-ws lifecycle events
   // are the only signal the run is doing something (they stream into the
   // sidebar Timeline); this strip keeps the main pane from looking dead.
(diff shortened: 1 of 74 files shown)


lambda/intents/index.js
```diff
@@ -1407,6 +1491,112 @@ export const handler = async (event) => {
       return response(405, { error: 'Method not allowed' });
     }
 
+    // POST /projects/{projectId}/intents/{intentId}/export — materialize a
+    // native AI-DLC workspace snapshot and return a short-lived S3 download.
+    // RUNNING is intentionally refused: stage and artifact rows can change
+    // while the archive is assembled, producing an internally inconsistent
+    // checkpoint.
+    if (intentId && httpMethod === 'POST' && path?.endsWith('/export')) {
+      const records = await store.getExecutionRecords(intentId, { includeOutputs: false });
```
JWThewes: This status check does not actually make the export a stable snapshot. We read the execution partition here, then resolve the plan and read Neptune artifacts later; a WAITING intent can be answered or edited in between. That can give us old stage/question state with newer artifacts in the same ZIP. Can we either export from an immutable revision, or capture and recheck a revision/status token immediately before upload and fail or retry if anything changed?

lambda/intents/index.js
```diff
@@ -1407,6 +1491,112 @@ export const handler = async (event) => {
...
+      try {
+        const legacyRefWarning = meta.aidlcRepoRef
+          ? []
+          : [
+              'This legacy intent did not pin a native upstream ref; the current deployment ref was used.',
+            ];
+        const exported = await createNativeExport({
+          s3,
+          bucket: ARTIFACTS_BUCKET(),
+          upstreamRef: meta.aidlcRepoRef || AIDLC_REPO_REF(),
```
JWThewes: I do not think `meta.aidlcRepoRef` is an execution pin yet. We persist it at intent creation and use it here, but AgentCore still reads the deployment-wide `AIDLC_REPO_REF`; after a pin update an existing intent can continue under the new ref and then export the old harness. Please thread the intent's ref through execution and preserve matching block/workflow snapshots, or record the actual ref per stage and reject mixed-ref exports.

lambda/intents/index.js
```diff
@@ -1147,6 +1152,85 @@ const mapIntent = (meta) => ({
...
+  if (provider === 'gitlab') return `git@gitlab.com:${value}.git`;
+  if (provider === 'bitbucket') return `git@bitbucket.org:${value}.git`;
+  return `git@github.com:${value}.git`;
+};
+
+const exportRepositories = (meta) =>
+  (meta.repos ?? []).map((repository) => {
+    const provider = meta.repoProviders?.[repository] || meta.gitProvider || 'github';
+    return {
+      name: repositoryName(repository),
```
JWThewes: We lose the canonical repository identity here. `org-a/api` and `org-b/api` are both valid project repositories, but both become `api`, so `repos.json` and the setup commands target the same directory. The reverse-engineering projector has the opposite problem and rejects `owner/repo` as a path segment. Could we keep `owner/repo` as the ID and derive a collision-safe local directory name once, then use that mapping everywhere?

lambda/intents/native-export.js
```diff
@@ -0,0 +1,695 @@
...
+      Key: distributionManifestKey(upstreamRef, harness),
+      Body: `${JSON.stringify(manifest, null, 2)}\n`,
+      ContentType: 'application/json',
+    }),
+  );
+  return { files, workspaceLayout };
+};
+
+const loadDistribution = async ({ s3, bucket, upstreamRef, harness }) => {
+  if (!isCommitSha(upstreamRef)) {
```
JWThewes, line 309: `aidlc_repo_ref` is documented and implemented elsewhere as SHA, tag, or branch, so this makes export fail for an otherwise supported deployment configuration. If export requires immutability, can we resolve the configured ref to a full SHA when the intent is created and persist that SHA, rather than rejecting tags and branches here?

lambda/shared/native-workspace-projector.js
```diff
@@ -0,0 +1,993 @@
...
+  }
+  return {
+    units: parsed.units.map((unit) => ({
+      name: unit.name,
+      depends_on: unit.depends_on,
+    })),
+    batches: parsed.batches,
+    skipMatrix: unitPlan.skipMatrix ?? {},
+    walkingSkeleton: unitPlan.walkingSkeleton ?? null,
+    autonomyMode: hasRecordedAutonomyMode ? unitPlan.autonomyMode : 'gated',
```
JWThewes: This changes workflow semantics by making a choice the user has not made. In cloud state, `autonomyMode` is intentionally null until the post-skeleton ladder; defaulting it to `gated` also causes us to emit an `AUTONOMY_MODE_SET` audit event that never happened. Please preserve `unset` and let the native workflow ask the ladder question.

frontend/src/pages/IntentView.tsx
```diff
@@ -163,6 +235,34 @@ export default function IntentView() {
     }
   };
 
+  const handleExport = async (harness?: NativeExportHarness) => {
+    setConfirmExport(false);
+    setExporting(true);
+    setActionError(null);
+    try {
+      const result = await intentsService.exportWorkflow(projectId, intentId, harness);
```
JWThewes, line 224: The API returns warnings specifically for provenance and runtime compatibility, but we never surface `result.warnings` here. In the no-setup case the user will not see them at all. Can we show warnings before or alongside the download, and add a test for the legacy-ref warning?

frontend/src/pages/IntentView.tsx
```diff
@@ -479,6 +688,243 @@ export default function IntentView() {
...
+              }}
+            >
+              {exporting ? 'Preparing workspace…' : 'Download workspace'}
+            </AlertDialogAction>
+          </AlertDialogFooter>
+        </AlertDialogContent>
+      </AlertDialog>
+
+      <AlertDialog
+        open={constructionExport !== null}
```
JWThewes: Suggestion: could we move this setup dialog into a `NativeExportSetupDialog` component and drive the four modes from shared step data? `IntentView` is now about 1,000 lines and the extraction, launch, and continue steps are repeated across each branch, which will make future setup changes easy to miss in one mode.

JWThewes (changes requested):
The feature direction looks useful, but I think the checkpoint and provenance guarantees need tightening before this is safe to merge. The inline comments cover the concrete changes I am asking for.

lambda/intents/index.js
jeromevdl: I will create some immutable checkpoints at the end of each stage, so that it will also permit the export while in status RUNNING (exporting the latest stable checkpoint).

jeromevdl: Addressed in [5a75ccf](https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/403/commits/5a75ccf308ae8cf9cc245cb9f9e12d396115fd29) by introducing lightweight immutable workflow checkpoints at completed stage boundaries. Checkpoints reference content-addressed artifact versions rather than storing ZIPs or artifact bodies. RUNNING exports use the latest checkpoint, while WAITING and terminal exports use live state with a final consistency recheck. This also enables safe exports during active execution without mixed snapshots.

lambda/intents/index.js
jeromevdl: Addressed in [47d71f8](https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/403/commits/47d71f8a7acc383005469608bf38a2e64fdf2b39). `aidlcRepoRef` is now an execution pin, not only an export pin. The intent’s resolved SHA and exact supporting block versions are stored on META, forwarded through orchestration, and used by AgentCore for the conductor and methodology library. Stage rows record the effective SHA, and export rejects mixed-revision executions. This covers parked sessions that resume on a newer AgentCore image.

lambda/intents/native-export.js
jeromevdl: Addressed in [47d71f8](https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/403/commits/47d71f8a7acc383005469608bf38a2e64fdf2b39). Configured tags and branches are resolved to a full commit SHA before intent creation and seeding. The immutable SHA is then used consistently for runtime snapshots, execution, checkpoints, and native export. 
Manually verified with v2.3.0: META, the running stage, and export-manifest.json all used 29a31f7899731b53f2b8d7f76cd223f9a8a25859.

lambda/intents/index.js
jeromevdl: Addressed in 0871912. Exports now preserve canonical repository IDs and derive one deterministic local directory mapping. Unique basenames remain unchanged; collisions such as `org-a/api` and `org-b/api` become `org-a_api` and `org-b_api`. The same mapping is used by `repos.json`, intent metadata, CodeKB paths, the export API, and frontend clone commands. Regression tests cover the duplicate-basename case.

lambda/shared/native-workspace-projector.js
jeromevdl: Addressed in [486a420](https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/403/commits/486a4200d5e52ff9f98be4190a1fc8ba557eef99). I had some issues with Kiro when letting unset, it didn't ask for the autonomy mode.

frontend/src/pages/IntentView.tsx
jeromevdl: Addressed in 86cd45a, creation of the `NativeExportSetupDialog`

frontend/src/pages/IntentView.tsx
jeromevdl: Addressed in 86cd45a:
- `frontend/src/pages/IntentView.tsx:222` opens the dialog when warnings exist, even without setup instructions.
- `frontend/src/components/intent/NativeExportSetupDialog.tsx:119` renders every warning in the amber alert. Lines 109–115 provide warning-only title/description.

jeromevdl:
[f968f86](https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/403/commits/f968f862451a5e29789c1625dda9fe95dc191ab8) adds immutable, commit-pinned methodology catalogs in S3. Seed writes the structured catalog before replacing SYSTEM records, and native export uses the intent’s historical catalog when the current SYSTEM revision differs. Missing catalogs can be rebuilt from the pinned commit. Manually verified with an intent pinned to 83ed7a8… after SYSTEM was reseeded to v2.3.0 (29a31f…); export succeeded and retained 83ed7a8… in the manifest.

JWThewes (changes requested):
# Re-review of 4b0d11e

Thanks @jeromevdl  — repo identity, autonomy mode and the warnings surfacing all look good. Three of
my comments are not fully closed, and I found a few new things.

## Not closed

**Tag/branch refs.** Unchanged. `index.js:1754` still passes
`meta.aidlcRepoRef || AIDLC_REPO_REF()` straight into `loadDistribution`, so on a
tag-pinned deployment every legacy intent hits the `isCommitSha` throw at
`native-export.js:277` and gets a 409 that blames the workflow snapshot. Please resolve
the env fallback through `resolveAidlcRepoRef`, or return an error naming the
misconfiguration.

**Execution pin.** The threading is real, but the pin does not survive a reseed. SYSTEM
blocks are always `version: 1` (`seed-blocks/index.js:112`), `methodologyPins` are
`{tenantId, version}` resolved by key (`v2-workflow-plan.js:78`), and
`clearSystemPartitions` rewrites those keys — which `deploy-terraform.sh` does on every
deploy. So a parked intent's next stage loads the new commit's bodies at the same `V#1`.
Either version SYSTEM blocks per ref, or have `run-stage` verify the loaded blocks'
`sourceRef` against `aidlcRepoRef` and fail loudly.

Two bugs alongside it: `resumeStageRow` ignores the `aidlcRepoRef` it is passed
(`v2-process-store.js:541-596` vs `run-stage.js:1541`), so it is never back-filled on
resume; and the mixed-ref check must `filter(Boolean)` because SKIPPED rows have no ref
(`index.js:1665`), so it cannot tell "skipped" from "ran under the old code". Could it
require a match for every row in `{SUCCEEDED, FAILED, RUNNING, WAITING_FOR_HUMAN}`?

**Checkpoints.** Mechanism and recheck placement are right. Two gaps:
`putWorkflowCheckpoint` is an unconditional put with no CAS on `orchestratorRunId`
(`v2-process-store.js:121`), and nothing invalidates the checkpoint on rewind — so a
retired run can clobber it, and a RUNNING export between `mark-running` and
`checkpoint-initial` serves pre-rewind state. Please stamp the run id, make the put
conditional on it, and delete the item in the rewind/repair paths.

## New

- `native-workspace-projector.js:698` — `workflow_id: intent.createdAt || now` is a
  copy-paste of `started_at` on line 700. `intent.intentId` is used correctly on line 609.
- `native-workspace-projector.js:299-308` — a stage skipped for *every* unit aggregates to
  `SUCCEEDED`, so we emit `[x] … EXECUTE`, a `STAGE_COMPLETED` audit entry that never
  happened, and a `memory_path` that does not exist.
- `native-workspace-projector.js:487` — `intent.prompt` is interpolated raw into
  `aidlc-state.md`. Prompts are multi-line; one containing `- **State Version**: 1`
  hijacks parsed state. Same at `:120` and `:112`.
- `terraform/modules/api/lambda/main.tf:2161-2172` — I think the `s3:prefix` condition
  defeats the purpose in the comment: `s3:prefix` is only set on List requests, so on a
  `GetObject` the statement does not apply and the miss comes back as 403, which
  `isNotFound` (`native-export.js:78`) rejects. That would fail the first export for every
  ref. You tested live, so please confirm — either way, treating 403 as a cache miss is
  cheap insurance.

## Follow-ups

Silent drops in the projection (unit `kind`, `skipMatrix`, null `artifactType` becoming
`intent.md`, non-`question` gates, unknown phases, `[?]`/`[R]` never emitted); 29s API
Gateway timeout against a 120s Lambda; `index.js:1806` mapping every error to 409; no
audit event for the export; `workflow-exports/` not purged on intent delete; Glacier
transition vs pinned `custom-rules/` VersionIds; `sourceRef` missing from
`RESERVED_FIELDS`; `IntentView.test.tsx:415` is a vacuous assertion (exact-match) and
`:565` asserts a string the backend never returns; `workspace-sync` has no rendering
coverage in the new dialog.

jeromevdl:
@JWThewes 

Thanks for the re-review. I addressed the findings as follows (hashes updated after rebasing onto main):

  - Tag/branch refs resolved to commit SHAs — 77ee65d
  - Pinned SYSTEM methodology refs validated fail-fast — 5ae3f74
  - resumeStageRow persists aidlcRepoRef — b46c453
  - Export validates refs for executed stages while ignoring PENDING/SKIPPED — bb645e8
  - Checkpoints carry run ownership, use CAS, and are invalidated on repair/rewind/recompose — 3f5aa02
  - Native workflow_id uses the intent ID — ff265b8
  - All-skipped unit stages remain SKIPPED — d5a758a
  - Structural Markdown fields are flattened — d3604e8
  - Controlled S3 cache lookups treat masked 403s as misses — 6ca2f09
  - Repository identity and collision-safe directories — 9ca758f
  - Unset construction autonomy remains unset — 203ab25
  - Export warnings are surfaced in the extracted dialog — a0f7755

  Follow-ups:

  - sourceRef is now reserved — 911b10a
  - Export error classification and projection validation/warnings were hardened — e8cdcce, dc05fa2
  - Successful exports emit an audit event — dc05fa2
  - Intent deletion purges versioned workflow-exports/ objects — dc05fa2
  - Added workspace-sync and stronger UI coverage — 655c896, 73f12b7
  - Noncurrent artifact versions now use immediately readable Glacier Instant Retrieval — 8138236

  Remaining non-blocking follow-up:

  - Export remains a synchronous API Gateway request, so the ~29-second integration limit can still be shorter than the Lambda timeout. Properly addressing this requires an asynchronous export job/polling flow; merely lowering the Lambda timeout would not solve the architectural limitation.

  Validation after rebase:

  - Backend: 139 files, 2,674 tests passed
  - Frontend: 73 files, 517 tests passed
  - Frontend typecheck passed
  - Lint and formatting passed
  - Terraform validation passed
  - Deployed and export tested locally

JWThewes:
@jeromevdl Once this one is able to be merged I'm happy to approve it - looks good!

JWThewes (dismissed):

JWThewes:
@jeromevdl I don't know why but there are two remaining merge conflicts on IntentView and tests.
Once this is resolved I'll re-approve directly

jeromevdl:
Rebased

JWThewes (approved):

PR #405, opened 2026-08-13
feat: hierarchical agent credentials and per-intent CLI selection

### LINKED ISSUE #404: [Feature]: Hierarchical agent credentials and per-intent CLI selection
### Description

Agent credentials are currently configured only at platform level. The Bedrock bearer token and Kiro API key are shared by every space and user, so all usage is charged to one Bedrock key or one Kiro identity. In a multi-user installation this makes cost attribution difficult and can exhaust the credits of the shared Kiro user very quickly.

The agent CLI is also selected at space level today, although the CLIs a user can actually run depend on which effective credentials are available when that user starts an intent.

Add agent credentials at three scopes:

1. **User**
2. **Space**
3. **Platform**

Resolve each credential type independently with this precedence:

```text
user > space > platform
```

An unset value falls through to the next scope. A configured but invalid credential must fail with an actionable error instead of silently falling back and charging a different scope.

### Use case

Organizations should be able to provide a platform-wide fallback, let a space use its own budget, or let an individual use personal credentials. This distributes usage and cost to the intended Bedrock account or Kiro user while retaining convenient shared defaults.

A user starting an intent should see and choose only the CLIs supported by the effective credentials available to that user in that space. If no credential is available at any scope, the intent cannot start and the user is directed to configure a personal credential first.

### Area

Agents / AI

### Credential scopes and permissions

- Platform admins manage platform credentials in **Admin -> Agents**.
- Space owners/admins manage space credentials in **Space Settings -> Agent**.
- Each user manages personal credentials in their user/profile settings.
- APIs never return secret values after saving them. They return only configured state and the effective source (`user`, `space`, or `platform`).
- Clearing a credential makes that scope unset and restores inheritance from the next scope.

Precedence is evaluated separately for the Bedrock bearer token and Kiro API key. This allows combinations such as a personal Kiro key together with a space-level Bedrock key.

### CLI and harness availability

CLI availability is the intersection of installed runtime capabilities and effective credentials:

| Effective credential | Available CLI / harness |
| --- | --- |
| Kiro API key | Kiro |
| Bedrock bearer token | Claude Code, OpenCode, Codex |
| Both | All installed CLIs above |
| Neither | None |

Kiro must never appear as available unless an effective Kiro API key exists. Likewise, the Bedrock-backed CLIs must not be selectable without an effective Bedrock bearer token.

### Intent start

- Move the execution CLI choice from being a space-level source of truth to the intent compose/start flow.
- Require the user to explicitly select one of the currently available CLIs before starting the intent.
- A space-level CLI setting may remain as a suggested default, but it must be confirmed at intent start.
- Snapshot the selected CLI on the intent so resumes and later stages use the same harness.
- Resolve credentials in the context of the user who starts the intent and bind the run to an opaque credential reference/source, not to a plaintext secret. Later gate answers by other collaborators must not silently switch the billing identity.
- If the bound credential is removed, unavailable, or invalid during a run, park/fail the run with an actionable credential error. Do not silently switch to another scope mid-intent.
- When no CLI is available, disable start and show a direct action to add a personal Bedrock or Kiro key.

### Runtime and secret handling

The runtime currently reads platform credentials at container startup. Supporting space and user credentials requires per-run resolution rather than relying only on global environment variables.

- Resolve the effective credential server-side at intent start/dispatch.
- Store credentials in an encrypted secret store with scope-specific paths and authorization.
- Persist only an opaque reference and credential source with execution metadata.
- Fetch/inject the secret only inside the execution boundary that needs it.
- Never include secret values in intent records, API responses, event payloads, logs, transcripts, or observability data.

## Acceptance criteria

- Bedrock and Kiro credentials can each be configured at platform, space, and user scope.
- For each credential type, resolution follows `user > space > platform` and falls through only when a value is unset.
- Effective CLI availability follows the credential mapping above, including mixed credential sources.
- Users choose the CLI when starting every new intent.
- Starting is blocked with a clear personal-key setup action when no CLI is available.
- The selected CLI and opaque credential binding are stable for the lifetime of the intent.
- Existing platform credentials continue to work as the fallback without migration by operators.
- Existing intents keep their already-snapshotted CLI behavior.
- Authorization prevents users from reading or modifying credentials outside their permitted scope.
- Automated tests cover precedence permutations, mixed Bedrock/Kiro scopes, no-credential behavior, authorization, credential removal/rotation, and secret redaction.

JWThewes:
## Summary

- add write-only Bedrock and Kiro credential settings at platform, space, and user scopes with `user > space > platform` precedence
- require an explicit CLI selection for Compose with AI and intent start, then pin the selected credential source for the intent lifetime
- resolve credentials per AgentCore invocation and add scoped APIs, IAM permissions, settings UI, capability indicators, and invalid-credential handling

## Testing

- commit hooks: formatting, lint, secret scanning, dependency audit, frontend typecheck, and 878 tests
- AgentCore suite: 773 tests
- intents suite: 203 tests
- agents suite: 10 tests
- frontend suite: 464 tests
- frontend, agents, and intents production builds
- Terraform validation for the changed API and AgentCore modules

Closes #404


lambda/agentcore/auth-resolver.js
```diff
@@ -30,20 +27,125 @@ const defaultGetParam = (client) => async (name) => {
...
+        ? (meta?.credentialBinding ?? legacyPlatformBinding(requestedCli))
+        : null);
+    if (binding && !bindingMatchesCli(binding, requestedCli)) {
+      throw Object.assign(new Error('Agent credential does not match the selected CLI'), {
+        code: 'credential_binding_mismatch',
+      });
+    }
+    if (binding) bindings = [binding];
+  } else {
+    const requestedCli = payload.requestedCli || meta?.agentCli || null;
```
jeromevdl, line 92: for discussion-assist-start on a DRAFT intent, payload does not contain `requestedCli` and `meta.agentCli` is still null. Therefore, no credentials are resolved, `availableClis` is empty, and the assistant fails. Could we pass and resolve the user’s selected CLI and credential binding for this command, and add a test for  this scenario?

jeromevdl:
Please also update documentation (at least `docs/using-the-platform/platform-settings.md`) which still says credentials are platform-only.

frontend/src/components/project-settings/AgentTab.tsx
```diff
@@ -303,37 +308,40 @@ export function AgentTab({ project, canEdit, onProjectUpdated }: Props) {
...
-            ok={isCliAvailable(project.agentCli ?? 'kiro')}
+            ok={isCliInstalled(project.agentCli ?? 'kiro')}
             okLabel={AGENT_CLI_CONFIG[project.agentCli ?? 'kiro'].label}
-            notOkLabel={`${AGENT_CLI_CONFIG[project.agentCli ?? 'kiro'].label} · unavailable`}
+            notOkLabel={`${AGENT_CLI_CONFIG[project.agentCli ?? 'kiro'].label} · not installed`}
             notOkTone="warning"
           />
         }
-        description="Which AI agent CLI runs this space's work — only CLIs installed in the deployment are selectable."
+        description="Highlighted when a user chooses a CLI for an intent; each user still makes an explicit selection."
```
jeromevdl: strange description, looks super generic and not targeting the user, more like a comment

frontend/src/components/settings/AgentCredentialScopeCard.tsx
```diff
@@ -0,0 +1,242 @@
...
+  agentsService,
+  type AgentCredentialStatus,
+  type SpaceAgentCredentialStatus,
+} from '@/services/agents';
+import { SettingsCard } from '@/components/settings/SettingsCard';
+import { ConfigStatusBadge } from '@/components/settings/ConfigStatusBadge';
+import { SecretField } from '@/components/settings/SecretField';
+import { SaveStatusButton, type SaveResult } from '@/components/settings/SaveStatusButton';
+
+type Scope = 'platform' | 'space' | 'personal';
```
jeromevdl, line 17: Wondering why we don't have the `intent` scope. As of now the intent binds to the selected scope, but if we want to create a dedicated api key for one intent ?

frontend/src/pages/IntentComposePage.tsx
```diff
@@ -64,6 +69,24 @@ const formatBytes = (bytes: number) =>
   `${(bytes / 1024 / 1024).toFixed(bytes >= 1024 * 1024 ? 1 : 2)} MB`;
 const ATTACHMENT_INGEST_POLL_MS = 500;
 const ATTACHMENT_INGEST_TIMEOUT_MS = 30_000;
+const AGENT_CLIS: AgentCli[] = ['kiro', 'claude', 'opencode', 'codex'];
+const AGENT_CLI_LABELS: Record<AgentCli, string> = {
```
jeromevdl: We have this config at different places in the frontend:
AgentTab.tsx, DefaultModelsCard.tsx, TierModelsSection.tsx and now also IntentView.tsx plus this one.

Could we centralize this in one place ?

lambda/agentcore/auth-resolver.js
```diff
@@ -30,20 +27,125 @@ const defaultGetParam = (client) => async (name) => {
...
+    resolvedProviders,
+    missingProviders,
+    missingCredentialBindings,
+  };
+};
+
+// Backward-compatible helper retained for focused tests and local tooling. It
+// resolves the legacy platform paths into the supplied object, but production
+// AgentCore no longer calls it or mutates process.env.
+export const resolveAgentAuth = async ({ env = {}, getParam } = {}) => {
```
jeromevdl: This function is not used anywhere apart from the tests.

lambda/agentcore/http-server.js
```diff
@@ -80,13 +80,29 @@ export const createBusyTracker = () => {
   };
 };
 
+// Only commands that can spawn an agent CLI need hierarchical credential
+// resolution. Engine-only commands must remain available during an SSM outage.
+const AGENT_AUTH_COMMANDS = new Set([
```
jeromevdl: We could probably simplify / merge this with the handler definition line 110 
There are also some if in the `resolveInvocationAgentAuth()` function that could benefit from this refactor, as we have several place where we check those strings.

```
const COMMANDS = {
    'init-ws': {
      handler: 'initWs',
      agentAuth: false,
    },
    'run-stage': {
      handler: 'runStage',
      agentAuth: true,
    },
    // ...
```

frontend/src/pages/IntentView.tsx
```diff
@@ -53,6 +55,27 @@ import {
...
+  opencode: 'OpenCode',
+  codex: 'Codex',
+} as const;
+const CREDENTIAL_SOURCE_LABELS = {
+  user: 'Personal',
+  space: 'Space',
+  platform: 'Platform',
+} as const;
+
+const credentialFailureMessage = (failureReason: string | null) => {
```
jeromevdl: A bit fragile, if the wording changes in the backend, it will silently fail. Maybe the API should return a proper structure (code +message)

lambda/agents/index.js
```diff
@@ -67,12 +74,19 @@ const AGENTCORE_RUNTIME_ARN = process.env.AGENTCORE_RUNTIME_ARN || '';
 // A session id >= 33 chars is required by InvokeAgentRuntime; the capabilities
 // command is stateless so any stable id works.
 const CAPABILITIES_SESSION_ID = 'aidlc-capabilities-probe-00000001';
+const PLATFORM_CREDENTIAL_BINDINGS = {
```
jeromevdl: in `shared/agent-credentials.js` we already define a set of constants that could be reused here to avoid creating a copy of the mapping

terraform/modules/compute/agentcore/main.tf
```diff
@@ -362,15 +362,18 @@ resource "aws_iam_role_policy" "agentcore" {
...
+          # via the bearer token / API key).
           Effect = "Allow"
           Action = ["ssm:GetParameter", "ssm:GetParameters"]
           Resource = [
             aws_ssm_parameter.bedrock_bearer_token.arn,
             aws_ssm_parameter.kiro_api_key.arn,
             aws_ssm_parameter.cli_models.arn,
             aws_ssm_parameter.tier_models.arn,
+            "arn:${local.partition}:ssm:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:parameter/${var.project_name}/${var.environment}/users/*/agent-credentials/*",
+            "arn:${local.partition}:ssm:${data.aws_region.current.region}:${data.aws_caller_identity.current.account_id}:parameter/${var.project_name}/${var.environment}/projects/*/agent-credentials/*",
```
jeromevdl: This resource is correctly limited to the current application and environment, but the wildcards still allow the shared AgentCore role to decrypt every user and space credential in that deployment. AgentCore also has permission to scan execution metadata containing credential bindings, so these paths are discoverable. Could credential lookup be moved behind the existing broker pattern, so AgentCore can request only the binding authorized for the current invocation? (AgentCore -> agent-credential broker -> SSM)

JWThewes:
@jeromevdl addressed your comments

lambda/agentcore/auth-resolver.js
```diff
@@ -30,20 +27,125 @@ const defaultGetParam = (client) => async (name) => {
...
+      base,
+      binding,
+      projectId: payload.projectId || meta?.projectId || null,
+    });
+    if (!value) {
+      missingProviders.push(binding.provider);
+      missingCredentialBindings.push(credentialBinding);
+      continue;
+    }
+    invocationEnv[credentialEnvName(binding.provider)] = value;
```
jeromevdl, line 196: This PR necessarily passes credentials through the CLI environment. Output redaction should be handled separately and consistently for both agent credentials and MCP environment variables. I’ll track application-level redaction and CloudWatch data-protection policies in a follow-up issue: #406

jeromevdl (dismissed):

jeromevdl:
> @jeromevdl addressed your comments

Not sure everything was addressed... Please mark resolved the comments you have fixed.

frontend/src/pages/IntentComposePage.tsx
JWThewes: Addressed in 3587662. I added a single typed `frontend/src/lib/agentCli.ts` source for CLI order, display/model labels, descriptions, credential-provider mapping, model help/placeholders, and credential-source labels. `AgentTab`, `DefaultModelsCard`, `TierModelsSection`, `IntentComposePage`, and `IntentView` now all consume that metadata. Frontend typecheck and the 68 focused tests pass.

lambda/agentcore/auth-resolver.js
JWThewes: Addressed in 4ce6d0e. `resolveAgentAuth()` and its test-only `defaultGetParam` path have been removed, along with the four tests that exclusively exercised it. I also removed the now-unused `BEDROCK_BEARER_TOKEN_SSM_PATH` and `KIRO_API_KEY_SSM_PATH` runtime environment variables from Terraform. The invocation-scoped auth tests and commit checks pass.

lambda/agentcore/http-server.js
JWThewes: Addressed in dfb1cb2. I introduced `command-registry.js` as the single source for each command’s handler and agent-auth mode. `dispatchInvocation()` now routes and decides whether/how to prepare credentials from that registry. The auth resolver uses mode-specific binding strategies (`execution`, `compose`, `discussion`, `capabilities`) rather than checking command strings, and duplicate binding validation is factored out. The full AgentCore suite passes: 45 test files / 782 tests.

frontend/src/pages/IntentView.tsx
JWThewes: Addressed in 6cdf2ec. Execution META and the intents API now carry a structured `failure: { code, message }` alongside the legacy `failureReason`. The orchestrator preserves the underlying stage failure code/message, the API normalizes old rows at the boundary, and `IntentView` no longer parses backend wording. Added persistence, orchestrator, API compatibility, and UI coverage; focused suites and the 631-test pre-commit suite pass.

lambda/agents/index.js
JWThewes: Addressed in dcbafb3. `PLATFORM_CREDENTIAL_BINDINGS` is now derived from the shared `AGENT_CREDENTIAL_PROVIDERS` registry instead of repeating the `bedrock`/`kiro` mapping locally. This keeps the capabilities probe in sync if the shared provider set changes. The focused agents tests pass.

terraform/modules/compute/agentcore/main.tf
JWThewes: Addressed in ef6a6ab. AgentCore no longer has direct SSM access to platform, space, or user agent credentials. Trusted API/orchestrator Lambdas now mint a five-minute HMAC-signed grant containing the exact credential binding(s), purpose, project, and execution. AgentCore redeems that grant through the existing credential broker; the broker verifies it and reads only the named SSM path(s), and AgentCore rejects any purpose, project, execution, duplicate, or binding mismatch before installing invocation-local credentials. The grant is stripped before the command handler runs. Started discussion assists also use the intent's pinned binding rather than the current collaborator's binding. Added tamper, expiry, wrong-user, cross-project, duplicate-response, and pinned-binding coverage; all affected bundles build, Terraform validates, and the final changed-test hook passed 459 tests.

lambda/agentcore/auth-resolver.js
JWThewes: Addressed in 0e37f0d and retained/hardened in ef6a6ab. For a DRAFT discussion assist, the discussions API accepts the selected agentCli, resolves that caller's effective user > space > platform binding, and sends requestedCli plus credentialBinding to AgentCore without sending the secret itself. The discussion auth mode uses that payload binding when META has no pinned CLI yet, so availableClis is populated correctly. Once an intent has started, the API and runtime instead use the intent's pinned CLI/binding. Tests cover both the DRAFT caller-binding path and the started-intent pinned-binding path.

frontend/src/components/project-settings/AgentTab.tsx
JWThewes: Addressed in 0e37f0d. The generic/comment-like wording was replaced with direct user-facing guidance: 'Set the CLI shown as recommended on the intent compose page. Members still choose a CLI for each intent.'

frontend/src/components/settings/AgentCredentialScopeCard.tsx
JWThewes: Clarified in 0e37f0d. The missing intent scope is intentional for this feature: an intent pins an opaque binding to the selected personal, space, or platform credential at start, but it does not own or copy a secret. That gives the run stable scope semantics while still allowing rotation at the bound path. A dedicated per-intent API key would require a separate storage, authorization, rotation, cleanup, and UI lifecycle, so it should be designed as a separate feature rather than added implicitly here. This is now stated explicitly in docs/using-the-platform/platform-settings.md and alongside the credential-scope type.

lambda/agentcore/auth-resolver.js
JWThewes: Agreed—application-output redaction and CloudWatch data-protection policy are intentionally tracked separately in open issue #406. This PR remains focused on credential selection and invocation isolation. The broker path only logs allowlisted error codes and ef6a6ab strips the short-lived grant before command handlers run, but comprehensive agent/MCP output redaction remains follow-up work in #406.

JWThewes:
Documentation was updated in 0e37f0d, including docs/using-the-platform/platform-settings.md. The docs now cover personal > space > platform precedence, where each scope is managed, SSM paths, per-invocation binding resolution, start-time pinning/rotation behavior, and why there is no intent-owned secret store. Related updates are also in prerequisites.md, projects.md, first-intent.md, and concepts/architecture.md. I have now replied to and resolved every inline review thread individually; the final broker/IAM fix is in ef6a6ab.

lambda/agentcore/auth-resolver.js
```diff
@@ -1,49 +1,207 @@
...
+      source: binding.source,
+    };
+    credentialBindings.push(credentialBinding);
+    const value = authorized.get(bindingKey(binding))?.value || '';
+    if (!value) {
+      missingProviders.push(binding.provider);
+      missingCredentialBindings.push(credentialBinding);
+      continue;
     }
+    invocationEnv[credentialEnvName(binding.provider)] = value;
```
eipasteur, line 196: I think this is where the new personal-credential boundary leaks. The value ends up in run-stage's `childEnv` (`run-stage.js:1947-1952`), and for Claude, Kiro and OpenCode the stdio MCP servers inherit that flat env; the comment block right above `childEnv` calls this an accepted trade-off. That was fine while the only thing in there was the platform token. With user bindings, a project owner/admin can configure a custom MCP server that reads a member's personal `KIRO_API_KEY` / `AWS_BEARER_TOKEN_BEDROCK` whenever that member runs an intent. `RESERVED_MCP_ENV_KEYS` only rejects `${VAR}` refs using those names; it doesn't remove them from what children inherit. Codex already does the right thing here (sanitized env, `env_vars` whitelist reserved to `aidlc`), which seems like the model to follow. Could we scrub the model-auth variables from custom MCP children for the other three CLIs (or go through a brokered per-process mechanism), and add a regression test proving a custom MCP server can't observe the selected user's credential? I know #406 covers output redaction. This feels like a separate issue since it's about inheritance, not logging.

lambda/shared/agent-credentials.js
```diff
@@ -0,0 +1,261 @@
...
+    ]),
+  );
+
+const fetchValues = async (ssm, paths) => {
+  const names = [...new Set(Object.values(paths))];
+  if (names.length === 0) return {};
+  const result = await ssm.send(
+    new GetParametersCommand({
+      Names: names,
+      WithDecryption: true,
```
eipasteur, line 137: I don't think the broker is the sole decryption boundary yet. `fetchValues` decrypts, and it's reached from `resolveEffectiveCredentialBindings()` in the intents and discussions Lambdas, and from both that and `readCredentialScopeStatus()` in the agents Lambda. The intents and discussions roles have `ssm:GetParameter(s)` over `users/*/agent-credentials/*` and `projects/*/agent-credentials/*` (`terraform/modules/api/lambda/main.tf:1750-1758`, `:2178-2188`), so a code-execution or logging bug in any of the three can read raw keys without a grant, even though they only need "configured + source" to mint a binding. Could we separate presence metadata from value redemption and reserve `WithDecryption: true` for the broker? One caveat: platform-scope clears write the literal `placeholder` SecureString, so platform presence currently requires decrypting. User/space scopes delete the parameter, so existence would be enough there. Might be worth moving the platform sentinel out of the secret value as part of this.

lambda/intents/index.js
```diff
@@ -2191,9 +2316,16 @@ export const handler = async (event) => {
           orchestratorExpiresAt: null,
           // Clear any stale failure from a prior attempt as we re-enter the pipeline.
           failureReason: null,
+          failure: null,
           startedBy: starter.sub,
           starterName: starter.displayName,
           starterEmail: starter.email,
+          ...(priorStatus === 'DRAFT'
+            ? {
+                agentCli: selectedAgentCli,
```
eipasteur, line 2322: I think there's a read-after-write race here. The pin is written, then `invokeOrchestrator` fires with only `{ action, intentId, executionId }`, and the orchestrator's `load-meta` step (`v2-orchestrator/index.js:316`) is a plain `GetItem` without `ConsistentRead` (`v2-process-store.js:111-115`). The CLI and binding it reads there are closed over for every grant in the run. If that read returns the previous DRAFT image: a new draft has no CLI/binding, so no grant is minted, `availableClis` is empty and the stage fails `no_cli`; a legacy draft still has its project CLI and a null binding, so the orchestrator falls back to a platform grant (`:319-322`). And since AgentCore rereads META on its own (`auth-resolver.js:118-120`), if its read is fresh while the orchestrator's was stale we'd get `credential_grant_required` or `credential_grant_mismatch` instead. So the run can intermittently fail, or bill a different scope than the one the starter selected. It seems like a strongly consistent initial META read, or carrying the pin in the handoff and verifying it against META, would close this, ideally with a stale-read regression test.

frontend/src/components/settings/AgentCredentialScopeCard.tsx
```diff
@@ -0,0 +1,244 @@
...
+      setPlatformFallback(null);
+      return;
+    }
+    if (!projectId) throw new Error('projectId is required for space credentials');
+    const result: SpaceAgentCredentialStatus = await agentsService.getProjectCredentials(projectId);
+    setSettings(result);
+    setPlatformFallback(result.platformFallback);
+  }, [projectId, scope]);
+
+  useEffect(() => {
```
eipasteur, line 85: For the space-scoped instance, changing `projectId` reruns this effect but only resets status/error. `bearerToken` and `kiroApiKey` survive, and the in-flight load has no identity guard. Neither the card (`AgentTab.tsx:267`) nor the `/space/:projectId/settings` route is keyed, so React reuses the instance when navigating from space A to space B: an unsaved A credential can get saved into B, and a late A response can overwrite B's displayed status. I'd suggest clearing the write-only inputs and the pending save/error state whenever `{ scope, projectId }` changes and ignoring stale responses (or simply keying the card on `projectId`), with a delayed A to B route-switch test.

frontend/src/pages/IntentComposePage.tsx
```diff
@@ -86,12 +94,49 @@ export default function IntentComposePage() {
...
+          );
+        }
+      })
+      .finally(() => {
+        if (!cancelled) setCapabilitiesLoading(false);
+      });
+    return () => {
+      cancelled = true;
+    };
+  }, [projectId]);
```
eipasteur, line 138: `selectedCli` is per-intent but this reset is keyed only on `projectId`. The compose route is unkeyed, so browser back/forward (or any same-route navigation) between draft A and draft B in the same space reuses the instance and B keeps A's CLI, with Start enabled without a fresh explicit selection. The normal in-app path seems to unmount (Project list to `/intent/:id`, then redirect to `/compose`), so this is mostly a history-navigation case, but it's real. The second effect is that `DiscussionProvider` clears `assistAgentCli` on `intentId` while the mirror effect above is keyed on `selectedCli` only, so after that navigation Start shows CLI X and Quorum assist sends none (see the discussions comment). Including `intentId` in this reset (or keying the page on it) plus a same-space draft-switch test would cover both.

lambda/shared/agent-credentials.js
```diff
@@ -0,0 +1,261 @@
...
+  userId = null,
+}) => {
+  const prefix = normalizeBase(base);
+  if (!prefix) throw new Error('Agent credential store is not configured');
+  const config = PROVIDER_CONFIG[assertProvider(provider)];
+  switch (assertSource(source)) {
+    case 'platform':
+      return `${prefix}/${config.parameterName}`;
+    case 'space':
+      return `${prefix}/projects/${assertIdentifier(projectId, 'projectId')}/agent-credentials/${
```
eipasteur, line 88: Space credentials live under this path, but the project teardown (`lambda/projects/index.js:1839`, not part of this diff) drops the Project vertex without deleting either SecureString, so a deleted space keeps its Bedrock/Kiro keys indefinitely, and they stay inside the broker's `projects/*/agent-credentials/*` read scope. Could we delete both provider parameters before dropping the vertex (idempotent, `ParameterNotFound` tolerated) and cover it in the deletion tests? Note the projects Lambda runs as `neptune_artifacts`, which only has SSM rights on `projects/*/mcp-secrets*`, so this needs a scoped `ssm:DeleteParameter` in Terraform as well. Project-tier MCP secrets seem to have the same gap today; fine to keep that out of this PR, but it might be worth handling in the same pass.

lambda/discussions/handlers.js
```diff
@@ -582,6 +590,59 @@ export const assistDiscussion = async (event, res) => {
...
+            code: 'agent_credential_required',
+          });
+        }
+        return {
+          requestedCli: execution.agentCli,
+          credentialBinding: execution.credentialBinding ?? { provider, source: 'platform' },
+        };
+      }
+    }
+    if (!credentialProvider) {
```
eipasteur, line 638: For a DRAFT assist, omitting `agentCli` falls through here to `{ requestedCli: null, credentialBinding: null }` instead of a 400. The handler then persists and broadcasts a running Quorum message, invokes AgentCore with no `requestedCli`/binding/grant, and returns 202; AgentCore's one-shot later hits `no_cli` and updates that message to a generic failed state. Since intents already enforce `agent_cli_required` for DRAFT start and compose, it feels like the assist endpoint should do the same before creating the pending message. It's also reachable from the UI: `useDiscussion` omits `agentCli` when `assistAgentCli` is null, which happens after the compose-page desync noted on `IntentComposePage.tsx:138`. The started-intent path keeps working as is (pinned CLI/binding win over the body, already covered by the existing test).

eipasteur (changes requested):
Reviewed at ef6a6ab1. Seven inline comments below: three on the credential trust boundary (MCP env inheritance, decryption scope, start race), two frontend state-reset issues, one cleanup gap on space deletion, one missing validation on DRAFT assists.

lambda/agentcore/auth-resolver.js
JWThewes: Fixed in c020478. Custom stdio MCP entries now explicitly override AWS_BEARER_TOKEN_BEDROCK and KIRO_API_KEY with empty values after authored env is applied, so Claude, Kiro, and OpenCode children cannot inherit (or restore) the selected user credential. Remote servers are unchanged and the trusted aidlc bridge keeps only its scoped env; Codex retains its existing sanitized allowlist behavior. Added a regression test that merges a personal-token parent env with both passive and hostile custom-server configs and verifies neither credential is observable. Focused tests (74) and the full AgentCore suite (787) pass.

lambda/shared/agent-credentials.js
JWThewes: Fixed in 5c2963b. The agents, intents, and discussions roles no longer have any SSM read permission on platform/user/space agent credential paths. Agents retains write/delete only for the management APIs. Set-state and effective-source lookup now go through a metadata-only broker Lambda that shares the dedicated credential_broker IAM role (the sole credential-path reader), returns only booleans/normalized bindings, and explicitly rejects the value-redemption action. The existing value broker remains invokable only by AgentCore, so the grant-signing Lambdas cannot use the metadata route to redeem values. This also preserves existing platform parameters/placeholder semantics without a migration. Added metadata allowlist/no-value regressions and caller tests; all 2,540 repository tests pass, affected packages build, and both Terraform modules validate.

lambda/intents/index.js
JWThewes: Good catch — this was a real read-after-write race. Fixed in fbb8c80. `getExecution` now supports an opt-in strongly consistent read, and both the orchestrator’s initial pinned-META load and AgentCore’s credential-grant verification use it; ordinary process reads remain eventually consistent. I also added stale-read regressions that return the old DRAFT image unless `consistentRead: true` is requested and verify that both sides use the freshly pinned user binding. Validation: 200 focused tests and the full 2,543-test suite pass.

frontend/src/components/settings/AgentCredentialScopeCard.tsx
JWThewes: Agreed — this was a valid cross-space state leak. Fixed in 9bf25e7. The Space card is now keyed by project ID, and the card itself resets write-only inputs plus loading/saving/clearing/result/error state whenever its scope identity changes. Async load, save, clear, retry, and delayed result timers are also identity-guarded, so a late Space A completion cannot update Space B. Added regressions for an unsaved A draft, a delayed A load after switching to B, and an in-flight A save during the switch. Validation: all 71 frontend test files / 473 tests pass, along with frontend build, typecheck, lint, and formatting.

frontend/src/pages/IntentComposePage.tsx
JWThewes: Agreed — this is a real same-route history-navigation bug. Fixed in ecddae6 by keying the compose-page content on both `projectId` and `intentId`. Moving from draft A to draft B now synchronously remounts the per-intent state: A’s CLI selection is gone, Start/Compose require a fresh explicit selection, and the old page’s discussion-assist cleanup sends `null` before B is used. Added a same-space A→B route-switch regression that verifies the CLI is unselected, Start is disabled, and `assistAgentCli` is cleared. Validation: all 71 frontend test files / 474 tests pass, plus build, typecheck, formatting, and lint.

lambda/shared/agent-credentials.js
JWThewes: Agreed — the new Space credential scope must be part of project teardown. Fixed in 6d30006. Project deletion now idempotently deletes both Bedrock and Kiro SecureStrings after all child intents are purged and before any project graph teardown; `ParameterNotFound` is tolerated, while any other SSM failure returns 500 with the Project vertex still present for retry. The Projects role received a separate delete-only `ssm:DeleteParameter` grant on `projects/*/agent-credentials/*`—no credential read or rotate access. Tests cover both parameter deletion and the retryable failure ordering. Validation: 124 focused tests, full 2,546-test backend suite, Projects bundle, Terraform fmt/validate, lint, and formatting all pass. I kept the pre-existing project MCP-secret cleanup gap separate from this PR-focused fix.

lambda/discussions/handlers.js
JWThewes: Agreed — fixed in aaaa7b4. DRAFT assists now perform a strongly consistent META preflight before retry mutation, lock acquisition, message creation/broadcast, or AgentCore invocation. If neither a body CLI nor a non-DRAFT pinned CLI exists, the endpoint returns 400 with `agent_cli_required`. DRAFT META is deliberately treated as unpinned, while started intents continue to use their pinned CLI/binding even when the body omits `agentCli`; successful duplicate request IDs remain idempotent. The regression verifies no Quorum message, metadata-broker call, or AgentCore call is created on rejection. Validation: 78 focused tests, full 2,547-test backend suite, Discussions bundle, lint, and formatting pass.

JWThewes:
@eipasteur Thanks for taking a look - I addressed your feedback :)

eipasteur (approved):

Issue #398, opened 2026-08-11
[Feature]: Offer optional advanced security features


jeromevdl:
### Description

Optional security features to add in terraform:
- WAF (applied on CloudFront distribution / API Gateway / Cognito) with a set of rules (customer can complete)
- All Lambda functions in VPC + endpoints
- Customer KMS key to encrypt everything (S3, DDB, Neptune)

### Use case

Enterprise customers might want additional level of security when deploying the solution, especially a WAF to protect the application / API / Cognito.

### Area

Infrastructure (Terraform)

PR #410, opened 2026-08-19
feat(networking): route provider authentication through static NAT egress

Alae-Tamouh:
*Issue #, if available:*
Relates partially to https://github.com/aws-samples/sample-collaborative-ai-dlc/issues/398

*Description of changes:*
Related to [#398](https://github.com/aws-samples/sample-collaborative-ai-dlc/issues/398), which proposes optional advanced security features including VPC placement for all Lambda functions. This PR implements the narrower static provider-egress use case as we require stable egress IP addresses for allow-listing on GitHub organizations and other providers.

This change:
- attaches the GitHub, GitLab, Bitbucket, and credential-broker Lambdas to the existing private subnets and Lambda security group;
- grants their custom execution roles the permissions required for VPC access;
- routes their outbound traffic through the existing NAT gateways;
- exposes the NAT gateway public IP addresses as a Terraform output for allow-listing.

This introduces a NAT dependency for these Lambdas, with corresponding NAT cost
and potential cold-start latency. It does not attempt the broader all-Lambda VPC
placement proposed by [#398](https://github.com/aws-samples/sample-collaborative-ai-dlc/issues/398).

This contribution can be treated independently or closed in favor of a broader implementation of [#398](https://github.com/aws-samples/sample-collaborative-ai-dlc/issues/398) if that direction is preferred.

### DIFF
--- a/terraform/modules/api/lambda/variables.tf
+++ b/terraform/modules/api/lambda/variables.tf
@@ -23,6 +23,11 @@ variable "private_subnet_ids" {
   type        = list(string)
 }
 
+variable "lambda_vpc_scope" {
+  description = "Lambda VPC placement scope"
+  type        = string
+}
+
 variable "neptune_endpoint" {
   description = "Neptune cluster endpoint"
   type        = string
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Unreleased]
 
+### Added
+
+- Optional static egress for OAuth connectors, credential resolution, and seed-blocks through `lambda_vpc_scope = "public-egress"`, with NAT public IP outputs and addresses printed in the deployment summary for external allow-lists.
+
 ## [2.0.0] - 2026-08-06
 
 Second and final step of the v2 release, building on `2.0.0-preview0`. Everything listed below is new since that preview; see the `2.0.0-preview0` entry for the v2 platform itself.
--- a/docs/getting-started/setup.md
+++ b/docs/getting-started/setup.md
@@ -282,6 +282,7 @@ Bootstrap creates `terraform/environments/dev.s3.tfbackend`. The environment arg
 ./scripts/bootstrap.sh dev
 cp terraform/environments/dev.tfvars.example terraform/environments/dev.tfvars
 # Set aws_region = "<aws-region>" in dev.tfvars.
+# For static provider egress IPs, set lambda_vpc_scope = "public-egress".
 # For a custom domain, also set app_domain plus either acm_certificate_arn or
 # route53_zone_id — see "Custom domain → Without the installer" above.
 ./scripts/deploy-terraform.sh dev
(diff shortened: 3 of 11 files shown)


jeromevdl:
Thank you, @Alae-Tamouh, for your contribution. This clearly adds value for enterprises with network allowlisting requirements.

We plan to address the broader scope of #398 later, but this is a good first step. However, I would prefer this behavior to be optional, since routing external traffic through the NAT gateway may incur additional costs.

Could we introduce a `lambda_vpc_scope` variable with these values?

- `required` (default): Only Lambdas requiring access to private VPC resources, currently primarily Neptune.
- `public-egress`: `required` plus all Lambdas accessing non-AWS public services, such as GitHub, GitLab, Bitbucket, and Jira. Ideally, this should also include `seed-blocks`, since it accesses codeload.github.com, even though that endpoint does not require IP allowlisting.
- `all`: To be introduced later as part of #398, placing every Lambda in the VPC.

If you don't have the bandwidth to address this, let me know and we'll take over. Thanks again.

Alae-Tamouh:
Thanks for the suggestion @jeromevdl, it makes sense to have this feature as an option, as not all users are gated behind enterprise allow-lists.
I implemented the `lambda_vpc_scope` with the two values as suggested.
I will let you handle the `all` value and additional security features in the dedicated feature https://github.com/aws-samples/sample-collaborative-ai-dlc/issues/398.

terraform/modules/api/lambda/variables.tf
```diff
@@ -23,6 +23,16 @@ variable "private_subnet_ids" {
   type        = list(string)
 }
 
+variable "lambda_vpc_scope" {
+  description = "Lambda VPC placement scope"
+  type        = string
+
+  validation {
```
jeromevdl: not sure we need the validation again here, as it is already provided in root

jeromevdl (dismissed):
@Alae-Tamouh thanks for adding the option, one minor remark if you want to remove the duplicated variables check

terraform/modules/api/lambda/variables.tf
Alae-Tamouh: Fixed

jeromevdl (dismissed):

JWThewes (commented):
@Alae-Tamouh thanks, this is a useful addition and the gating is clean. With `required` the plan is identical to before, so there is no risk for existing deployments. Three things I would like to see before we merge, all operator-facing rather than code:

**1. Docs.** `lambda_vpc_scope` currently only exists in Terraform — no hits in `docs/` or the README, and no doc mentions NAT or egress allow-listing at all today. Could you add a short section under "Operator setup (one time per deployment)" in `docs/using-the-platform/git-integration.md`, plus a note in the tfvars part of `docs/getting-started/setup.md`?

Two points worth stating explicitly there:

- The number of addresses to allow-list depends on the environment. `aws_eip.nat` is `count = var.environment == "prod" ? 2 : 1` (`terraform/modules/networking/main.tf:60`), so a dev deployment exposes one IP and prod two. Someone who allow-lists a dev deployment and later promotes to prod would otherwise get intermittent provider 403s from the second, unlisted address.
- What the flag does *not* change. `source-control` and `trackers` are already unconditionally VPC-attached (`terraform/modules/api/lambda/main.tf:843-844` and `:1638-1639`), so all repository-provider and Jira traffic already leaves through the NAT, as does the AgentCore runtime (`network_mode` defaults to `VPC`). This flag is specifically about the OAuth connectors, the credential broker and seed-blocks. The comment in `dev.tfvars.example` reads as if provider traffic is not NAT-routed by default, which is not the case.

**2. Changelog.** Please add an `### Added` entry under `## [Unreleased]`. This is a user-facing deployment option plus a new output, the same category as the custom-domain and enterprise-SSO variables that both landed with their entries.

**3. Print the addresses.** `nat_egress_public_ips` is only reachable via `terraform output` right now. Everything else in this repo that an operator has to paste into a third-party system is printed by `print_deployment_summary` in `scripts/deploy-terraform.sh` — the OIDC callback, SAML ACS and entity ID at lines 154-160, and the external DNS records via `print_dns_instructions`. Could you emit the IPs there when `lambda_vpc_scope` is `public-egress`? Allow-listing is the whole point of the change, so the values should land in front of the operator at apply time.

Alae-Tamouh:
@JWThewes Added the requested changes to the doc, changelog, and printing out nat_egress_public_ips.

JWThewes (approved):
Looks good to me

PR #411, opened 2026-08-19
fix(frontend): discover repositories through the bound auth method

Alae-Tamouh:
*Issue #, if available:* None

*Description of changes:*
## Problem
The Add Repositories picker always uses personal OAuth discovery. A space bound through the platform GitHub App can therefore fail to list repositories when the current user has no personal GitHub OAuth connection.
image
image

## Reproducer
1. Configure the platform GitHub App.
2. Create a space containing a GitHub repository and bind its source control using `GitHub App`.
3. Sign in as a space owner or administrator who has not connected a personal GitHub OAuth identity.
4. Open **Space settings → Source control**.
5. Select **Add** under Repositories.

The picker calls the personal OAuth repository endpoint (`/github/repos`) and fails with `OAUTH_NOT_CONFIGURED` or a GitHub connection error. No repositories from the space's GitHub App installation are shown.

### Expected behavior
Because the space is bound through `github-app`, the picker should use the GitHub App repository endpoint (`/github/app/repos`) and list repositories available to the installation without requiring personal OAuth.

## Solution
Propagate the source-control status already loaded by the binding section and select GitHub App discovery only for a confirmed `github-app` binding. All other bindings retain OAuth discovery. Status is propagated after rebinding so
the picker updates without a page reload.

## Impact
GitHub App-bound spaces can add repositories without requiring personal OAuth. Existing OAuth, GitLab, and Bitbucket behavior is unchanged.

### DIFF
--- a/frontend/src/components/project-settings/RepositoriesTab.test.tsx
+++ b/frontend/src/components/project-settings/RepositoriesTab.test.tsx
@@ -1,7 +1,9 @@
 import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import { describe, expect, it, vi, beforeEach } from 'vitest';
 import type { Project } from '@/services/projects';
-import type { ProjectSourceControlStatus } from '@/services/sourceControl';
+import type { ProjectSourceControlStatus, SourceControlAuthType } from '@/services/sourceControl';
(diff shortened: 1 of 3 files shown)


jeromevdl (approved):
Good catch @Alae-Tamouh, thanks for the fix.

PR #412, opened 2026-08-19
fix(trackers): hide OAuth reconnect for GitHub App bindings in Trackers tab

Alae-Tamouh:
*Issue #, if available:* None

*Description of changes:*
## Problem
The project tracker settings show a Reconnect action for every non-legacy Git-backed tracker. For GitHub Issues backed by a GitHub App project binding, that action starts personal GitHub OAuth even though the binding has no personal OAuth identity to refresh. On App-only deployments it can fail with OAUTH_NOT_CONFIGURED.

## Reproducer
1. Configure the platform with a GitHub App.
2. Create a space whose GitHub repository is bound using `github-app`.
3. Open **Space settings → Trackers**.
4. Observe that the GitHub Issues tracker displays **Reconnect**.
5. Select **Reconnect**.

The UI starts a personal GitHub OAuth flow even though the tracker authenticates
through the space's GitHub App binding. That flow cannot repair or refresh the
App binding. If no GitHub OAuth app is configured, it also fails with
`OAUTH_NOT_CONFIGURED`.

## Solution
Load the project's source-control binding status and hide Reconnect only when the matching GitHub repository is authenticated through `github-app`. Reconnect remains available for GitHub OAuth, GitLab OAuth, Jira, and when binding status cannot be loaded.

## Impact
Users no longer see a recovery action that cannot repair GitHub App-backed trackers. OAuth-backed tracker recovery is unchanged.

### DIFF
--- a/frontend/src/components/project-settings/TrackersTab.tsx
+++ b/frontend/src/components/project-settings/TrackersTab.tsx
@@ -13,6 +13,7 @@ import {
   trackerIdForGitProvider,
   type GitTrackerProviderId,
 } from '@/services/gitProvider';
+import { sourceControlService, type ProjectSourceControlStatus } from '@/services/sourceControl';
 import { getTrackerProvider, TRACKER_PROVIDERS } from '@/lib/trackerProviders';
 import { useTrackerProviders } from '@/hooks/useTrackerProviders';
 import { MigrateTrackerCard } from '@/components/MigrateTrackerCard';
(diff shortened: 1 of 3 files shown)


frontend/src/components/project-settings/TrackersTab.tsx
```diff
@@ -53,8 +55,24 @@ export function TrackersTab({ project, canEdit, reload }: Props) {
       .catch(() => setTrackerConnections([]));
   }, []);
 
+  useEffect(() => {
```
jeromevdl, line 65: Non-blocking: Could we only fetch source-control status when `canEdit` is true and the project has a non-legacy github-issues binding? 
`canReconnect()` only consumes this data for GitHub Issues, so the current unconditional effect adds an unused request for GitLab, Jira-only, read-only, and tracker-free views.

jeromevdl (changes requested):
Thank you @Alae-Tamouh for this. 

One small improvement required (see my comment) before approving it.

frontend/src/components/project-settings/TrackersTab.tsx
Alae-Tamouh: Well-spotted, thanks. I changed it.

jeromevdl (approved):

PR #413, opened 2026-08-19
fix(frontend): notify subscribers after persisted cache hydration

Alae-Tamouh:
*Issue #, if available:* None

*Description of changes:*
## Problem
Projects and sprints are hydrated from sessionStorage inside mount effects. Subscribers render before hydration and initially observe an empty external store. When the persisted entry is fresh, network revalidation returns early, so no version change occurs and mounted views can remain empty indefinitely.
image

## Solution
Notify the existing project and per-project sprint external-store subscribers immediately after adopting persisted cache entries.

## Impact
Dashboard, space, and sprint views rerender with fresh persisted data without requiring an unnecessary network request. Cache TTL and refresh behavior are unchanged.

### DIFF
--- a/frontend/src/hooks/useProjectsCache.test.ts
+++ b/frontend/src/hooks/useProjectsCache.test.ts
@@ -1,6 +1,25 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import { renderHook, waitFor } from '@testing-library/react';
 import { deriveActivity } from './useProjectsCache';
 import type { Intent, IntentStatus } from '@/services/intents';
+import type { Sprint } from '@/services/sprints';
+
+const cacheServiceMocks = vi.hoisted(() => ({
+  listProjects: vi.fn(() => Promise.reject(new Error('unexpected projects fetch'))),
+  listSprints: vi.fn(() => Promise.reject(new Error('unexpected sprints fetch'))),
+}));
+
+vi.mock('@/services/projects', () => ({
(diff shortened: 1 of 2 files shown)


frontend/src/hooks/useProjectsCache.test.ts
```diff
@@ -42,3 +56,50 @@ describe('deriveActivity — dashboard per-project counts', () => {
...
+          },
+        ],
+        fetchedAt: Date.now(),
+      }),
+    );
+
+    const { useProjectsCache } = await import('./useProjectsCache');
+    const { result } = renderHook(() => useProjectsCache());
+
+    await waitFor(() => expect(result.current.projects).toHaveLength(1));
```
jeromevdl, line 92: Minor: Could we explicitly assert that `projectsService.list` was not called after hydrating a fresh entry? The rejected mock does not enforce this because `revalidateProjects()` catches the failure while retaining the persisted cache. The sprint test should likewise assert that `sprintsService.list` was not called.

jeromevdl (changes requested):
Thanks @Alae-Tamouh for this fix. One minor test improvement before approving.

frontend/src/hooks/useProjectsCache.test.ts
Alae-Tamouh: Good catch. I added `not.toHaveBeenCalled()` assertions for both `projectsService.list` and `sprintsService.list`.

jeromevdl (approved):

PR #414, opened 2026-08-19
fix(auth): sign out only after definitive session expiry

Alae-Tamouh:
*Issue #, if available:* None

*Description of changes:*
## Problem
When Cognito can no longer refresh a session, API calls fail with 401 but AuthContext retains the previous user. The UI remains authenticated and continues issuing failing requests.
Logging out on every API 401 is unsafe because provider endpoints can also return GitHub, GitLab, Bitbucket, or Jira 401 responses while the Cognito session remains valid.

image
image

## Reproducer
1. Sign in and open any authenticated page.
2. In another browser tab or through the Cognito administration API, revoke the user's refresh token or invalidate the session.
3. Return to the application after the access token expires and trigger an API request, such as refreshing the spaces list.
4. Observe that the request fails with 401, but the application still considers the previous user authenticated and remains on the protected page.
5. Trigger additional actions and observe repeated 401 responses instead of a return to the login page.

## Solution
Classify session resolution as valid, definitively expired, or transiently unavailable. On an API 401, force a Cognito refresh before notifying AuthContext. A small epoch-aware bridge coalesces expiry notifications and prevents late requests from an old session from signing out a newly authenticated local or SSO session.

## Impact
Expired users return cleanly to login. Network failures and provider-specific 401 responses no longer cause false logout, and stale requests cannot expire a replacement session.

### DIFF
--- a/frontend/src/contexts/AuthContext.test.tsx
+++ b/frontend/src/contexts/AuthContext.test.tsx
@@ -4,7 +4,11 @@ import { beforeEach, describe, expect, it, vi } from 'vitest';
 
 const mocks = vi.hoisted(() => ({
   isAuthenticated: vi.fn(),
+  getCurrentUser: vi.fn(),
+  logout: vi.fn(),
   loginWithSso: vi.fn(),
+  completeSsoLogin: vi.fn(),
+  consumeReturnTo: vi.fn(),
 }));
 
 vi.mock('../services/auth', () => ({
@@ -12,13 +16,19 @@ vi.mock('../services/auth', () => ({
 }));
 
 import { AuthProvider, useAuth } from './AuthContext';
+import {
+  currentSessionEpoch,
+  notifySessionExpired,
+  resetSessionExpiry,
+} from '../services/sessionExpiry';
 
 function SsoLoginHarness() {
-  const { isLoading, loginWithSso } = useAuth();
+  const { user, isLoading, loginWithSso, completeSsoLogin } = useAuth();
 
   return (
     <>
       <span>{isLoading ? 'loading' : 'idle'}</span>
+      <span>{user?.username ?? 'anonymous'}</span>
       <button
         type="button"
         onClick={() => {
@@ -27,13 +37,24 @@ function SsoLoginHarness() {
       >
         Start SSO
       </button>
+      <button
+        type="button"
+        onClick={() => {
+          void completeSsoLogin();
+        }}
+      >
+        Complete SSO
+      </button>
     </>
   );
 }
 
 beforeEach(() => {
   vi.clearAllMocks();
+  resetSessionExpiry();
   mocks.isAuthenticated.mockResolvedValue(false);
+  mocks.logout.mockResolvedValue(undefined);
+  mocks.consumeReturnTo.mockReturnValue('/dashboard');
 });
 
 describe('AuthProvider SSO state', () => {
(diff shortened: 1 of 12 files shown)


frontend/src/services/api.ts
```diff
@@ -33,6 +46,17 @@ async function fetchWithAuth(path: string, options: RequestInit = {}): Promise<R
   });
 
   if (!response.ok) {
+    // A 401 is not proof the Cognito session is dead: endpoints forward
+    // upstream provider statuses verbatim, so a revoked GitHub/GitLab/Jira
+    // token surfaces as a 401 too. Only sign out when a forced refresh shows
+    // the session itself is unusable. (403 is authorization — admin gates —
+    // and never means expiry.)
+    if (response.status === 401 && !isSessionExpiryNotified()) {
```
jeromevdl: Could we ensure simultaneous 401 responses share a single Cognito refresh request? Currently, each request may independently call `resolveSession({ forceRefresh: true })`, creating redundant refresh calls. 
This currently creates redundant Cognito calls and could become problematic if refresh-token rotation is enabled later.

frontend/src/services/auth.ts
```diff
@@ -299,23 +326,34 @@ export const authService = {
...
+      // and rethrows every transient failure, so this is always definitive.
+      return { session: null, expired: true };
     } catch (error) {
       console.error('Get session error:', error);
-      return null;
+      return { session: null, expired: isSessionRefusedError(error) };
     }
   },
 
+  async getSession(): Promise<AuthSession | null> {
```
jeromevdl: When the Cognito session has definitively expired, WebSocket and Yjs connections receive null from `getSession()` and keep retrying instead of logging the user out. The logout only happens after a REST request detects the expiry. Should realtime session checks also notify AuthContext when expiry is definitive?

frontend/src/contexts/AuthContext.test.tsx
```diff
@@ -62,4 +76,25 @@ describe('AuthProvider SSO state', () => {
     });
     expect(await screen.findByText('idle')).toBeInTheDocument();
   });
+
+  it('starts a new session epoch after successful SSO completion', async () => {
```
jeromevdl: Could we add a test that starts authenticated, calls `notifySessionExpired()`, and verifies that the user is cleared and `authService.logout()` is called once? The bridge and API are tested separately, but the actual UI authentication transition is not.

jeromevdl (changes requested):
Great finding @Alae-Tamouh. I never faced this so thank you for your PR.
A few comments before approving.

frontend/src/services/api.ts
Alae-Tamouh: It seems this concern is already addressed by the current amplify implementation used: https://github.com/aws-amplify/amplify-js/blob/3c7a96f9aa3d19c4f84dfee4a88a458e5438cbb5/packages/auth/src/providers/cognito/utils/refreshAuthTokens.ts#L74-L75
If it's the case, no need to reimplement it.

frontend/src/services/auth.ts
Alae-Tamouh: They should, I'll address it shortly

frontend/src/contexts/AuthContext.test.tsx
Alae-Tamouh: Agreed, Ill add it

JWThewes:
@Alae-Tamouh are you going to work on Jeromes comments? Otherwise I can work on this one. Super great finding an I'm eager to get this one in for our next release.

Alae-Tamouh:
@JWThewes Thanks for the reminder, I addressed Jerome's comments

jeromevdl (approved):

PR #416, opened 2026-08-21
chore: Add backend coverage and mutation testing harness

jeromevdl:
## Summary

  - measure backend coverage in CI and retain detailed reports
  - compare PR coverage with the latest successful `main` baseline
  - add [Stryker](https://stryker-mutator.io/docs/stryker-js/getting-started/) mutation testing across backend domains
  - run an informational mutation pilot on changed backend files

## Mutation PR pilot

  - mutates entire changed production files
  - ignores static mutants
  - skips draft PRs
  - cancels obsolete runs after new pushes
  - stops mutation execution after 30 minutes
  - publishes partial results after timeout
  - never blocks a PR because of surviving mutants
  - exposes HTML/JSON reports as workflow artifacts

## Validation

  - 2,498 backend tests passing
  - 601 `shared` tests passing
  - `sso-token`: 23 mutants, 78.26% mutation score
  - lint, formatting, secret scanning and package audits passing

## Follow-up verification

This PR adds the coverage reporting workflow, but GitHub only runs a `workflow_run` workflow after it exists on `main`.

After this PR is merged, the next PR modifying backend code must confirm that:

  - the coverage comment is posted on the correct PR;
  - it lists the backend files modified by that PR;
  - it compares their coverage against the latest successful run on `main`.


stryker.config.mjs
```diff
@@ -0,0 +1,62 @@
...
+    '.claude/**',
+    '.codegraph/**',
+    '.idea/**',
+    '.kiro/**',
+    '.opencode/**',
+    '.venv/**',
+    '.worktrees/**',
+    'frontend/**',
+    'site/**',
+  ],
```
JWThewes, line 43: **Blocker.** Stryker doesn't read `.gitignore` — per the shipped schema the only always-ignored paths are `node_modules`, `.git`, `/reports`, `*.tsbuildinfo`, `/stryker.log`, `.stryker-tmp`. So `terraform/` gets copied into the sandbox on every scope run. On my checkout that's **2.7 GB** (1.0 GB of `.terraform` provider binaries), ×29 scopes for a full run.

Also `.codegraph`, `.kiro`, `.opencode`, `.worktrees` don't exist in this repo, while the 2.7 GB dir that does isn't listed — the denylist is going to rot. Can we invert it?

```js
ignorePatterns: ['**', '!lambda/**', '!test/**', '!package.json', '!package-lock.json', '!vitest*.config.js']
```

Missing today beyond terraform: `docs/` (3.5 MB), `lambda/*/.build/`, `.husky/`.

stryker.config.mjs
```diff
@@ -0,0 +1,62 @@
...
+  .split(',')
+  .filter(Boolean);
+const explicitMutate = process.env.STRYKER_MUTATE_FILES
+  ? JSON.parse(process.env.STRYKER_MUTATE_FILES)
+  : null;
+const mutate =
+  explicitMutate ??
+  (scope === 'untested'
+    ? untestedDirectories.map((directory) => `lambda/${directory}/**/*.js`)
+    : [`lambda/${scope}/**/*.js`, `!lambda/${scope}/test/**`]);
```
JWThewes, line 16: **Blocker.** `lambda/${scope}/**/*.js` doesn't exclude `.build/`, which terraform populates with bundled Lambda output (`.gitignore:55`, and it's in `.oxlintignore` too). Combined with the `ignorePatterns` gap above, anyone who has run terraform and then `npm run test:mutation` will mutate minified bundles.

`run-changed-backend-mutations.mjs` already filters `/.build/` — can we apply the same exclusion here?

stryker.config.mjs
```diff
@@ -0,0 +1,62 @@
...
+const reportBase = process.env.STRYKER_REPORT_BASE ?? `reports/mutation/${scope}`;
+const incremental = process.env.STRYKER_INCREMENTAL !== 'false';
```
JWThewes, line 18: **Blocker — please drop `incremental`.** `reports/` is gitignored and not cached in CI, so this buys us nothing there, but it makes local scores non-reproducible. `run-backend-mutations.mjs:65-66` deletes `${scope}.json` and `.html` before each run but not `${scope}-incremental.json`.

Reproduced on `building-blocks`:

```
--scope=building-blocks                    -> 66.78%
--scope=building-blocks --include-static   -> 62.42%
--scope=building-blocks                    -> 62.42%   <- should be 66.78
   INFO IncrementalDiffer  Result: 330 of 330 mutant result(s) are reused.
```

So `--include-static` is one-way sticky. Either drop `incremental` entirely or `rmSync` the incremental file alongside the other two.

stryker.config.mjs
```diff
@@ -0,0 +1,62 @@
...
+export default {
+  $schema: './node_modules/@stryker-mutator/core/schema/stryker-schema.json',
+  testRunner: 'vitest',
+  vitest: {
+    configFile: 'vitest.mutation.config.js',
+    related: scope !== 'untested',
+  },
+  mutate,
+  coverageAnalysis: 'perTest',
+  concurrency: 4,
```
JWThewes, line 29: Nit: hardcoded. `os.availableParallelism() - 1`? Related: on a 4-vCPU runner this means 4 test-runner processes, which multiplies the testcontainer count for the infra scopes (see my note on `vitest.mutation.config.js`).

vitest.mutation.config.js
```diff
@@ -0,0 +1,25 @@
...
+
+const scope = process.env.STRYKER_SCOPE;
+if (!scope) {
+  throw new Error('STRYKER_SCOPE is required; run mutation tests through the npm scripts');
+}
+
+const selectedProjects =
+  scope === 'untested'
+    ? backendProjects
+    : backendProjects.filter((project) => project.name === scope);
```
JWThewes, line 12: Suggest dropping the `untested` scope. Selecting all 28 projects forces `globalSetup` + Docker. I measured **902 mutants across 7 files** — and those 7 are standalone API handlers nothing else imports, so every mutant comes back `NoCoverage`. Coverage already tells us that for free (`all: true`, `include: lambda/**/*.js`).

Same cost lands in the PR path: touching `lambda/users/index.js` (326 lines) spins up the full suite just to conclude "still untested".

vitest.mutation.config.js
```diff
@@ -0,0 +1,25 @@
...
+if (selectedProjects.length === 0) {
+  throw new Error(`No Vitest project found for mutation scope "${scope}"`);
+}
+
+export default defineConfig({
+  test: {
+    projects: selectedProjects.map((project) => project.config),
+    globalSetup: selectedProjects.some((project) => project.requiresInfrastructure)
+      ? backendGlobalSetup
+      : undefined,
```
JWThewes, line 23: Needs validation before merge. `@stryker-mutator/vitest-runner` calls `createVitest()` per test-runner process, and Stryker logs `Creating 4 test runner process(es)`. For any of the 13 infra scopes that's **4 gremlin-server JVMs + 4 DynamoDB-Local JVMs** on a 4-vCPU runner.

The validation section only covers `sso-token` (23 mutants, no containers), so this path is completely unexercised. Can we get one infra scope through CI before merging?

.github/workflows/mutation-pr.yml
```diff
@@ -0,0 +1,70 @@
...
+      - name: Finalize mutation summary
+        if: always()
+        env:
+          MUTATION_EXECUTION_INCOMPLETE: ${{ steps.mutation.outcome != 'success' }}
+          MUTATION_WORKFLOW_URL: >-
+            ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: >-
+          npm run test:mutation:changed --
+          --report-only
+          --base=${{ github.event.pull_request.base.sha }}
```
JWThewes, line 61: **Blocker.** `--report-only` skips the execution loop, so `failures` is always empty (`run-changed-backend-mutations.mjs:89`). If a scope actually *failed* rather than timing out, the first pass writes `Mutation execution failed for: \`scope\``, then this step overwrites it with:

> The 30-minute pilot limit was reached or execution did not complete. Results above are partial.

Wrong cause, and the failing scope is lost. Reproduced locally with `MUTATION_EXECUTION_INCOMPLETE=true ... --report-only`.

`MUTATION_EXECUTION_INCOMPLETE: ${{ steps.mutation.outcome != 'success' }}` conflates "timed out" with "exited non-zero". Either gate on `outcome == 'cancelled'`, or persist `failures` into `summary.json` and re-read it in `--report-only`.

.github/workflows/mutation-pr.yml
```diff
@@ -0,0 +1,70 @@
...
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+    paths:
+      - lambda/**/*.js
+      - '!lambda/**/test/**'
+      - '!lambda/**/.build/**'
```
JWThewes, line 13: `paths` is missing the harness's own files, so changes to `stryker.config.mjs`, `vitest.mutation.config.js`, either script, or this workflow never trigger it. `terraform.yml:10,14` and `frontend.yml` both include their own workflow file plus deps — let's match that.

Two smaller things while we're here:
- quoting is inconsistent (`lambda/**/*.js` unquoted vs `'!lambda/**/test/**'` quoted)
- no `workflow_dispatch:` — every other workflow in this repo has one, and I'd like to be able to kick off a mutation run manually

.github/workflows/mutation-report.yml
```diff
@@ -0,0 +1,59 @@
...
+  workflow_run:
+    workflows:
+      - Backend Mutation PR
+    types:
+      - completed
+
+permissions:
+  actions: read
+  contents: read
+  pull-requests: write
```
JWThewes, line 13: Two things on this workflow:

1. No `concurrency` block, while `coverage-report.yml` has one. Two quick pushes can race the sticky comment.
2. `summary.md` is generated entirely on the untrusted runner and posted here by a token with `pull-requests: write` — a fork can make the bot post arbitrary markdown. Impact is comment content only, but it's the standard `workflow_run` footgun and worth escaping.

.github/workflows/coverage-report.yml
```diff
@@ -0,0 +1,76 @@
...
+        if: steps.baseline.outputs.found_artifact != 'true'
+        uses: davelosert/vitest-coverage-report-action@8b157684c6a6b259b97d45e72b44242865c0f6a5 # v2
+        with:
+          name: Backend
+          github-token: ${{ github.token }}
+          comment-on: pr
+          json-summary-path: coverage/coverage-summary.json
+          json-final-path: coverage/coverage-final.json
+          file-coverage-mode: changes
+          sort-by: lines-asc
```
JWThewes, line 76: Nit: two near-identical blocks. Collapse into one step —

```yaml
json-summary-compare-path: ${{ steps.baseline.outputs.found_artifact == 'true' && 'coverage-main/coverage-summary.json' || '' }}
```

.github/workflows/coverage-report.yml
```diff
@@ -0,0 +1,76 @@
...
+jobs:
+  report:
+    if: >-
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    steps:
+      # Keep the privileged workflow on trusted default-branch code. The action
+      # documents workflow_run support, but changed-file rows still require an
+      # end-to-end check on the first PR after this workflow reaches main.
```
JWThewes, line 29: We can drop the hedging here (and the matching paragraph in the README) — it's verifiable from the action source rather than needing an end-to-end check. `getPullRequestNumber.ts` handles `workflow_run` including the fork `head_sha` fallback, and `getPullChanges.ts` resolves changed files via `pulls.listFiles`. I read through v2.12.2 at the pinned SHA.

Keeping the first sentence about staying on default-branch code is still worthwhile.

.github/workflows/test.yml
```diff
@@ -34,7 +34,33 @@ jobs:
...
+        if: matrix.node-version == 24
+        run: npm run test:coverage
+
+      - name: Upload backend coverage report
+        if: matrix.node-version == 24
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: backend-coverage-node-24
+          path: coverage/
+          if-no-files-found: warn
```
JWThewes, line 50: Should be `error`, like the baseline upload right below. As-is, missing coverage keeps this workflow green and fails `coverage-report.yml` instead — and `workflow_run` failures don't surface as PR checks, so it fails silently.

Worth noting: I tried a run where `globalSetup` blew up, and vitest still wrote an (empty) `coverage-final.json`, so `if-no-files-found` won't catch that case either way.

.github/workflows/test.yml
```diff
@@ -34,7 +34,33 @@ jobs:
         run: npm ci
 
       - name: Run unit tests
+        if: matrix.node-version == 22
         run: npm test
 
+      - name: Run unit tests with coverage
+        if: matrix.node-version == 24
+        run: npm run test:coverage
```
JWThewes, line 42: Flagging rather than blocking: after this split, Node 22 only exercises `npm test` and Node 24 only `--coverage`. Neither leg covers both paths any more, and v8 instrumentation on top of testcontainers is a new flake surface on the 24 leg. Fine if that's a deliberate trade, just want it recorded.

vitest.config.js
```diff
@@ -16,20 +16,46 @@ const globalSetup = [
...
+  'discussions',
+  'intents',
+  'migrate-tracker-fields',
+  'projects',
+  'purge-neptune',
+  'questions',
+  'sprint-graph',
+  'sprints',
+  'tasks',
+  'timeline-events',
```
JWThewes, line 33: I verified all 13 entries are correct today (grepped `NEPTUNE_ENDPOINT` / `DYNAMODB_LOCAL_ENDPOINT` / `GREMLIN_PORT` across `lambda/*/test/`), but nothing keeps them correct.

Either derive the set from that grep, or add a test asserting it's in sync. Otherwise the next lambda with integration tests fails inside a mutation run with a confusing missing-container error.

vitest.config.js
```diff
@@ -16,20 +16,46 @@ const globalSetup = [
...
-        include: ['test/**/*.test.js'],
-        setupFiles,
-      },
-    })),
+    projects: backendProjects.map((project) => project.config),
     globalSetup,
     coverage: {
       provider: 'v8',
-      reporter: ['text', 'html', 'lcov'],
+      reporter: ['text', 'html', 'lcov', 'json-summary', 'json'],
```
JWThewes, line 58: Minor: adding `json` on top of `all: true` over 521 files makes `coverage-final.json` large, and it ships in the same artifact as two HTML reports (`html` writes `coverage/`, `lcov` writes `coverage/lcov-report/`). Fine at 3-day retention, just be aware of the size.

scripts/run-backend-mutations.mjs
```diff
@@ -0,0 +1,129 @@
+import { spawnSync } from 'node:child_process';
```
JWThewes, line 1: **Blocker: no tests.** `scripts/test/` has tests for `release-tools`, `sso-config`, `docker-proxy-build-args` and `agent-e2e-proxy`, all run via `npm run test:release` in CI. These two new scripts add 377 lines with real branching (scope grouping, missing reports, `--report-only`, failure paths) and no tests. Feels off in a PR about test quality.

**Also: the two scripts are ~2/3 duplicated.** Both independently re-implement lambda enumeration, `untestedDirectories`, the `strykerBin` path, the spawn env block, the failure loop and metrics summarisation — and the enumeration is a *third* copy of what's already in `vitest.config.js`. One script with a `--changed`/`--base` flag, or a shared `scripts/lib/mutation.mjs`?

scripts/run-backend-mutations.mjs
```diff
@@ -0,0 +1,129 @@
...
+    return entry.name.endsWith('.js') ? total + statSync(path).size : total;
+  }, 0);
+};
+
+const sourceBytesByScope = new Map(
+  testedScopes.map((scope) => [scope, sourceBytes(join(lambdaRoot, scope))]),
+);
+const orderedTestedScopes = testedScopes.toSorted(
+  (left, right) =>
+    sourceBytesByScope.get(left) - sourceBytesByScope.get(right) || left.localeCompare(right),
```
JWThewes, line 47: Can we drop this? A recursive byte-counting walk plus a custom comparator, just to order scopes by size. `.toSorted()` alphabetically gets us the same practical outcome.

scripts/run-backend-mutations.mjs
```diff
@@ -0,0 +1,129 @@
...
+const availableScopes = [...orderedTestedScopes, 'untested'];
+
+if (requestedScope && !availableScopes.includes(requestedScope)) {
+  console.error(
+    `Unknown mutation scope "${requestedScope}". Expected one of: ${availableScopes.join(', ')}`,
+  );
+  process.exit(2);
+}
+
+const scopes = requestedScope ? [requestedScope] : availableScopes;
```
JWThewes, line 58: `--scope=X` narrows `scopes` to one entry, and the summary at L91 iterates `scopes` — so a single-scope run overwrites `reports/mutation/summary.json` with one row and clobbers the full-run summary. Verified locally.

scripts/run-backend-mutations.mjs
```diff
@@ -0,0 +1,129 @@
...
+      ...process.env,
+      STRYKER_SCOPE: scope,
+      STRYKER_UNTESTED_DIRECTORIES: untestedDirectories.join(','),
+      STRYKER_IGNORE_STATIC: String(!includeStatic),
+    },
+    stdio: 'inherit',
+  });
+  if (result.signal) {
+    console.error(`Mutation scope "${scope}" interrupted by ${result.signal}`);
+    process.exit(1);
```
JWThewes, line 82: `process.exit(1)` here discards every result collected so far — Ctrl-C or an OOM kill mid-run loses the whole summary. `run-changed-backend-mutations.mjs:101-105` handles the same case by pushing to `failures` and continuing. Let's pick one behaviour.

scripts/run-backend-mutations.mjs
```diff
@@ -0,0 +1,129 @@
...
+      },
+    ];
+  });
+  const summaryPath = join(root, 'reports', 'mutation', 'summary.json');
+  mkdirSync(dirname(summaryPath), { recursive: true });
+  writeFileSync(
+    summaryPath,
+    `${JSON.stringify({ scopes: rows, failedScopes: failures }, null, 2)}\n`,
+  );
+  console.table(rows);
```
JWThewes, line 122: `otherStatuses` renders as `[object Object]` in `console.table`. Confirmed in my run — flatten it or drop it from the table.

scripts/run-changed-backend-mutations.mjs
```diff
@@ -0,0 +1,248 @@
...
+const lambdaRoot = join(root, 'lambda');
+const reportRoot = join(root, 'reports', 'mutation', 'pr');
+const dryRun = process.argv.includes('--dry-run');
+const reportOnly = process.argv.includes('--report-only');
+const executionIncomplete = process.env.MUTATION_EXECUTION_INCOMPLETE === 'true';
+const workflowUrl = process.env.MUTATION_WORKFLOW_URL;
+const argument = (name) =>
+  process.argv.find((value) => value.startsWith(`--${name}=`))?.slice(name.length + 3);
+const base = argument('base') ?? 'origin/main';
+const explicitFiles = argument('files');
```
JWThewes, line 17: `.slice(name.length + 3)` is correct but cryptic — `parseArgs` from `node:util` is stdlib and would cover both scripts.

Separately, `--files=` isn't documented anywhere, and neither is `test:mutation:changed:dry`.

scripts/run-changed-backend-mutations.mjs
```diff
@@ -0,0 +1,248 @@
...
+const percent = (metrics) =>
+  metrics.totalMutants === 0 ? 'n/a' : `${metrics.mutationScore.toFixed(1)}%`;
```
JWThewes, line 19: Returns `NaN%` when `totalMutants > 0` but everything is ignored — `mutationScore` is `NaN` in that case, and the `=== 0` guard doesn't catch it.

scripts/run-changed-backend-mutations.mjs
```diff
@@ -0,0 +1,248 @@
...
+  process.exitCode = failures.length > 0 ? 1 : 0;
+} else {
+  const scopeRows = [];
+  const fileRows = [];
+  const noteworthyMutants = [];
+  const missingScopes = [];
+
+  for (const [scope, files] of filesByScope) {
+    const reportPath = join(reportRoot, `${scope}.json`);
+    if (!existsSync(reportPath)) {
```
JWThewes, line 134: Nothing clears `reports/mutation/pr/*.json` before a run. If a scope's Stryker run fails locally, this `existsSync` picks up the *previous* run's report and the summary presents stale numbers as current. Fine in CI (fresh checkout), but a footgun locally — `run-backend-mutations.mjs:65-66` does clean up, so let's be consistent.

.gitignore
```diff
@@ -41,6 +41,8 @@ node_modules
 
 # Test coverage output
 coverage/
+reports/
+.stryker-tmp/
```
JWThewes, line 45: `.stryker-tmp` should go in `.oxlintignore` too. Stryker leaves it behind when a run crashes, and oxlint would then lint 500+ sandbox copies. `coverage` and `lambda/*/.build` are already in there.

README.md
```diff
@@ -411,8 +411,35 @@ Run the unit tests and generate a coverage report:
 ```bash
 npm test                 # run all unit tests
 npm run test:coverage    # run tests with a coverage report (HTML in coverage/)
+npm run test:mutation:dry # validate the mutation-testing harness without running mutants
+npm run test:mutation     # mutate all backend JavaScript sources
```
JWThewes, line 415: This isn't really a runnable command. Measured mutant counts on this branch:

| Scope | Mutants | Note |
|---|---:|---|
| `shared` | 13,890 | |
| `v2-orchestrator` | 4,050 | baseline test run alone takes **61 s** |
| `untested` | 902 | full suite + Docker |
| `building-blocks` | 330 | 15 s end-to-end |

Extrapolating from `building-blocks`, `v2-orchestrator` on its own is hours. Can we either add an explicit "expect many hours, use `--scope=`" warning or drop this line and document only the scoped form?

README.md
```diff
@@ -411,8 +411,35 @@ Run the unit tests and generate a coverage report:
...
+by default because they require reloading the test environment and dominate the
+runtime on the largest domains. Pass `--include-static` for an exhaustive run.
+Mutation scores are initially informational so the first runs establish a
+baseline before any blocking threshold is chosen. Pull requests that change
+backend JavaScript also run an informational mutation campaign over each
+changed production file. The PR comment uses the official Stryker metrics and
+highlights surviving or uncovered mutants without blocking the pull request.
+Draft pull requests are skipped, static mutants are excluded, previous runs are
+cancelled after a new push, and the pilot stops after 30 minutes. A timeout
+produces a partial informational report rather than failing the pull request.
```
JWThewes, line 441: 27 lines in essentially one paragraph, mixing local commands, CI mechanics, the static-mutant rationale and a follow-up TODO. The rest of this file is short paragraph + code block. Worth splitting into a few paragraphs, or moving the CI mechanics over to `CONTRIBUTING.md`.

The "first PR after these workflows reach `main` must confirm..." sentence can go — see my note in `coverage-report.yml`.

JWThewes (changes requested):
Went through this on a scratch worktree — installed deps and actually ran the harness (Stryker on `sso-token`, `ws-authorizer`, `building-blocks`; dry runs on `shared`, `v2-orchestrator`, `untested`; the changed-files script incl. `--report-only`; plus lint/format/coverage). Comments inline are backed by those runs.

**What I like:** the privileged/unprivileged split is done properly (compute on the PR token, commenting via `workflow_run`), actions are SHA-pinned in line with the rest of the repo, `break: null` means nothing can block a PR, and lint/format pass clean. Coverage trending + mutation testing are both worth having.

**Why I'm requesting changes:** a few concrete defects, and the expensive/risky paths were never exercised — validation only covers `sso-token` (23 mutants, no testcontainers).

### Blockers

1. **`ignorePatterns` misses `terraform/`** (`stryker.config.mjs:32-43`). Stryker doesn't read `.gitignore`; on my checkout `terraform/` is 2.7 GB (1.0 GB of `.terraform` providers) and gets copied into the sandbox on every scope run.
2. **`.build/` bundles get mutated** (`stryker.config.mjs:12-16`). The `mutate` glob only excludes `test/**`. The changed-files script filters `/.build/` already.
3. **`incremental` makes local scores non-reproducible** (`stryker.config.mjs:18`). `--include-static` is one-way sticky — reproduced: 66.78% -> 62.42% -> 62.42% on `building-blocks`. Zero CI benefit since `reports/` isn't cached.
4. **`--report-only` destroys failure detail and reports a false cause** (`mutation-pr.yml:52-61`). A failed scope gets reported as a 30-minute timeout, and the scope name is dropped.
5. **No tests for 377 lines of new script code** (`scripts/`). `scripts/test/` has tests for all four existing scripts, run in CI via `npm run test:release`.

### Strongly want

- One infrastructure scope validated in CI — `concurrency: 4` means 4 gremlin + 4 DynamoDB-Local JVMs on a 4-vCPU runner, and that's untested.
- `infrastructureLambdas` in `vitest.config.js` derived or guarded by a test (correct today, will rot).
- Harness files added to `mutation-pr.yml`'s `paths`, plus `workflow_dispatch` (both are existing conventions here).
- `.stryker-tmp` in `.oxlintignore`.
- `if-no-files-found: error` on the coverage upload.

### Nice to have

- Drop the `untested` scope — 902 mutants + Docker to learn what `all: true` coverage already reports.
- Dedupe the two scripts (~2/3 overlap, and lambda enumeration is a third copy of `vitest.config.js`).
- Temper the `npm run test:mutation` README line — `shared` alone is 13,890 mutants.

Happy to pair on the `ignorePatterns` inversion if that's easiest.

PR #429, opened 2026-09-04
feat(install): support any Docker-API-socket container runtime (#420)

### LINKED ISSUE #420: [Feature]: Allow other container runtimes (e.g. finch) for installation
### Description

Currently, one needs to have docker installed to install the project. Some people, however, choose to use other container management software, such as podman or finch. Please update the installation routine to not have docker hardcoded and instead allow a user to pick (e.g. via a documented override env variable).

### Use case

(see above)

### Area

Infrastructure (Terraform)

### Additional context

E.g. the CDK [allows an override](https://docs.aws.amazon.com/cdk/v2/guide/build-containers.html) via the `CDK_DOCKER` environment variable.

jeromevdl:
*Issue #, if available:* fixes #420

Docker is hardcoded as the only container runtime for installation. Make it configurable so users on Podman, Rancher Desktop, and other socket-based runtimes can install without Docker.

- scripts/install.sh: prerequisite check honors CONTAINER_RUNTIME (default docker) instead of hardcoding "docker"
- terraform: document the DOCKER_HOST socket override on all three kreuzwerker/docker provider blocks (now consistent)
- docs: README summary + prerequisites section explaining the CONTAINER_RUNTIME / DOCKER_HOST contract; list verified runtimes (Docker, Podman, Rancher Desktop) and compatible ones (Colima, OrbStack); Finch and CLI-only build tools are unsupported (no Docker-API socket)

The image build talks to the Docker Engine API socket via DOCKER_HOST, so any runtime exposing such a socket works with no code change. Verified end-to-end (build → ECR push → AgentCore runtime) with Podman and Rancher Desktop.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -50,6 +50,8 @@ PERSISTED_AUTH_MODE="local"
 SOURCE=""
 ASSUME_YES="${AIDLC_YES:-0}"
 ALLOW_DOWNGRADE="${AIDLC_ALLOW_DOWNGRADE:-0}"
+# Binary checked by require_commands; image builds select their daemon with DOCKER_HOST.
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
 
 usage() {
     cat <<'EOF'
@@ -586,7 +588,7 @@ require_terraform_version() {
 
 require_commands() {
     local missing=0 command
-    local commands="git node npm terraform aws docker"
+    local commands="git node npm terraform aws $CONTAINER_RUNTIME"
     [[ "${AIDLC_TEST_MODE:-0}" == 1 ]] && commands="git node"
     for command in $commands; do
         if ! command -v "$command" >/dev/null 2>&1; then
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -50,8 +50,6 @@ PERSISTED_AUTH_MODE="local"
 SOURCE=""
 ASSUME_YES="${AIDLC_YES:-0}"
 ALLOW_DOWNGRADE="${AIDLC_ALLOW_DOWNGRADE:-0}"
-# Binary checked by require_commands; image builds select their daemon with DOCKER_HOST.
-CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
 
 usage() {
     cat <<'EOF'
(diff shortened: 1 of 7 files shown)


scripts/install.sh
```diff
@@ -50,6 +50,8 @@ PERSISTED_AUTH_MODE="local"
 SOURCE=""
 ASSUME_YES="${AIDLC_YES:-0}"
 ALLOW_DOWNGRADE="${AIDLC_ALLOW_DOWNGRADE:-0}"
+# Binary checked by require_commands; image builds select their daemon with DOCKER_HOST.
+CONTAINER_RUNTIME="${CONTAINER_RUNTIME:-docker}"
```
JWThewes: **Important — remove the redundant CLI prerequisite.** This variable only changes which executable `require_commands` looks for; it does not select the daemon. The Terraform image build uses the Docker-compatible API endpoint from `DOCKER_HOST`, and the local deployment path never invokes this CLI. As written, a valid socket-only or remote Docker-compatible endpoint is rejected unless an unrelated CLI is installed. Please keep this KISS: remove the container runtime from `require_commands` and document `DOCKER_HOST` as the actual configuration knob, or validate the endpoint itself if a preflight is required.

scripts/install.sh
```diff
@@ -586,7 +588,7 @@ require_terraform_version() {
 
 require_commands() {
     local missing=0 command
-    local commands="git node npm terraform aws docker"
+    local commands="git node npm terraform aws $CONTAINER_RUNTIME"
```
JWThewes: **Important — add regression coverage for the override.** The release-tool fixture currently always creates a fake `docker`, and no test mentions `CONTAINER_RUNTIME` or `DOCKER_HOST`. A future reversion to the hardcoded Docker check would therefore remain green. Please add a focused installer test with a controlled `PATH` and only the selected alternative CLI, for example `podman`, available. If the CLI check is removed per the other comment, test that a non-default `DOCKER_HOST` does not require `docker` on `PATH`.

JWThewes (dismissed):
Requesting changes for the two blocking items noted inline: remove the redundant container CLI prerequisite, since the socket/API endpoint is the actual dependency, and add regression coverage for the alternative-runtime path. The overall direction and documentation are otherwise good.

scripts/install.sh
jeromevdl: true, removing

scripts/install.sh
jeromevdl: removing CONTAINER_RUNTIME and test added

JWThewes (approved):
Looks good

PR #432, opened 2026-09-04
Feat: codefile traceability (#418)

### LINKED ISSUE #418: [Feature]: Restore CodeFile nodes and implementation traceability in the v2 knowledge graph
## Problem

The v2 Intent knowledge graph no longer contains code nodes, while the former Sprint graph exposed `CodeFile` vertices.

The graph currently connects requirements, stories, artifacts, and units of work, but stops before the implementation. We cannot navigate from an intent or unit to the files it created or changed, nor answer questions such as:

- Which files implement this requirement or story?
- Which intent or unit changed this file?
- Which implementation or test files are associated with an acceptance criterion?

This is a significant loss of end-to-end SDLC traceability.

## Previous v1 model

The v1 graph used this model:

```text
Sprint ─────CONTAINS───────> CodeFile
Task ───────IMPLEMENTED_BY─> CodeFile
UserStory ──IMPLEMENTED_BY─> CodeFile  (optional shortcut)
```

The complete implementation chain was:

```text
Requirement ─BREAKS_INTO─> UserStory
UserStory ───BREAKS_INTO─> Task
Task ────────IMPLEMENTED_BY─> CodeFile
```

A `CodeFile` contained:

```text
id
file_path
repository
commit_ref
summary
sprint_id
```

Its UUID identity meant that the node represented a file in the context of a particular Sprint/implementation revision, rather than one mutable global node per repository path.

## Proposed v2 model

Restore the same concepts using the v2 equivalents:

```text
Intent ──────CONTAINS───────> CodeFile
UnitOfWork ──IMPLEMENTED_BY─> CodeFile
Story ───────IMPLEMENTED_BY─> CodeFile  (evidence-backed shortcut)
```

`UnitOfWork` is the natural v2 equivalent of the v1 `Task`.

Preserve the existing label, edge vocabulary, and properties where possible:

```text
CodeFile:
  id
  file_path
  repository
  commit_ref
  summary
  intent_id
  unit_slug
  stage_instance_id
  file_kind             # implementation | test | configuration | documentation
  traceability_source   # git | aidlc-traceability
```

Use a deterministic, revision-scoped identity such as:

```text
hash(intent_id, repository, commit_ref, file_path)
```

This preserves the v1 semantics and code history while making ingestion idempotent. A repository-and-path-only identity would overwrite revision provenance and make `commit_ref` ambiguous.

A canonical global `SourceFile` node and `CodeFile ─REVISION_OF─> SourceFile` relationship can be considered later if cross-intent file history requires it.

## Use AIDLC structured traceability when available

Recent AIDLC v2 workflows introduced stable element IDs and per-stage `traceability.json` artifacts in [awslabs/aidlc-workflows#401](https://github.com/awslabs/aidlc-workflows/pull/401).

For `code-generation`, each coverage entry maps an upstream ID such as an AC, NFR, BR, FR, or NFR to an existing workspace-relative implementation or test file:

```json
{
  "stage": "code-generation",
  "unit": "u1-auth",
  "coverage": [
    { "id": "AC1.1.1", "status": "OK", "target": "src/auth/login.ts" },
    { "id": "NFR1.1", "status": "OK", "target": "src/cache/redis.ts" }
  ]
}
```

When this artifact is present and valid, ingest it as authoritative evidence and create `IMPLEMENTED_BY` edges between matching graph elements and `CodeFile` nodes.

## Backward compatibility

Compatibility must be capability-based, not based on a numeric `workflowVersion`. Versions are local to each workflow, and custom or forked workflows may gain structured traceability at any version.

- **All workflows:** use the Git results already captured after stage execution to create revision-scoped `CodeFile` nodes and link them to the Intent and `UnitOfWork`.
- **Traceability-capable workflows:** detect and validate a produced `traceability` artifact, then add precise specification-to-file edges from its coverage entries.
- **Older workflows:** retain code topology without fabricating requirement or acceptance-criterion links. Missing structured traceability is a supported degraded mode, not an execution failure.
- **Historical intents:** allow lazy enrichment or an explicit graph rebuild where sufficient Git/artifact data still exists.

This also lets custom workflows emit the same contract without depending on a specific upstream AIDLC version.

## Tests

As in v1, test files can initially remain `CodeFile` nodes, identified through `file_kind: test`. Dedicated `TestCase`, `TestExecution`, and test-result nodes should be introduced separately once executable test evidence is available.

## Acceptance criteria

- [ ] Files created or modified by v2 construction stages appear as `CodeFile` nodes in the Intent graph.
- [ ] The model retains the v1 `CodeFile` label and `IMPLEMENTED_BY` edge semantics.
- [ ] `Intent ─CONTAINS─> CodeFile` and `UnitOfWork ─IMPLEMENTED_BY─> CodeFile` relationships are created.
- [ ] Code file identity is revision-scoped using intent, repository, commit, and workspace-relative path.
- [ ] Code nodes retain provenance to the intent, unit, stage instance, repository, and commit SHA.
- [ ] Reprocessing the same stage and commit is idempotent.
- [ ] A valid `traceability.json` is ingested when present.
- [ ] `OK` coverage targets create evidence-backed `IMPLEMENTED_BY` links between known graph elements and implementation/test files.
- [ ] Workflows without `traceability.json` still produce code nodes and intent/unit links.
- [ ] Missing structured traceability never causes legacy workflow execution to fail.
- [ ] The Intent graph API and existing graph UI expose and render the restored code nodes and relationships.
- [ ] Coverage includes both traceability-capable and legacy workflow fixtures.

## Follow-ups

- Promote acceptance criteria to first-class graph nodes with stable IDs.
- Add dedicated `TestCase`, `TestExecution`, and test-result evidence.
- Add a canonical `SourceFile` model if cross-intent revision history is needed.
- Link code symbols where language-aware indexing is available.
- Add reverse impact analysis from changed code to affected requirements and tests.

jeromevdl:
*Issue #, if available:* #418

## Description of changes:
Brings back the CodeFile vocabulary and implementation traceability in the graph. Construction-stage files become CodeFile nodes linked to their Intent and UnitOfWork, and a valid `traceability.json` adds requirement→file evidence edges when present.

### Changes

- Every stage's changed files become CodeFile nodes with provenance (intent, unit, stage instance, repo, commit), linked `Intent --CONTAINS--> CodeFile` and `UnitOfWork --IMPLEMENTED_BY--> CodeFile`. This comes from git, so it works for legacy workflows that produce no traceability manifest.
- When a stage/unit produces a valid `traceability.json`, known derived elements (Requirement, Story, …) get an `IMPLEMENTED_BY` edge to the exact file revision. Detected from the produced manifest, not gated on workflow version.
- Missing, malformed, or oversized manifests and Neptune outages don't fail the stage. The git topology is still ingested and a degraded event is emitted.
- CodeFile identity is `hash(intent_id, repository, commit_ref, file_path)`, idempotent for the same stage + commit.
- A re-run or rewind supersedes prior revisions of the same file. History stays queryable in Neptune, but the graph renders one node per file — the same supersede/isCurrent handling used for other node types.
- API exposes the nodes, CONTAINS, and implementation edges; the UI adds the node panel fields, capability flag, and layer toggle.

### Notes

- No infra changes: Neptune/Gremlin is schemaless and the agentcore Lambda already has Neptune write access.
- Out of scope (not built): AC-as-nodes (maybe as a follow-up issue), TestCase/TestExecution, SourceFile/REVISION_OF, symbol linking, reverse impact.

### Testing

Backend suites (graph-writer, code-traceability, run-stage, knowledge-graph) and the frontend suite cover both the traceability and legacy paths, including invalid-JSON degrade, multi-repo, size cap, CME retry, and the supersede/render-latest behavior. Run `npm test` (Docker required for the testcontainers-backed backend suites).

Closes #418

### DIFF
--- a/frontend/src/pages/IntentGraphPage.tsx
+++ b/frontend/src/pages/IntentGraphPage.tsx
@@ -78,7 +78,9 @@ export default function IntentGraphPage() {
               <TooltipTrigger asChild>
                 <button
                   onClick={() => setLayer('all')}
-                  aria-label="Items and Units layer"
+                  aria-label={
+                    hasCodeTraceability ? 'Items, Units and Code layer' : 'Items and Units layer'
+                  }
                   aria-pressed={layer === 'all'}
                   className={cn(
                     'flex items-center gap-1 rounded-sm px-1.5 sm:px-2 py-1 text-[10px] font-medium transition-all',
--- a/frontend/src/pages/IntentGraphPage.tsx
+++ b/frontend/src/pages/IntentGraphPage.tsx
@@ -12,14 +12,15 @@ export default function IntentGraphPage() {
 
   // Shared SWR graph cache — the same fetch the workbench popovers and the
   // derived-items section use (see useIntentGraph).
-  const { nodes, edges, loading, error } = useIntentGraph(projectId, intentId);
-  // Layer toggle: 'artifacts' hides the derived projection (typed items +
-  // unit DAG, nodes tagged graphLayer='derived'); 'all' shows everything.
+  const { nodes, edges, loading, error, hasCodeTraceability } = useIntentGraph(projectId, intentId);
+  // Layer toggle: 'artifacts' hides graph projections (typed items, unit DAG,
+  // and CodeFile implementation revisions); 'all' shows every capability the
+  // API actually returned.
   const [layer, setLayer] = useState<'artifacts' | 'all'>('artifacts');
 
   const { visibleNodes, visibleEdges } = useMemo(() => {
     if (layer === 'all') return { visibleNodes: nodes, visibleEdges: edges };
-    const kept = nodes.filter((n) => n.graphLayer !== 'derived');
+    const kept = nodes.filter((n) => !n.graphLayer);
     const keptIds = new Set(kept.map((n) => n.id));
     return {
       visibleNodes: kept,
@@ -87,10 +88,16 @@ export default function IntentGraphPage() {
                   )}
                 >
                   <Box className="h-3 w-3" />
-                  <span className="hidden sm:inline">+ Items &amp; Units</span>
+                  <span className="hidden sm:inline">
+                    {hasCodeTraceability ? '+ Items, Units & Code' : '+ Items & Units'}
+                  </span>
                 </button>
               </TooltipTrigger>
-              <TooltipContent>Include derived items &amp; units</TooltipContent>
+              <TooltipContent>
+                {hasCodeTraceability
+                  ? 'Include derived items, units, and implementation files'
+                  : 'Include derived items & units'}
+              </TooltipContent>
             </Tooltip>
           </div>
         }
--- a/lambda/agentcore/code-traceability.js
+++ b/lambda/agentcore/code-traceability.js
@@ -9,6 +9,21 @@ import { createGraphWriter, closeGraphSource } from './mcp/graph-writer.js';
 
 const TRACEABILITY_FILE = 'traceability.json';
 
+// Agent-produced traceability.json is untrusted input. Cap the size before
+// readFile+JSON.parse so a pathological artifact degrades (treated as invalid)
+// instead of risking OOM. 5 MiB is far above any legitimate coverage manifest.
+const MAX_TRACEABILITY_BYTES = 5 * 1024 * 1024;
+
+// Neptune serializes concurrent vertex/edge writes optimistically; parallel unit
+// lanes committing at once can collide with a ConcurrentModificationException.
+// Reuse the codebase's attempt-loop-with-linear-backoff shape (cli/codex-store.js)
+// so a transient collision is retried before the projection degrades.
+const CME_RETRY_DELAYS_MS = [100, 250, 500];
+export const isConcurrentModification = (error) =>
+  /ConcurrentModification|ConcurrentModificationException|concurrent(?:ly)? modif/i.test(
+    `${error?.name ?? ''} ${error?.message ?? ''} ${error?.code ?? ''}`,
+  );
+
 export const normalizeWorkspacePath = (value) => {
   if (typeof value !== 'string' || !value.trim()) return null;
   const portable = value.trim().replaceAll('\\', '/');
@@ -103,7 +118,8 @@ export const loadProducedTraceability = async ({ repoDir, changedFiles, stageId,
   for (const file of candidates) {
     try {
       const artifactPath = path.resolve(repoDir, file);
-      if (!(await lstat(artifactPath)).isFile()) continue;
+      const stat = await lstat(artifactPath);
+      if (!stat.isFile() || stat.size > MAX_TRACEABILITY_BYTES) continue;
       const parsed = JSON.parse(await readFile(artifactPath, 'utf8'));
       const result = validateTraceabilityDocument(parsed, {
         expectedStage: stageId,
@@ -184,7 +200,13 @@ export const collectCodeTraceabilityBatches = async ({
   return batches;
 };
 
-export const ingestStageCodeTraceability = async ({ openGraph, scope, ...input }) => {
+export const ingestStageCodeTraceability = async ({
+  openGraph,
+  scope,
+  sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms)),
+  createWriter = createGraphWriter,
+  ...input
+}) => {
   const batches = await collectCodeTraceabilityBatches(input);
   if (!batches.length || !openGraph) {
     return { batches: batches.length, codeFiles: 0, evidenceEdges: 0, statuses: [] };
@@ -193,11 +215,23 @@ export const ingestStageCodeTraceability = async ({ openGraph, scope, ...input }
   let g = null;
   try {
     g = await openGraph();
-    const writer = createGraphWriter({ g, scope });
+    const writer = createWriter({ g, scope });
     let codeFiles = 0;
     let evidenceEdges = 0;
     for (const batch of batches) {
-      const result = await writer.ingestCodeFiles(batch);
+      // Retry a transient CME (parallel unit lanes colliding on the shared
+      // Intent anchor); only a persistent failure propagates to degrade.
+      let result;
+      for (let attempt = 0; ; attempt += 1) {
+        try {
+          result = await writer.ingestCodeFiles(batch);
+          break;
+        } catch (error) {
+          if (!isConcurrentModification(error) || attempt >= CME_RETRY_DELAYS_MS.length)
+            throw error;
+          await sleep(CME_RETRY_DELAYS_MS[attempt]);
+        }
+      }
       codeFiles += result.codeFiles;
       evidenceEdges += result.evidenceEdges;
     }
--- a/lambda/agentcore/code-traceability.js
+++ b/lambda/agentcore/code-traceability.js
@@ -1,0 +1,213 @@
+// Stage-exit CodeFile projection. Git remains the universal source of changed
+// files; a valid, stage-produced traceability.json adds precise evidence edges.
+// Missing/malformed traceability is an expected legacy mode and never blocks a
+// stage — callers still ingest the Git topology.
+
+import path from 'node:path';
+import { lstat, readFile } from 'node:fs/promises';
+import { createGraphWriter, closeGraphSource } from './mcp/graph-writer.js';
+
+const TRACEABILITY_FILE = 'traceability.json';
+
+export const normalizeWorkspacePath = (value) => {
+  if (typeof value !== 'string' || !value.trim()) return null;
+  const portable = value.trim().replaceAll('\\', '/');
+  if (portable.startsWith('/') || /^[a-zA-Z]:\//.test(portable)) return null;
+  const normalized = path.posix.normalize(portable).replace(/^\.\//, '');
+  if (!normalized || normalized === '.' || normalized === '..' || normalized.startsWith('../')) {
+    return null;
+  }
+  return normalized;
+};
+
+export const traceabilitySlug = (value) =>
+  String(value ?? '')
+    .trim()
+    .toLowerCase()
+    .replace(/[^a-z0-9]+/g, '-')
+    .replace(/^-+|-+$/g, '');
+
+export const classifyFileKind = (filePath) => {
+  const normalized = String(filePath ?? '').toLowerCase();
+  const name = path.posix.basename(normalized);
+  if (
+    /(^|\/)(__tests__|tests?|spec)(\/|$)/.test(normalized) ||
+    /\.(test|spec)\.[^.]+$/.test(name)
+  ) {
+    return 'test';
+  }
+  if (/(^|\/)(docs?|documentation)(\/|$)/.test(normalized) || /\.(md|mdx|rst|adoc)$/.test(name)) {
+    return 'documentation';
+  }
+  if (
+    /(^|\/)(config|terraform|\.github)(\/|$)/.test(normalized) ||
+    /(^|\.)(env|ya?ml|json|toml|ini|tf|tfvars|lock)$/.test(name) ||
+    /^(dockerfile|makefile)$/.test(name)
+  ) {
+    return 'configuration';
+  }
+  return 'implementation';
+};
+
+export const validateTraceabilityDocument = (
+  value,
+  { expectedStage = null, expectedUnit = null } = {},
+) => {
+  if (!value || typeof value !== 'object' || Array.isArray(value)) {
+    return { valid: false, reason: 'document must be an object' };
+  }
+  const stage = typeof value.stage === 'string' ? value.stage.trim() : '';
+  const unit = typeof value.unit === 'string' ? value.unit.trim() : '';
+  if (!stage || !unit || !Array.isArray(value.coverage)) {
+    return { valid: false, reason: 'stage, unit, and coverage are required' };
+  }
+  if (expectedStage && stage !== expectedStage) {
+    return { valid: false, reason: `stage mismatch (${stage})` };
+  }
+  if (expectedUnit && unit !== expectedUnit) {
+    return { valid: false, reason: `unit mismatch (${unit})` };
+  }
+
+  const coverage = [];
+  for (const [index, entry] of value.coverage.entries()) {
+    if (!entry || typeof entry !== 'object' || Array.isArray(entry)) {
+      return { valid: false, reason: `coverage[${index}] must be an object` };
+    }
+    const id = typeof entry.id === 'string' ? entry.id.trim() : '';
+    const status = typeof entry.status === 'string' ? entry.status.trim().toUpperCase() : '';
+    const target = entry.target == null ? null : normalizeWorkspacePath(entry.target);
+    if (!id || !status || (entry.target != null && !target) || (status === 'OK' && !target)) {
+      return { valid: false, reason: `coverage[${index}] is invalid` };
+    }
+    coverage.push({ id, status, target });
+  }
+  return { valid: true, document: { stage, unit, coverage } };
+};
+
+const isPresentFile = async (root, filePath) => {
+  try {
+    const entry = await lstat(path.resolve(root, filePath));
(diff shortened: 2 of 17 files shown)


lambda/agentcore/mcp/graph-writer.js
```diff
@@ -1156,6 +1181,148 @@ export const createGraphWriter = ({ g, scope = {}, clock } = {}) => {
...
+        intent_id: scope.intentId,
+        unit_slug: effectiveUnitSlug,
+        stage_instance_id: effectiveStageInstanceId,
+        file_kind: file.fileKind ?? 'implementation',
+        traceability_source: file.traceabilitySource ?? 'git',
+        updated_at: now(),
+        // This revision is the current node for its file_path. Empty marker =
+        // current (see isCurrentRow); re-ingesting a rewound revision clears any
+        // stale mark it carried from a prior supersede.
+        superseded_at: '',
```
JWThewes, line 1265: 🟡 **Current CodeFile selection has no canonical order before fan-in**

Every unit-stage writer makes its branch-local revision current immediately, before the orchestrator's serialized lane merge establishes which commit is actually on the intent branch. Completion order therefore decides which revision renders, even when that branch has not merged or is later skipped.

```javascript
superseded_at: '',
...
const priorRevisions = (
  await g.V().has(CODE_FILE_LABEL, 'intent_id', scope.intentId)
    .has('repository', repository)
    .has('file_path', file.filePath)
```

**Why it matters:** Parallel branch ingestion assigns currentness early → a race winner is mistaken for the canonical merged revision → graph consumers can see an arbitrary or unmerged implementation as the one current file.

**Suggestion:** Persist immutable per-stage revisions at stage exit, but select or stamp the single current revision during serialized lane merge/fan-in using the merged intent commit/ancestry. If pre-fan-in display is required, scope currentness by unit instead of globally by intent/repository/path.

_[code-review-sage]_

lambda/agentcore/mcp/graph-writer.js
```diff
@@ -1156,6 +1181,148 @@ export const createGraphWriter = ({ g, scope = {}, clock } = {}) => {
...
+      const priorRevisions = (
+        await g
+          .V()
+          .has(CODE_FILE_LABEL, 'intent_id', scope.intentId)
+          .has('repository', repository)
+          .has('file_path', file.filePath)
+          .valueMap(true)
+          .toList()
+      ).map(flattenValueMap);
+      for (const prior of priorRevisions) {
```
JWThewes, line 1293: 🔴 **Concurrent ingests can supersede every revision of the same file**

The current transition is a non-atomic read-then-write sequence. If lanes A and B ingest the same path concurrently, both first write their own revision with an empty superseded_at, both can read both revisions as current, and each then marks the other revision superseded. The knowledge-graph reader filters every superseded CodeFile, so no revision remains visible.

```javascript
for (const prior of priorRevisions) {
  if (prior.id === id || !isCurrentRow(prior)) continue;
  await g
    .V()
    .has(CODE_FILE_LABEL, 'id', prior.id)
    .property(cardinality.single, 'superseded_at', now())
    .next();
}
```

**Why it matters:** Concurrent lane writes to a shared path → each writer supersedes the other's newly-current vertex → all revisions carry superseded_at → the file and all of its implementation edges disappear from the Intent graph under a normal parallel-workflow interleaving.

**Suggestion:** Make the current transition atomic and ordered per (intent, repository, path), preferably by deferring it to serialized fan-in. Otherwise use a compare-and-set/current-pointer transaction that can leave exactly one winner, and add a concurrent two-writer test proving one and only one revision remains current.

_[code-review-sage]_

lambda/agentcore/code-traceability.js
```diff
@@ -0,0 +1,249 @@
...
+  }
+  if (expectedStage && stage !== expectedStage) {
+    return { valid: false, reason: `stage mismatch (${stage})` };
+  }
+  if (expectedUnit && unit !== expectedUnit) {
+    return { valid: false, reason: `unit mismatch (${unit})` };
+  }
+
+  const coverage = [];
+  for (const [index, entry] of value.coverage.entries()) {
```
JWThewes, line 97: 🔴 **A small manifest can trigger tens of thousands of Neptune traversals**

The 5 MiB file-size check does not cap coverage cardinality, identifier length, or unique evidence IDs. A 2.54 MiB document with 50,000 OK entries is accepted by this validator, and ingestCodeFiles subsequently performs at least one, often two, Neptune lookups for every unique ID before writing edges.

```javascript
const coverage = [];
for (const [index, entry] of value.coverage.entries()) {
  ...
  coverage.push({ id, status, target });
}
```

**Why it matters:** Agent/repository-controlled traceability.json entry point → the artifact-to-Neptune trust boundary accepts an unbounded coverage array → a crafted valid manifest amplifies into tens of thousands of remote traversals on the awaited stage-success path → Neptune can be saturated or the worker can time out before recording SUCCEEDED, creating a denial of service despite the promised degraded mode.

**Suggestion:** Reject manifests above explicit limits for coverage entries, unique IDs per target, ID length, and aggregate evidence edges before building the batch. Batch-resolve accepted IDs and enforce a wall-clock/work budget so excess evidence deterministically returns invalid/degraded; add a test near each limit.

_[code-review-sage]_

lambda/agentcore/code-traceability.js
```diff
@@ -0,0 +1,249 @@
...
+    }
+  }
+  if (valid.length === 0) return { status: 'invalid', document: null, files: candidates };
+
+  // A non-unit stage passes expectedUnit=null, so two manifests declaring
+  // DIFFERENT units can both validate. Merging their coverage under the first
+  // manifest's unit would mis-stamp the batch, so keep only manifests that
+  // agree with the first unit.
+  const first = valid[0].document;
+  const merged = valid.filter(({ document }) => document.unit === first.unit);
```
JWThewes: 🟡 **Valid manifests for later units are silently discarded**

When a non-unit stage changes multiple valid traceability.json files with different unit values, the loader keeps only the files matching the lexicographically first manifest's unit. It still returns status='valid', so the discarded valid manifests produce neither evidence edges nor a degraded signal.

```javascript
const first = valid[0].document;
const merged = valid.filter(({ document }) => document.unit === first.unit);
return {
  status: 'valid',
  document: {
    stage: first.stage,
    unit: first.unit,
    coverage: merged.flatMap(({ document }) => document.coverage),
  },
```

**Why it matters:** Multiple valid unit manifests in one non-unit stage → all units except the first are filtered out while status remains valid → requirement-to-file evidence silently disappears and operators are told ingestion succeeded.

**Suggestion:** Return one traceability batch per distinct unit (preserving each manifest's evidence), or reject the ambiguous set as invalid and emit a bounded diagnostic naming the conflicting units. Add a multi-manifest test for the chosen contract.

_[code-review-sage]_

frontend/src/pages/IntentGraphPage.tsx
```diff
@@ -87,10 +90,16 @@ export default function IntentGraphPage() {
                   )}
                 >
                   <Box className="h-3 w-3" />
-                  <span className="hidden sm:inline">+ Items &amp; Units</span>
+                  <span className="hidden sm:inline">
```
JWThewes, line 93: 🟡 **The maintained graph documentation omits the new implementation layer**

This change adds CodeFile to the API contract and changes the Graph page from two layers to an Items, Units & Code view, but docs/concepts/data-model.md still presents the normative vertex/edge model without CodeFile or IMPLEMENTED_BY and docs/using-the-platform/intent-observability.md still documents only Artifacts and Items & Units.

```tsx
{hasCodeTraceability ? '+ Items, Units & Code' : '+ Items & Units'}
```

**Why it matters:** The runtime contract changes without its maintained docs → users and future maintainers rely on a model that says the implementation layer does not exist → integrations, support guidance, and later graph changes can contradict shipped behavior.

**Suggestion:** Update the conceptual diagram, vertex/edge tables, graph-layer explanation, and Graph page guide in this PR to document CodeFile identity/provenance, IMPLEMENTED_BY relationships, current-revision behavior, and the capability-dependent Code layer label.

_[code-review-sage]_

JWThewes (dismissed):

jeromevdl:
Thanks for the review. Addressed in c69cf7f:                                                                                                                              
                                                                                                                                                                                     
  🔴 Concurrent ingests can supersede every revision (graph-writer.js) — fixed. Confirmed real: the reader filters superseded CodeFiles, so two parallel lanes each superseding the other's freshly-current revision would drop the file entirely. The supersede sweep is now scoped by unit_slug, so a lane only supersedes prior revisions of its own unit. A file    touched by several units renders once per unit (accurate — it reflects real multi-unit authorship), and a unit re-run still supersedes only that unit's prior revision. Parallel lanes are separate stage executions with distinct unitSlug, so they can no longer collide on the same currentness. Added a two-unit concurrent-writer test proving both revisions stay current.                                                                                                                                                                      
                                                                                                                                                                                     
  🔴 A small manifest can trigger tens of thousands of Neptune traversals (code-traceability.js) — fixed. Added explicit caps before building the batch: max 1000 coverage entries and max 512-char ids. Over the limit → the manifest degrades to invalid like any other bad artifact. Test added near the limit.                                                    
                                                                                                                                                                                     
  🟡 Valid manifests for later units silently discarded (code-traceability.js) — investigated; the scenario isn't reachable, but the code was still wrong. Every stage that produces traceability.json is for_each: unit-of-work and writes exactly one file to its own per-unit record dir; the only non-unit stage that touches traceability (build-and-test) produces none. So a single ingest never legitimately sees more than one manifest — the "keep the first unit, merge the rest" branch was dead code that also lied by returning valid.        
 Removed it: a single ingest now uses its one manifest, and >1 valid manifest (only possible from anomalous agen output) degrades to invalid, retaining the Git topology and emitting the v2.traceability.degraded event instead of silently dropping evidence. Test added.                                                                                     
                                                                                                                                                                                     
  🟡 Docs omit the implementation layer — fixed. Updated docs/concepts/data-model.md (conceptual diagram, vertex/edge tables, new "Implementation layer" section covering CodeFile identity/provenance, IMPLEMENTED_BY, current-revision behavior) and docs/using-the-platform/intent-observability.md (the capability-dependent Code graph layer).                   
                                                                                                                                                                                     
  🟡 No canonical order before fan-in (graph-writer.js) — partially addressed / follow-up. The unit-scoped currentness above removes the cross-lane supersede race and the "arbitrary winner across units" ambiguity. Full fan-in-time selection using merged intent-branch ancestry (so an unmerged/skipped branch's revision is never stamped current) is a larger orchestrator change I've left out of this PR — happy to open a follow-up issue if you'd like that stricter guarantee.                                                              
                                                                                                                                                                                     
Tests: code-traceability 17/17, graph-writer 77/77, full suite 429/429.

JWThewes:
Thanks @jeromevdl. I rechecked the latest head, `834a7851325070e1445675026e55deef2e979a67`. Most of the original findings are addressed. Scoping currentness by unit matches the suggested pre-fan-in alternative, so canonical selection from the merged intent branch can remain a follow-up.

I still found two reproducible bugs that should be fixed before merging:

1. **P1 — Files inside newly created directories silently lose traceability.**

   In [code-traceability.js, lines 179–184](https://github.com/aws-samples/sample-collaborative-ai-dlc/blob/834a7851325070e1445675026e55deef2e979a67/lambda/agentcore/code-traceability.js#L179-L184), the collector assumes `gitResult.results[].files` contains individual files. However, `commitAll` builds that list from `git status --porcelain` before staging, which collapses new untracked directories.

   Reproduction with the real Git engine: create `src/index.ts` and `records/u1/traceability.json` in new directories. Both files are committed successfully, but the reported paths are `["records/", "src/"]`. The collector rejects those directories through `isPresentFile`, returning **zero batches**. Neither the implementation file nor its evidence reaches the graph.

   Please enumerate individual paths from the committed Git diff and add a regression test that exercises the real Git-engine output for new directories.

2. **P2 — Park/resume loses traceability for work committed before the pause.**

   The [projection hook in run-stage.js](https://github.com/aws-samples/sample-collaborative-ai-dlc/blob/834a7851325070e1445675026e55deef2e979a67/lambda/agentcore/commands/run-stage.js#L2525-L2549) runs only after the success gates, while Git commits and pushes work before the pending-question check. A parked stage returns before projection.

   Reproduction with a per-unit `code-generation` stage: modify tracked `app.ts`, commit the work while parking on a human question, then answer and resume without further edits. The resumed stage reaches `SUCCEEDED`, but its clean Git result contains no files to project, so `app.ts` never enters the graph.

   Please retain pending commit/file results across park/retry and project them when the stage eventually succeeds, or reconstruct the complete stage changes from Git. Add coverage for a successful resume with no additional edits.

There is also a small documentation correction: [data-model.md](https://github.com/aws-samples/sample-collaborative-ai-dlc/blob/834a7851325070e1445675026e55deef2e979a67/docs/concepts/data-model.md#L125-L129) and [intent-observability.md](https://github.com/aws-samples/sample-collaborative-ai-dlc/blob/834a7851325070e1445675026e55deef2e979a67/docs/using-the-platform/intent-observability.md#L58) still describe one current node per file. The new behavior is one current revision per **unit and file**, so those descriptions should reflect that.

Validation: the existing focused suites passed locally (**248 backend tests and 15 frontend tests**), and CI is green. I added local regression tests for the two scenarios above; **both fail on the latest head**, with zero collected batches/files where the committed implementation should be present.

JWThewes (changes requested):
See previous comment

jeromevdl:
All three findings are addressed:

  1. P1 — New-directory traceability: fixed in `80f4593`.
     Git now reports exact committed file paths. Added real-Git regression coverage.

  2. P2 — Park/resume traceability: fixed in `e4c56de`.
     Parked commits are retained and projected after a successful resume with no additional edits.

  3. Documentation: corrected in `79220e4`.
     The docs now describe one current revision per unit and file.

JWThewes:
@jeromevdl there has been no push after my comment. Are we missing a push here?

jeromevdl:
> @jeromevdl there has been no push after my comment. Are we missing a push here?

my bad, didn't push :)

JWThewes (dismissed):
Thanks, I rechecked the pushed head, `79220e49b9f18260fc3d83100271d3573495607f`. The new-directory fix, park followed by a clean successful resume, and documentation correction are now verified. Canonical selection at fan-in can remain the previously agreed follow-up.

I found one remaining reproducible issue to fix before merging:

**P2 — Ordinary failure/retry still loses committed code traceability.**

In [run-stage.js](https://github.com/aws-samples/sample-collaborative-ai-dlc/blob/79220e49b9f18260fc3d83100271d3573495607f/lambda/agentcore/commands/run-stage.js#L2324-L2328), prior commit references are carried only when `resumeFrom` is set. The current leg's references are [persisted only when parking on a human question](https://github.com/aws-samples/sample-collaborative-ai-dlc/blob/79220e49b9f18260fc3d83100271d3573495607f/lambda/agentcore/commands/run-stage.js#L2440-L2457). A CLI failure returns before that persistence, even though the engine has already committed the work. On a clean retry, there are no new Git files to project.

Reproduction with the real Git engine and a per-unit `code-generation` stage:

1. Modify tracked `app.ts` and add a valid `records/billing/traceability.json` mapping `REQ1` to `app.ts`.
2. Let the CLI exit with code 1. The engine commits both files, then the stage returns `cli_nonzero_exit` without projection.
3. Retry the same stage with exit code 0 and no additional edits.
4. The retry reaches **`SUCCEEDED`**, but returns **`changedFiles: []` and zero collected batches**. Neither `app.ts` nor its implementation evidence reaches the graph.

Please retain accumulated commit references before post-commit failure/park returns and preserve/reload them across supported retries, or reconstruct the complete stage changes from a durable Git baseline. A fresh stage-row write and the explicit retry/reset path also need to preserve the required provenance; merely saving it on failure will not be sufficient. Add a regression for a failed attempt followed by a successful retry with no additional edits.

Validation: **426 focused backend tests and 15 frontend tests passed**, and CI is green. An independent two-leg regression confirms that park → answer → clean resume now preserves code and evidence. The additional failure → clean retry regression fails on this head. These lifecycle regressions use real Git commits and the real collector, with CLI execution and successful remote-push acknowledgement stubbed.

JWThewes (approved):

PR #435, opened 2026-09-07
Fix(433): intent header UI/UX

### LINKED ISSUE #433: [Bug]: Simplify header in Intent screen
### Description

The top of the intent screen contains too much information, especially it contains the details of the environment with very technical informations (sha, versions, ...), see screenshot below. This is way too complex and should not be the first visible thing in the screen. We should add all these little details in a new screen or pop over / modal accessible from the menu on the right (`...`).

The header should only contain the title, the discuss button and the `...` menu (the only useful / actionable elements), and maybe the status of the intent (Running, succeeded,...). All the other elements (scope, harness, environment, issue) should be in another screen or pop over / modal accessible from a sub menu "Intent config". Even the "Reshape remaining stages" should be in the `...` menu as a submenu with dedicated screen or modal.

### Steps to reproduce

1. Create a space
2. Create an intent 
3. Start an intent

### Logs or screenshots

Image

### Environment

_No response_

jeromevdl:
*Issue #, if available:*  #433

This PR simplifies the intent workspace header and moves secondary configuration controls into focused dialogs.

### Changes

  - Moved scope progress below the intent title as an interactive phase breadcrumb.
  - Progress counts now reflect the stages selected by the effective scope.
  - Phases without selected stages are omitted.
  - Phase details show resolved, running, failed, and pending stages.
  - Moved environment, execution, agent, model, credentials, and source details into the **Intent configuration** dialog.
  - Added tracker source links and provider icons.
  - Added direct access from configuration to **Reshape**.
  - Moved scope reshaping into a dedicated dialog.
  - Completed, running, skipped, and earlier stages remain locked during reshaping.
  - Updated status colors and icons for waiting, running, successful, and failed states.
  - Kept the Discuss action as an icon beside the actions menu.

  This is a frontend-only change and requires no backend update.

### Screenshots

image

image

image

image

image

Closes #433 

### DIFF
--- a/frontend/src/components/intent/IntentConfigurationDialog.tsx
+++ b/frontend/src/components/intent/IntentConfigurationDialog.tsx
@@ -1,8 +1,16 @@
-import { Boxes, ExternalLink } from 'lucide-react';
+import {
+  Boxes,
+  Check,
+  ExternalLink,
+  GitBranch,
+  LoaderCircle,
+  Workflow,
+  XCircle,
+} from 'lucide-react';
 import { useIntent } from '@/contexts/IntentContext';
 import { AGENT_CLI_METADATA, AGENT_CREDENTIAL_SOURCE_LABELS } from '@/lib/agentCli';
 import { getIntentStageSelection } from '@/lib/intentStageSelection';
-import { formatTrackerSourceLabel } from '@/lib/trackerSourceLabel';
+import { getTrackerProvider } from '@/lib/trackerProviders';
 import { Badge } from '@/components/ui/badge';
 import { Button } from '@/components/ui/button';
 import {
@@ -15,10 +23,27 @@ import {
   DialogTitle,
 } from '@/components/ui/dialog';
 import { cn } from '@/lib/utils';
+import type { IntentSource } from '@/services/intents';
 
 interface IntentConfigurationDialogProps {
   open: boolean;
   onOpenChange: (open: boolean) => void;
+  onOpenReshape?: () => void;
+}
+
+function trackerProviderId(provider: string) {
+  if (provider === 'github') return 'github-issues';
+  if (provider === 'gitlab') return 'gitlab-issues';
+  if (provider === 'bitbucket') return 'bitbucket-issues';
+  return provider;
+}
+
+function formatSourceReference(source: IntentSource) {
+  const resourceType = (source.resourceType || 'issue').toLowerCase();
+  const resourceId = source.resourceId.startsWith('#')
+    ? source.resourceId
+    : `#${source.resourceId}`;
+  return `${resourceType} ${resourceId}`;
 }
 
 function Definition({
@@ -51,7 +76,11 @@ function Definition({
   );
 }
 
-export function IntentConfigurationDialog({ open, onOpenChange }: IntentConfigurationDialogProps) {
+export function IntentConfigurationDialog({
+  open,
+  onOpenChange,
+  onOpenReshape,
+}: IntentConfigurationDialogProps) {
   const { detail, compiled, initializationPhasePaths } = useIntent();
   if (!detail) return null;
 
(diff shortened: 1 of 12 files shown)


frontend/src/components/intent/RecomposePanel.tsx
```diff
@@ -110,15 +120,47 @@ export function RecomposePanel({
...
+    return latestRanIndex;
+  }, [frozen, intent.currentPhase, intent.currentStage, orderedNodes, phases]);
   const lockedStageIds = useMemo(
     () =>
       new Set([
         ...(compiled?.graph.nodes ?? [])
           .filter((n) => initPhasePath != null && n.phasePath === initPhasePath)
           .map((n) => n.stageId),
         ...frozen.keys(),
+        ...orderedNodes.slice(0, Math.max(0, progressBoundaryIndex)).map((n) => n.stageId),
```
JWThewes, line 161: Earlier-stage locking is only enforced by `toggleStage`: a completed composer proposal is passed directly to `applyAndRelaunch`, so it can change a stage included in this new `lockedStageIds` set. The server's frozen-state check only knows about stages with execution rows; an earlier stage excluded by the original projection has no row and can therefore be added, causing `/recompose` to relaunch from it. Please enforce this boundary in the canonical backend/composer frozen grid and validate proposals against it, rather than only disabling manual checkboxes.

frontend/src/components/intent/IntentConfigurationDialog.tsx
```diff
@@ -0,0 +1,265 @@
...
+  if (provider === 'gitlab') return 'gitlab-issues';
+  if (provider === 'bitbucket') return 'bitbucket-issues';
+  return provider;
+}
+
+function formatSourceReference(source: IntentSource) {
+  const resourceType = (source.resourceType || 'issue').toLowerCase();
+  const resourceId = source.resourceId.startsWith('#')
+    ? source.resourceId
+    : `#${source.resourceId}`;
```
JWThewes: This unconditionally adds `#`, which regresses Jira references (`TAS-01` becomes `issue #TAS-01`). The existing `formatTrackerSourceLabel` helper already preserves Jira keys and centralizes provider-specific formatting. Please reuse it instead of introducing a second formatter.

frontend/src/components/layout/IntentPipelineBar.tsx
```diff
@@ -1,121 +1,240 @@
...
   return (
-    <div className="h-11 border-b bg-background flex items-center px-3 gap-1 overflow-x-auto md:overflow-visible">
-      <div className="flex items-center gap-0.5 shrink-0">
+    <div className="space-y-2" data-testid="intent-phase-breadcrumb">
+      {intent.scope && (
+        <div className="flex items-center">
+          <ScopeBadge scope={intent.scope} className="px-2 py-0 text-[10px]" />
+        </div>
+      )}
+      <div className="flex w-full min-w-max overflow-x-auto pr-4">
```
JWThewes: `min-w-max` and `overflow-x-auto` are on the same element. On narrow viewports the element grows to max-content width, so it does not scroll internally and instead makes the surrounding content overflow. Please use a constrained outer `w-full overflow-x-auto` wrapper and an inner `flex min-w-max` row, or remove `min-w-max` here.

frontend/src/components/intent/IntentConfigurationDialog.tsx
```diff
@@ -0,0 +1,265 @@
...
+                variant="outline"
+                className={cn(
+                  'shrink-0 gap-1.5 text-[10px]',
+                  waiting && 'border-agent-waiting/30 bg-agent-waiting/10 text-agent-waiting',
+                  running && 'border-agent-running/30 bg-agent-running/10 text-agent-running',
+                  succeeded && 'border-agent-success/30 bg-agent-success/10 text-agent-success',
+                  failed && 'border-destructive/30 bg-destructive/10 text-destructive',
+                )}
+              >
+                {(waiting || running) && <LoaderCircle className="h-3 w-3 animate-spin" />}
```
JWThewes: WAITING is a parked state, but this renders the same animated spinner as RUNNING. Existing status UI uses a pause/question icon for waiting and reserves spinning loaders for active execution. Please keep the waiting color but use the established non-spinning waiting icon.

JWThewes (dismissed):
Requesting changes for four issues: the new earlier-stage lock can be bypassed by composer proposals, Jira source labels regress, the breadcrumb overflow container is not responsive, and WAITING is shown as actively running. Focused tests, type-checking, and lint pass, but these cases are not covered.

jeromevdl:
Commit: 5c4721f

  - Composer proposals preserve all locked stage values.
  - Jira keys use the shared source-label formatter.
  - Breadcrumb scrolling works on narrow screens.
  - WAITING uses a non-spinning pause icon.

JWThewes (approved):
Looks good to me.
Created #439 as a follow-up

PR #441, opened 2026-09-08
ci: deploy main to isolated demo environment

### LINKED ISSUE #440: ci: deploy main to an isolated demo-main environment
## Problem

Published releases deploy to the shared demo account, but commits on `main` have no continuously updated environment. Reusing the release deployment unchanged would also collide on Terraform state and account-global IAM/CloudFront names.

## Proposed change

- Keep release deployments on GitHub Environment `demo-release`.
- Deploy every immutable `main` commit SHA to GitHub Environment `demo-main`.
- Place the main deployment in `eu-central-1` while the release demo remains in `eu-west-1`.
- Keep Terraform `environment = \"prod\"` for production behavior, but use `project_name = \"collaborative-ai-dlc-main\"` for resource-name isolation.
- Use separate Terraform state (`main/terraform.tfstate`) and a distinct public hostname/CloudFront distribution.
- Validate that both release tags and main commit SHAs are reachable from `main` before requesting AWS credentials.
- Document the GitHub Environment, IAM/OIDC, state, regional, quota, credential, and OAuth prerequisites.

## Acceptance criteria

- [ ] A push to `main` invokes the reusable demo deployment with `${{ github.sha }}`.
- [ ] Release deployments continue to accept only annotated tags and target `demo-release`.
- [ ] `demo-main` uses a separate concurrency group, Terraform namespace, state key, and AWS region.
- [ ] The workflow uses GitHub OIDC and environment-scoped trust subjects.
- [ ] Regression coverage locks in release/main isolation and prevents long-lived AWS credentials.
- [ ] Operator documentation explains all required GitHub and AWS configuration.

JWThewes:
## Problem / Motivation

Published releases deploy to the demo account, but commits on `main` do not have a continuously updated environment. Reusing the release workflow and Terraform identifiers unchanged would collide on remote state and account-global IAM/CloudFront names.

## Why it matters

Without an isolated main environment, changes are only exercised in the shared deployment after a release is published. A second stack also needs explicit state, namespace, OIDC, and hostname isolation to avoid replacing or taking ownership of release resources.

## What changed

- Generalized the reusable demo workflow to validate either annotated release tags or immutable commit SHAs before requesting AWS credentials.
- Renamed the release workflow target to GitHub Environment `demo-release`.
- Added a `Deploy Main Demo` workflow that deploys `${{ github.sha }}` on every push to `main` using GitHub Environment `demo-main`.
- Kept Terraform `environment = "prod"` for production behavior while setting `project_name = "collaborative-ai-dlc-main"` for account-global name isolation.
- Added a backend-resolution job and deployment concurrency keyed by the actual Terraform state bucket/key. Terraform consumes the same captured backend; deployment summaries retain per-target attribution.
- Documented the `eu-central-1` main stack, `main/terraform.tfstate`, OIDC role trust, state access, quotas, credentials, OAuth callbacks, and hostname requirements.
- Documented the existing `AdministratorAccess` policy on the separate release/main deployment roles, with a reviewed narrower policy recommended as separate hardening work.
- Extended release workflow regression coverage for release/main isolation, backend concurrency, case-sensitive state keys, safe output encoding, and missing backend configuration.

## Tests

- `node --test scripts/test/*.test.mjs` — 51/51 passed.
- Focused release/main deployment regression passed after rebasing onto `origin/main`.
- Oxfmt check passed for all changed files.
- Both workflow files parsed successfully as YAML.
- Oxlint completed with 0 errors (existing repository warnings only).
- Secretlint passed.
- Full Vitest startup reached the existing Gremlin Testcontainers global setup, then stopped because no local container runtime is available.

## Manual verification

AWS and GitHub resources are intentionally not mutated by this PR. Before merge, configure the `demo-main` GitHub Environment and its environment-scoped OIDC role/state variables as documented. The existing release role must trust `environment:demo-release`.

Backend resolution and deployment both reference the selected GitHub Environment, so required reviewers and wait timers apply to each job. The resolver uses `deployment: false`; environments with custom deployment protection GitHub Apps must use `deployment: true` there, as documented.

## Screenshots / video

N/A — CI workflow and operator documentation changes only; there is no rendered UI delta.

Closes #440

### DIFF
--- a/.github/workflows/deploy-demo.yml
+++ b/.github/workflows/deploy-demo.yml
@@ -156,7 +156,6 @@ jobs:
             AWS_REGION
             AWS_ROLE_ARN
             BEDROCK_MODEL
-            TF_PROJECT_NAME
             TF_STATE_BUCKET
             TF_STATE_KEY
             TF_STATE_REGION
--- a/.github/workflows/deploy-demo.yml
+++ b/.github/workflows/deploy-demo.yml
@@ -4,8 +4,28 @@ on:
   workflow_call:
     inputs:
       ref:
-        description: Immutable release tag to deploy
+        description: Immutable release tag or commit SHA to deploy
         required: true
+        type: string
+      ref_type:
+        description: How to validate ref (release-tag or commit)
+        required: false
+        default: release-tag
+        type: string
+      deployment_target:
+        description: Stable target identifier used for concurrency and reporting
+        required: false
+        default: release
+        type: string
+      github_environment:
+        description: Protected GitHub Environment containing deployment variables
+        required: false
+        default: demo-release
+        type: string
+      terraform_project_name:
+        description: Project namespace used to isolate account-global AWS resource names
+        required: false
+        default: collaborative-ai-dlc
         type: string
   workflow_dispatch:
     inputs:
@@ -19,7 +39,7 @@ permissions:
   id-token: write
 
 concurrency:
-  group: deploy-demo
+  group: deploy-demo-${{ inputs.deployment_target || 'release' }}
   cancel-in-progress: false
 
 jobs:
@@ -27,7 +47,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 180
     environment:
-      name: demo
+      name: ${{ inputs.github_environment || 'demo-release' }}
       url: ${{ steps.deployment.outputs.url }}
     env:
       AWS_ACCOUNT_ID: ${{ vars.AWS_ACCOUNT_ID }}
@@ -35,7 +55,10 @@ jobs:
       AWS_ROLE_ARN: ${{ vars.AWS_ROLE_ARN }}
       BEDROCK_MODEL: ${{ vars.BEDROCK_MODEL }}
       DEPLOY_REF: ${{ inputs.ref }}
+      DEPLOY_REF_TYPE: ${{ inputs.ref_type || 'release-tag' }}
+      DEPLOY_TARGET: ${{ inputs.deployment_target || 'release' }}
       TF_ENVIRONMENT: prod
+      TF_PROJECT_NAME: ${{ inputs.terraform_project_name || 'collaborative-ai-dlc' }}
       TF_INPUT: 'false'
       TF_IN_AUTOMATION: 'true'
       # Clean runners do not have archives recorded in remote Terraform state.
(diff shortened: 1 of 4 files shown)


docs/development/demo-deployment.md
```diff
@@ -202,16 +312,17 @@ and policy management, `iam:PassRole`, Lambda, API Gateway, Cognito, EC2,
 Elastic Load Balancing, ECS, ECR, S3, DynamoDB, Neptune, CloudFront, CloudWatch,
 EventBridge, SQS, Secrets Manager, Systems Manager, and Bedrock AgentCore.
 
-Do not attach `AdministratorAccess` as a bootstrap policy. Keep deployment
-disabled until a reviewed customer-managed policy is available; an approval
-mistake must not grant the workflow unrestricted control of the account.
+A reviewed customer-managed policy is recommended. AWS-managed
+`AdministratorAccess` gives the workflow unrestricted control of the account;
+existing demo accounts that deliberately retain it should treat that as an
+accepted risk and attach the AWS-managed policy ARN deliberately.
```
jeromevdl: This is a real gap wth previous version, why such a change ?

.github/workflows/deploy-demo.yml
```diff
@@ -100,13 +156,14 @@ jobs:
             AWS_REGION
             AWS_ROLE_ARN
             BEDROCK_MODEL
+            TF_PROJECT_NAME
```
jeromevdl: Minor:
` TF_PROJECT_NAME: ${{ inputs.terraform_project_name || 'collaborative-ai-dlc' }}`

TF_PROJECT_NAME is set so cannot be empty.

.github/workflows/deploy-demo.yml
```diff
@@ -19,23 +39,26 @@ permissions:
   id-token: write
 
 concurrency:
-  group: deploy-demo
+  group: deploy-demo-${{ inputs.deployment_target || 'release' }}
```
jeromevdl: Concurrency is done on `deployment_target` but what actually collides is Terraform state: `TF_STATE_KEY` / `project_name`. Those decide which resources a run touches.

**Problem**: two callers with different `deployment_target` but same state key/project_name (e.g. someone copies the caller and renames the target) land in separate concurrency groups. Both start at once. Both target same state. Only the S3 backend lock (`use_lockfile=true`) saves you — second run fails on lock contention instead of queueing cleanly.                                                                                                                                                                                                                                                                                   


```suggestion
  group: deploy-demo-${{ inputs.terraform_state_key }}   # or project_name                                                                                                                                                                                                                                                
```

Same state = same group = clean serialize. Different state = parallel, safe. cancel-in-progress: false so a queued apply waits instead of getting killed mid-run

jeromevdl (commented):
Few things, especially one question on the `AdministratorAccess`

docs/development/demo-deployment.md
JWThewes: I checked the live AWS and GitHub configuration: the release deployment already uses `AdministratorAccess` on `CollaborativeDemoGitHubDeploy`. The main environment uses the same managed policy on `CollaborativeMainDemoGitHubDeploy`, with each role trusting its corresponding GitHub Environment.

Updated the docs in 2bdce4a to explain this existing configuration and remove the assumption that a reviewed customer-managed deployment policy is already in use. Creating IAM roles or CloudFront resources does not inherently require `AdministratorAccess`; a reviewed, narrower deployment policy remains the recommended approach and separate hardening work. The docs also make clear that separate state keys and project names do not reduce the permissions granted by `AdministratorAccess`.

This commit only corrects the documentation; it does not change AWS permissions.

.github/workflows/deploy-demo.yml
JWThewes: Agreed. Removed `TF_PROJECT_NAME` from the required-variable list in 57d0cd0. The non-empty fallback and the earlier project-name format validation remain in place.

The release/main deployment regression, YAML parsing, formatting, secret scan, and whitespace checks passed.

.github/workflows/deploy-demo.yml
JWThewes: Addressed in e13d3b7. Deployment concurrency now uses a SHA-256 hash of the actual Terraform state bucket/key pair, with `cancel-in-progress: false`. Callers sharing a backend share the group regardless of their target label or project name; different buckets or keys get different groups. Hashing also preserves distinctions between case-sensitive S3 keys despite GitHub's case-insensitive concurrency groups.

The backend comes from GitHub Environment variables, which are only available after a job starts. A small `resolve-backend` job therefore captures the backend and publishes the group before the deployment job is scheduled. Terraform consumes that same captured backend, so the group and the state it actually uses cannot drift apart. The resolver has no GitHub token permissions and does not request AWS credentials.

The docs explain that environment protection rules apply to both backend resolution and deployment, including the two approval stages when reviewers are required. They also document the existing latest-pending-run queue behavior and the configuration needed for custom deployment protection apps.

All 51 script tests passed, including shared-state grouping, distinct buckets/keys, key casing, safe output encoding, and missing configuration. YAML parsing, formatting, JavaScript lint, secret scanning, and whitespace checks also passed.

jeromevdl (approved):

PR #442, opened 2026-09-08
fix(intents): log the caught error object in the handler catch

sigent-ai-dev:
## Summary

The outer `catch` in the intents Lambda handler used the bare-catch form (`catch {}`) and logged only a static `console.error('intents handler error')`, discarding the caught exception. Every 500 on any `/api/projects/*/intents/*` endpoint appeared in CloudWatch as an identical opaque string — no stack, no code, no path context — so an entire class of production 500s was undiagnosable.

Bind the caught exception and log its `message`, `name`, `code`, and `stack`, plus API-Gateway request context (`resource`, `httpMethod`, `projectId`, `intentId`). The 500 response body is unchanged, so the public contract is preserved.

Before:

\`\`\`
2026-09-07T15:20:00.706Z <rid> ERROR intents handler error
\`\`\`

After:

\`\`\`
2026-09-07T15:20:00.706Z <rid> ERROR intents handler error {
  message: '...', name: '...', code: '...', stack: '...',
  resource: '/api/projects/{projectId}/intents/{intentId}',
  httpMethod: 'GET', projectId: '...', intentId: '...'
}
\`\`\`

Motivation: on a real deployment I observed 1,638 500s over a ~45-minute window, all of which rendered as the same log line, forcing a diff of the source to even identify which catch block was firing. This lets operators see the actual failure without needing to re-run against a debugger.

## Test plan

- [x] `lambda/intents` vitest suite: **213/213 passing**.
- [x] `node --check lambda/intents/index.js` — clean.
- [x] `oxfmt --check` — clean.
- [x] `oxlint` — clean.
- [x] Confirmed 500 body remains `{"error":"Internal server error"}` (unchanged public contract).

Change scope: a single `catch (error) {}` binding + a richer `console.error` payload. Same file, same handler, no new imports.

## Changelog

Entry added under `[Unreleased] → Fixed`.

### DIFF
--- a/lambda/intents/index.js
+++ b/lambda/intents/index.js
@@ -4509,8 +4509,22 @@ export const handler = async (event) => {
     }
 
     return response(405, { error: 'Method not allowed' });
-  } catch {
-    console.error('intents handler error');
+  } catch (error) {
+    // Bind the caught exception and log it with request context. Prior form
+    // was `catch {}` + a static `console.error('intents handler error')`, so
+    // every 500 in CloudWatch was an identical opaque string with no stack
+    // and no way to distinguish the offending path. The 500 response
+    // contract is preserved — the body still exposes only a generic message.
+    console.error('intents handler error', {
+      message: error?.message,
+      name: error?.name,
+      code: error?.code,
+      stack: error?.stack,
+      resource: event?.resource,
+      httpMethod: event?.httpMethod,
+      projectId: event?.pathParameters?.projectId,
+      intentId: event?.pathParameters?.intentId,
+    });
     return response(500, { error: 'Internal server error' });
   } finally {
     if (conn) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - Optional static egress for OAuth connectors, credential resolution, and seed-blocks through `lambda_vpc_scope = "public-egress"`, with NAT public IP outputs and addresses printed in the deployment summary for external allow-lists.
 - Configurable container runtime via `DOCKER_HOST` — any Docker-API-socket runtime works without requiring a container CLI (Podman, Rancher Desktop verified); Finch unsupported ([#420](https://github.com/aws-samples/sample-collaborative-ai-dlc/issues/420)).
 
+### Fixed
+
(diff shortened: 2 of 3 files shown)


lambda/intents/index.js
```diff
@@ -4509,8 +4509,22 @@ export const handler = async (event) => {
...
     return response(405, { error: 'Method not allowed' });
-  } catch {
-    console.error('intents handler error');
+  } catch (error) {
+    // Bind the caught exception and log it with request context. Prior form
+    // was `catch {}` + a static `console.error('intents handler error')`, so
+    // every 500 in CloudWatch was an identical opaque string with no stack
+    // and no way to distinguish the offending path. The 500 response
+    // contract is preserved — the body still exposes only a generic message.
+    console.error('intents handler error', {
```
JWThewes, line 4518: 🟡 **The new diagnostic contract has no regression test**

This change's only runtime behavior is the shape of the `console.error` call, but the PR changes no test and the existing outer-error tests assert only the 500 status or state rollback. Nothing verifies that an Error's message, name, code, and stack and the API Gateway context reach the log while the response body stays generic.

```js
console.error('intents handler error', {
  message: error?.message,
  name: error?.name,
  code: error?.code,
  stack: error?.stack,
  resource: event?.resource,
  httpMethod: event?.httpMethod,
  projectId: event?.pathParameters?.projectId,
  intentId: event?.pathParameters?.intentId,
});
```

**Why it matters:** The diagnostic payload is unasserted → a later refactor can drop the exception or request fields while the existing suite remains green → production 500s can silently become opaque again.

**Suggestion:** Add a focused handler test that forces the outer catch, spies on `console.error`, asserts the structured exception and request-context fields, and confirms the unchanged generic 500 response.

_[code-review-sage]_

JWThewes (changes requested):

JWThewes:
@sigent-ai-dev if you could just add a test. Otherwise looks great!

sigent-ai-dev:
Added a regression test for the diagnostic contract in `7f1920b`.

The test forces the outer catch via an orchestrator invoke that throws a distinctive `OrchestratorInvokeError` (with a `code` and `stack`), spies on `console.error`, and asserts:

- the response stays generic (`500` + `{"error":"Internal server error"}`),
- `console.error` is called with `'intents handler error'` plus a context object carrying `message`, `name`, `code`, `stack`, `resource`, `httpMethod`, `projectId`, `intentId`.

Locks the shape so a future refactor that drops the exception fields or the request context fails here instead of silently making 500s opaque again.

Placed next to the existing "rolls back to DRAFT when the orchestrator invoke fails" test — same 500-forcing mechanism, so no new mocking machinery.

Test count: 213 → 214, all passing.

lambda/intents/index.js
sigent-ai-dev: Added in 7f1920b. The new test forces this exact catch via an orchestrator invoke that throws `OrchestratorInvokeError` (custom `name` + `code`), spies on `console.error`, and asserts each field of the structured payload (message/name/code/stack + resource/httpMethod/projectId/intentId) plus the unchanged generic 500 body. 214/214 passing.

JWThewes (approved):
Looks good

PR #443, opened 2026-09-08
fix(terraform): raise v2-executions on-demand read ceiling to avoid hot-partition throttling

sigent-ai-dev:
## Symptom

A ~2h capture on a live deployment showed **2,861 API-Gateway 5xxs** concentrated on `intents/*` endpoints:

| count | endpoint |
|---|---|
| 1,443 | \`GET /api/projects/{projectId}/intents/{intentId}/outputs\` |
| 1,086 | \`GET /api/projects/{projectId}/intents/{intentId}\` |
| 262 | \`POST /api/projects/{projectId}/intents/{intentId}/realtime-token\` |
| 32 | \`GET /api/projects/{projectId}/intents/{intentId}/graph\` |
| 26 | \`GET /api/projects/{projectId}/intents/metrics\` |

Alongside, **7,609 DynamoDB `ReadThrottleEvents`** in the same window. The intents Lambda's outer catch now (post-#442) exposes the underlying error:

```
ProvisionedThroughputExceededException: The level of configured
  provisioned throughput for the table was exceeded.
    at Object.getExecution (...)
    at BufferedInvokeProcessor.handler (...)
resource: '/api/projects/{projectId}/intents/{intentId}/realtime-token',
httpMethod: 'POST',
```

## Root cause

Reads hit a single hot partition on \`collaborative-ai-dlc-v2-executions\` (\`pk = EXEC#<intentId>\`) whenever the intent-view page polls a live intent. The table is on \`PAY_PER_REQUEST\` for non-prod environments, which enforces a **3,000 RCU/s per-partition ceiling** and rejects sustained bursts above **2× historical peak** while adaptive capacity catches up. A handful of concurrent viewers on the same live intent (each polling \`/outputs\` and \`/intents/{id}\` on refresh) exceeds those thresholds; the AWS SDK retries a few times, then the exception bubbles to the handler as a 500.

The table itself is small (~269 items / 236 KB) — this is not a capacity problem, purely bursty concurrent reads on one partition key.

## Fix

Configurable \`on_demand_throughput\` on the \`v2-executions\` table and each of its three GSIs, defaulting to:

- \`max_read_request_units = 4000\`
- \`max_write_request_units = 1000\`

Exposed at the root as:

- \`v2_executions_max_read_request_units\` (default 4000; \`0\` opts out, adaptive-only)
- \`v2_executions_max_write_request_units\` (default 1000; \`0\` opts out)

Blocks are rendered conditionally so they are a **no-op on \`PROVISIONED\` (prod)** tables — \`read_capacity\` + autoscaling continues to govern there unchanged.

## Related

Complementary to #442, which made the underlying error diagnosable by binding the caught exception in the intents Lambda handler.

## Test plan

- [x] \`terraform init -backend=false\` — clean.
- [x] \`terraform validate\` at repo root — \`Success! The configuration is valid.\`
- [x] \`terraform validate\` on the changed module (\`terraform/modules/compute/agentcore\`) — \`Success!\`
- [x] \`terraform fmt -recursive -check\` — clean.
- [x] No changes to Lambda code, IAM, or any resource beyond the DynamoDB table + GSIs.

## Changelog

Entry added under \`[Unreleased] → Fixed\`.

### DIFF
--- a/terraform/modules/compute/agentcore/main.tf
+++ b/terraform/modules/compute/agentcore/main.tf
@@ -225,6 +225,21 @@ data "aws_ecr_image" "agentcore" {
 #   GSI3 = sparse maintenance index for active executions and parked PR waits
 # ---------------------------------------------------------------------------
 
+# On-demand throughput ceilings are only meaningful on PAY_PER_REQUEST tables
+# and only when a non-zero value is supplied. `on_demand_throughput` blocks are
+# rendered conditionally so a provisioned prod table (or a caller that opts
+# out with 0) sees no diff.
+locals {
+  v2_executions_on_demand = (
+    local.billing_mode == "PAY_PER_REQUEST"
+    && var.v2_executions_max_read_request_units > 0
+    && var.v2_executions_max_write_request_units > 0
+    ) ? [{
+      max_read_request_units  = var.v2_executions_max_read_request_units
+      max_write_request_units = var.v2_executions_max_write_request_units
+  }] : []
+}
+
 resource "aws_dynamodb_table" "v2_executions" {
   name           = "${var.project_name}-v2-executions-${var.environment}"
   billing_mode   = local.billing_mode
@@ -232,6 +247,14 @@ resource "aws_dynamodb_table" "v2_executions" {
   range_key      = "sk"
   read_capacity  = local.read_capacity
   write_capacity = local.write_capacity
+
+  dynamic "on_demand_throughput" {
+    for_each = local.v2_executions_on_demand
+    content {
+      max_read_request_units  = on_demand_throughput.value.max_read_request_units
+      max_write_request_units = on_demand_throughput.value.max_write_request_units
+    }
+  }
 
   attribute {
     name = "pk"
@@ -279,6 +302,13 @@ resource "aws_dynamodb_table" "v2_executions" {
       attribute_name = "GSI1SK"
       key_type       = "RANGE"
     }
+    dynamic "on_demand_throughput" {
+      for_each = local.v2_executions_on_demand
+      content {
+        max_read_request_units  = on_demand_throughput.value.max_read_request_units
+        max_write_request_units = on_demand_throughput.value.max_write_request_units
+      }
+    }
   }
 
   global_secondary_index {
@@ -294,6 +324,13 @@ resource "aws_dynamodb_table" "v2_executions" {
       attribute_name = "GSI2SK"
       key_type       = "RANGE"
     }
+    dynamic "on_demand_throughput" {
+      for_each = local.v2_executions_on_demand
+      content {
+        max_read_request_units  = on_demand_throughput.value.max_read_request_units
+        max_write_request_units = on_demand_throughput.value.max_write_request_units
+      }
+    }
   }
 
   global_secondary_index {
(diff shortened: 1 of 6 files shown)


terraform/modules/compute/agentcore/main.tf
```diff
@@ -233,6 +248,14 @@ resource "aws_dynamodb_table" "v2_executions" {
   read_capacity  = local.read_capacity
   write_capacity = local.write_capacity
 
+  dynamic "on_demand_throughput" {
```
JWThewes: 🔴 **Maximum-throughput caps cannot fix a hot partition**

AWS and the Terraform provider define `on_demand_throughput` as the maximum allowed request rate. Adding 4,000 read and 1,000 write units therefore caps the currently uncapped table and all three GSIs; it does not pre-warm them, raise historical warm throughput, or change the 3,000-RCU hard limit of the base-table partition serving these `GetItem` calls. The GSI caps are also unrelated to the reported base-table reads.

```hcl
dynamic "on_demand_throughput" {
  for_each = local.v2_executions_on_demand
  content {
    max_read_request_units  = on_demand_throughput.value.max_read_request_units
    max_write_request_units = on_demand_throughput.value.max_write_request_units
  }
}
```

**Why it matters:** Polling remains concentrated on the same partition -> requests above that partition's limit are still throttled -> the reported intent API 5xx responses persist; meanwhile the new table-wide and GSI maximums reject otherwise scalable aggregate traffic -> the deployment can gain additional read or write throttling.

**Suggestion:** Remove these maximum-throughput blocks and first record the exception's `ThrottlingReasons` and resource ARN. If the reason is insufficient warm throughput, configure and load-test `warm_throughput`; if it is `TableReadKeyRangeThroughputExceeded`, coalesce/cache or reduce the 8-second polling reads, move further toward event-driven updates, or redesign/shard the hot key. Apply settings only to the affected table or index.

_[code-review-sage]_

terraform/modules/compute/agentcore/main.tf
```diff
@@ -225,6 +225,21 @@ data "aws_ecr_image" "agentcore" {
 #   GSI3 = sparse maintenance index for active executions and parked PR waits
 # ---------------------------------------------------------------------------
 
+# On-demand throughput ceilings are only meaningful on PAY_PER_REQUEST tables
+# and only when a non-zero value is supplied. `on_demand_throughput` blocks are
+# rendered conditionally so a provisioned prod table (or a caller that opts
+# out with 0) sees no diff.
+locals {
+  v2_executions_on_demand = (
+    local.billing_mode == "PAY_PER_REQUEST"
```
JWThewes: 🟡 **Either zero value silently disables both throughput settings**

The public variables and PR description say each setting independently accepts `0` to opt out, and the provider permits either maximum by itself. This local emits the block only when both values are positive, so setting the write value to `0` also discards a positive read value, and vice versa.

```hcl
local.billing_mode == "PAY_PER_REQUEST"
&& var.v2_executions_max_read_request_units > 0
&& var.v2_executions_max_write_request_units > 0
```

**Why it matters:** An operator follows the documented independent opt-out contract -> the combined `&&` removes the entire block -> the still-requested setting is silently absent from the plan and the deployment behaves differently from its configuration.

**Suggestion:** Model read and write limits independently: emit the block when either is enabled and omit or explicitly remove each disabled attribute according to provider semantics. Add plan-level coverage for positive/positive, positive/zero, zero/positive, and zero/zero transitions.

_[code-review-sage]_

JWThewes:
Closing in favor of #444. We do not want to tune on-demand throughput ceilings; the intended fix is only to use PAY_PER_REQUEST for DynamoDB in every environment.

PR #459, opened 2026-09-11
fix: lost answers in human-gate races

jeromevdl:
## Issue

A human answer could arrive after the question grace window but before the CLI exited or the orchestrator bound its callback. The persisted answer was then either treated as no gate, allowing the stage to complete without delivering it, or misclassified as a callback conflict.

## Fix

- Keep a stage parked when it still owns a pending or already-decided gate, then resume the conversation with the persisted answer.
- Recover the answer-before-bind race using strongly consistent DynamoDB reads and strict stage, lane, and callback ownership checks.
- Share one legacy-compatible ownership predicate across runtime and orchestrator paths.
- Validate gate decisions against the supported answered, approved, and rejected statuses.

## Validation

- npm test: 158 files, 2,846 tests passed.
- Frontend: 77 files, 540 tests passed.
- TypeScript, formatting, lint-staged, secret scanning, and pre-commit checks passed.

### DIFF
--- a/lambda/v2-orchestrator/index.js
+++ b/lambda/v2-orchestrator/index.js
@@ -42,6 +42,7 @@ import {
   planSegments,
   stageInstanceId as planStageInstanceId,
 } from '../shared/v2-execution-plan.js';
+import { humanTaskMatchesOwner, isHumanTaskAnswerStatus } from '../shared/v2-process-keys.js';
 import { resolveSkipTo, skipTargetsFrom, resolveRecomposeSkips } from '../shared/stage-skip.js';
 import { broadcastToIntentChannel } from '../shared/ws-fanout.js';
 import { resolveRuntimeTarget } from '../shared/runtime-target.js';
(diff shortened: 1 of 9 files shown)


lambda/v2-orchestrator/index.js
```diff
@@ -846,23 +847,64 @@ const handler = async (event, ctx, deps = defaultDeps()) => {
...
+            gateAfterBindFailure?.callbackId == null ||
+            gateAfterBindFailure.callbackId === callbackId;
+          const callbackOwnerCompatible =
+            gateAfterBindFailure?.callbackOwner == null ||
+            gateAfterBindFailure.callbackOwner === expectedCallbackOwner;
+          answeredEarly =
+            isHumanTaskAnswerStatus(gateAfterBindFailure?.status) &&
+            ownsExpectedStage &&
+            callbackIdCompatible &&
+            callbackOwnerCompatible;
```
JWThewes, line 898: **[P2] Preserve attached steering before resuming an early answer**

An `answered` status does not mean the answer request has finished persisting its payload. In `lambda/intents/index.js`, `answerHumanTask` saves the HUMAN decision before `createSteering` saves the optional course correction. This new recovery branch can observe the decision between those writes and resume immediately, while `run-stage` consumes pending steering once at entry.

I reproduced this with a deterministic test delaying the STEER write after the answer CAS: the resumed stage consumed an empty steering list, and the answer endpoint subsequently returned 200. The agent can therefore proceed without guidance submitted alongside the answer; that guidance may arrive at a later stage or never reach the agent if this was the final stage.

Please make the complete answer, including attached steering, available before treating it as resumable, and cover this interleaving in a regression test.

JWThewes (changes requested):
The change is focused, and the shared ownership predicate fits the existing runtime/orchestrator patterns. I ran six relevant suites (579 tests passed) and an additional durable replay check confirming that an unused early-answer callback does not prevent completion.

Requesting changes for the answer/steering ordering issue below. The new recovery path can resume before the course correction attached to the same answer has been persisted. Please make the complete answer available before early resume and add a regression test for that interleaving.

JWThewes (approved):

PR #461, opened 2026-09-13
feat(terraform): harden data stores with CMK encryption, PITR, and deletion protection

jeromevdl:
## Summary

In-place data-store hardening: customer-managed encryption support, point-in-time recovery, deletion protection, and safer teardown. Safe for existing deployments with **zero resource re-creation and zero data loss**; new installs get the hardened posture at first deploy.

A `terraform plan` on an existing (unencrypted-Neptune, AWS-owned-DynamoDB) state shows only `~ update` / `+ create` (the optional KMS key) — never `-/+ replace`.

## What's included

- **KMS** — new `terraform/modules/kms` with three modes:
  - `default` (backward-compatible AWS-owned key, no KMS admin needed) — module and example default
  - `create` (Terraform provisions one shared CMK; rotation on, 30-day deletion window)
  - `existing` (BYOK via `kms_key_arn`)

  The key policy authorizes the **DynamoDB** and **RDS (Neptune)** service principals through `kms:ViaService`, scoped to this account — one shared key covers both stores.
- **DynamoDB** — SSE (CMK-capable) on all tables; PITR + deletion protection on all durable tables (`sessions`, `notifications`, `agent_questions`, `agent_outputs`, `blocks`, `environment_registry`, `discussion_read_state`, `yjs_documents`, the `git/` tables, and `v2_executions`).
- **Neptune** — deletion protection, `backup_retention_period` (default 7), and a **unique** `final_snapshot_identifier` (random suffix) so repeated deploy/destroy cycles never collide on the snapshot name. Prod cannot skip the final snapshot.
- **S3** — `force_destroy` gated on non-prod for application and frontend buckets.
- **Teardown** — `destroy.sh` and `install.sh` refuse an effective `prod` environment (including a differently named tfvars whose `environment` value is `prod`), and clear deletion protection via a two-phase apply whose target list is **cross-checked by tests** to prevent drift. Break-glass procedure documented.
- **Docs** — prerequisites (KMS permissions, `kms_key_arn` for existing keys), setup (prod must be named exactly `prod`; break-glass teardown ordering), README.
- **Tests** — new `scripts/test/terraform-hardening.test.mjs`; new prod-refusal cases in `release-tools.test.mjs`.

## Not included

- **Neptune encryption-at-rest** — the only replacement-forcing change; needs an opt-in variable plus a snapshot→restore migration, handled separately.

## Test plan

- [x] `terraform validate` clean; `terraform fmt` clean
- [x] `terraform-hardening.test.mjs` — 4/4 pass
- [x] `release-tools.test.mjs` prod-refusal + plan-guard cases pass
- [x] `terraform plan` against a representative existing state confirms zero `-/+ replace`


terraform/main.tf
```diff
@@ -305,6 +342,21 @@ module "vpc_endpoints" {
   }
 }
 
+# Optional shared customer-managed key for DynamoDB now and Neptune after the
+# opt-in encryption migration. The default mode creates no resources and
+# preserves the service-owned key posture of existing deployments.
+module "data_kms" {
+  count  = var.kms_mode == "create" ? 1 : 0
```
JWThewes: **[P1] Keep KMS mode changes from deleting a key still protecting data**

Changing from `kms_mode = "create"` to `"existing"` using the same key ARN sets this module's count to zero. In a mocked plan, Terraform deletes the key and alias while leaving the tables using that key unchanged, and `scripts/inspect-terraform-plan.mjs` accepts the plan. Scheduling key deletion makes the key unusable during the waiting period, so this can break table access. Switching to `"default"` also deletes the old key, which existing backups may still require for restoration. Please separate key retirement from mode selection and include KMS key deletion in the existing persistent-resource plan safeguards.

scripts/destroy.sh
```diff
@@ -39,6 +39,21 @@ if [[ ! -f "$BACKEND_FILE" ]]; then
     exit 1
 fi
 
+TFVARS_ENVIRONMENT="$(
+    awk -F= '$1 ~ /^[[:space:]]*environment[[:space:]]*$/ {
+        value = $2
+        sub(/#.*/, "", value)
```
JWThewes: **[P1] Resolve the production environment without partially parsing HCL**

For an environment alias such as `live`, valid HCL like `environment = "prod" // production` evaluates to `prod` in Terraform but passes this guard because only `#` comments are stripped. A commented-out `environment = "dev"` assignment inside a block comment before the real production assignment also bypasses the check. I reproduced both cases with Terraform console and a safe Terraform command stub: the script invokes both the protection-disabling apply and destroy. Please resolve the effective environment through Terraform or a proper HCL parser before any mutation, and cover these valid syntax cases in the refusal tests.

scripts/destroy.sh
```diff
@@ -62,8 +77,36 @@ terraform -chdir="$TF_DIR" state pull > "$BACKUP_FILE"
...
+    -target=module.dynamodb.aws_dynamodb_table.environment_registry
+    -target=module.dynamodb.aws_dynamodb_table.discussion_read_state
+    -target=module.dynamodb.aws_dynamodb_table.yjs_documents
+    -target=module.git.aws_dynamodb_table.git_connections
+    -target=module.git.aws_dynamodb_table.git_provider_connections
+    -target=module.git.aws_dynamodb_table.source_control_bindings
+    -target=module.git.aws_dynamodb_table.tracker_connections
+    -target=module.agentcore.aws_dynamodb_table.v2_executions
+)
+terraform -chdir="$TF_DIR" apply \
```
JWThewes: **[P2] Avoid recreating missing resources while preparing teardown**

This targeted apply reconciles complete resources, not just their deletion-protection attributes. After a partial deployment or partially successful destroy, retrying the script creates missing tables, Neptune resources, and their dependencies before trying to delete them again. A mocked plan with these targets and missing resources confirmed creation actions. This adds provisioning requirements and failure points to a cleanup path that previously only destroyed resources; any failed creation also prevents the subsequent destroy because the script uses `set -e`. Please constrain preparation to protection updates on existing resources and reject creates, replacements, and unrelated changes before applying.

JWThewes (changes requested):
Requesting changes for the three resource-lifecycle and teardown issues noted inline: KMS mode changes can delete a key still protecting data, valid HCL syntax can bypass the production teardown refusal, and retrying teardown can recreate resources that were already removed.

The optional encryption defaults, module boundaries, and unchanged resource identities are otherwise sensible and consistent with existing patterns.

Validation: all 51 relevant script tests passed; Terraform formatting, validation, and shell syntax checks passed. A mocked base-to-PR upgrade of the DynamoDB and Neptune modules showed no replacements. The findings were reproduced with mocked Terraform plans or a safe Terraform command stub; no live AWS deployment was exercised.

jeromevdl:
@JWThewes I addressed the three lifecycle/teardown findings in 968232d7.

- I actually removed the creation of a CMK. Users can provide their own key or use the default encryption. Users generally use their own KMS key anyway, handled by a dedicated terraform. It simplifies the lifecycle.
- I replaced the partial awk parsing with Terraform evaluation of the effective environment value before any AWS mutation. Regression coverage includes // comments and block-commented assignments.
- Teardown now intersects deletion-protection targets with terraform state list, creates a saved preparation plan, and inspects its JSON before applying it. Only in-place true → false deletion-protection updates are accepted; creates, replacements, deletions, and unrelated updates are rejected. Resources already removed by a partial teardown are skipped, while stale/inconsistent state still fails safely.

Targeted regression tests, formatting/syntax checks, and terraform validate pass. No live AWS deployment.

scripts/destroy.sh
```diff
@@ -39,6 +39,31 @@ if [[ ! -f "$BACKEND_FILE" ]]; then
...
+TEMP_DIR="$(mktemp -d "${TMPDIR:-/tmp}/aidlc-destroy.XXXXXX")"
+cleanup() {
+    rm -rf "$TEMP_DIR"
+}
+trap cleanup EXIT
+
+cp "$TF_DIR/variables.tf" "$TEMP_DIR/variables.tf"
+if ! EFFECTIVE_ENVIRONMENT_IS_PROD="$(
+    printf '%s\n' 'var.environment == "prod"' |
+        terraform -chdir="$TEMP_DIR" console -var-file="$TFVARS_FILE" 2>"$TEMP_DIR/console.stderr"
```
JWThewes: **[P1] Evaluate the production guard with the same Terraform variable inputs as teardown**

@jeromevdl The original HCL comment cases are fixed, but copying only `variables.tf` into this temporary directory drops Terraform's automatically loaded variable files (`terraform.tfvars`, `terraform.tfvars.json`, `*.auto.tfvars`, and `*.auto.tfvars.json`). The subsequent plan/destroy runs in the real Terraform directory and can therefore resolve a different environment.

Reproduction:

1. Set `environment = "prod"` in `terraform/production.auto.tfvars`.
2. Leave `environment` out of `terraform/environments/live.tfvars` (for example, put only `aws_region = "us-east-1"` there), with a matching `live.s3.tfbackend`.
3. Run `scripts/destroy.sh live --yes`.

Terraform evaluates the actual configuration as production, but this guard sees the default `dev` and allows the script to reach both the protection-disabling apply and destroy. I reproduced this with Terraform 1.15.0 for variable evaluation and safely stubbed mutation commands; no AWS resources were touched.

Please make the guard resolve the same effective environment as the subsequent Terraform commands, preserving all applicable variable sources and their precedence, and fail closed if it cannot establish that equivalence. Add a regression for production supplied through an automatically loaded variable file while the explicit environment file omits it. That regression should exercise real Terraform input loading; the current tests stub the console result and cannot catch this mismatch.

scripts/destroy.sh
```diff
@@ -39,6 +39,33 @@ if [[ ! -f "$BACKEND_FILE" ]]; then
...
 
+TEMP_DIR="$(mktemp -d "${TMPDIR:-/tmp}/aidlc-destroy.XXXXXX")"
+cleanup() {
+    rm -rf "$TEMP_DIR"
+}
+trap cleanup EXIT
+
+if ! EFFECTIVE_ENVIRONMENT_IS_PROD="$(
+    printf '%s\n' 'var.environment == "prod"' |
+        terraform -chdir="$TF_DIR" console \
```
JWThewes, line 53: **[P2] Initialize Terraform before evaluating the teardown guard**

@jeromevdl The production bypass is fixed. One small follow-up: `terraform console` now runs before the script's `terraform init`. In a fresh checkout with valid dev tfvars/backend files, `scripts/destroy.sh dev --yes` exits with `Backend initialization required` before reaching initialization. I reproduced this with Terraform 1.15.0; no AWS resources were touched.

Please initialize the selected backend/modules before evaluating the guard, keeping production refusal before destructive operations, and add a regression with a backend configuration—the new fixture copies only `variables.tf` and misses this case.

Minor documentation cleanup: the PR description still advertises KMS key creation and three modes, which have been removed.

scripts/destroy.sh
```diff
@@ -39,6 +39,36 @@ if [[ ! -f "$BACKEND_FILE" ]]; then
...
+trap cleanup EXIT
+
+echo "Initializing Terraform for environment: $ENVIRONMENT"
+terraform -chdir="$TF_DIR" init -reconfigure -backend-config="$BACKEND_FILE"
+
+if ! EFFECTIVE_ENVIRONMENT_IS_PROD="$(
+    printf '%s\n' 'var.environment == "prod"' |
+        terraform -chdir="$TF_DIR" console \
+            -var-file="$TFVARS_FILE" \
+            -var="deletion_protection=false" \
```
JWThewes: **[P1] Resolve command-specific Terraform variable inputs before allowing teardown**

The console guard still need not receive the same variables as plan/destroy. Terraform applies `TF_CLI_ARGS_<command>` only to the named command, so production supplied through `TF_CLI_ARGS_plan` and `TF_CLI_ARGS_destroy` is invisible here.

Reproduction:

1. Use an environment alias such as `live`, with matching backend settings, and leave `environment` out of `environments/live.tfvars` (for example, include only `aws_region`).
2. Set `TF_CLI_ARGS_plan='-var=environment=prod'` and `TF_CLI_ARGS_destroy='-var=environment=prod'`.
3. Run `scripts/destroy.sh live --yes`.

The guard evaluates the default `dev` and returns `false`, while the subsequent commands resolve `prod`. I reproduced this with real Terraform 1.15.0 and a local backend: the saved plan reported `variables.environment.value = "prod"`, and the script completed destruction of a local-only marker resource. No AWS resources were touched.

Please establish one effective variable set for the guard, preparation, and destroy, or fail closed on command-specific variable overrides that cannot be reconciled. Add a real-Terraform regression for this case; the current auto-tfvars fixture explicitly clears these environment variables.

Terraform documents the command-specific behavior here: https://developer.hashicorp.com/terraform/cli/config/environment-variables#tf_cli_args-and-tf_cli_args_name

terraform/modules/compute/agentcore/main.tf
```diff
@@ -313,6 +314,15 @@ resource "aws_dynamodb_table" "v2_executions" {
     }
   }
 
+  server_side_encryption {
+    enabled     = var.kms_key_arn != ""
+    kms_key_arn = var.kms_key_arn != "" ? var.kms_key_arn : null
+  }
```
JWThewes: **[P2] Complete runtime KMS authorization when enabling BYOK**

Supplying `kms_key_arn` enables customer-managed encryption, but the generated AgentCore role—and the other application roles accessing these DynamoDB tables—has no `kms:Decrypt` permission. With a standard external key policy that delegates authorization to IAM, an authorized deployment role can configure encryption while the application roles remain unable to decrypt table keys. DynamoDB performs these decrypt requests on behalf of the caller; service-principal access alone does not authorize those application roles.

An external key policy that directly grants the callers access can avoid this, but the prerequisite currently only says to delegate the required DynamoDB service access and does not describe the runtime-principal requirement.

Please either add conditional, key-scoped runtime permissions (restricted to use through DynamoDB) for the affected callers, or document and test an explicit external key-policy contract that grants those principals access. Cover the runtime authorization as well as the deployment principal's ability to enable encryption.

This finding comes from inspection of the generated IAM policies and AWS's documented authorization model, not a live KMS integration test: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/encryption.usagenotes.html

JWThewes (changes requested):
Requesting changes for the remaining production-guard input mismatch and BYOK runtime-authorization gap described inline.

The initialization-order fix and the earlier KMS-lifecycle and resource-recreation fixes address the previously reproduced cases.

Validation on 590c14c: all 54 relevant release-tooling and Terraform-hardening tests passed; Terraform formatting and shell syntax checks passed; Terraform validation passed with temporary placeholder backend settings. The production-guard bypass was reproduced with real Terraform 1.15.0, a local backend, and a local-only terraform_data resource. The KMS finding is based on the generated IAM policies and AWS documentation. No live AWS deployment or KMS integration test was performed.

scripts/destroy.sh
```diff
@@ -39,6 +39,49 @@ if [[ ! -f "$BACKEND_FILE" ]]; then
     exit 1
 fi
 
+# Terraform applies command-specific CLI arguments only to the named command.
+# Variable inputs injected this way could make the guard evaluate a different
+# environment than plan or destroy. Harmless command-specific flags remain
+# supported, but command-specific variable overrides fail closed.
+for cli_args_name in TF_CLI_ARGS_plan TF_CLI_ARGS_destroy; do
+    cli_args_value="${!cli_args_name-}"
+    if [[ "$cli_args_value" == *"-var"* ]]; then
```
JWThewes: **[P1] Close the remaining production-guard input bypasses**

The new check still allows the guard and teardown to resolve different environments in two cases:

1. **Console-specific override:** set `environment = "prod"` in `terraform/production.auto.tfvars`, leave `environment` out of `environments/live.tfvars`, and set `TF_CLI_ARGS_console='-var=environment=dev'`. Running `scripts/destroy.sh live --yes` makes the guard return `false`, while destroy uses `prod`. `TF_CLI_ARGS_console` is not checked here.
2. **Quoted variable flags:** leave `environment` out of `environments/live.tfvars` so the guard uses the default `dev`, and set both `TF_CLI_ARGS_plan` and `TF_CLI_ARGS_destroy` to `'-v"ar"=environment=prod'`. The raw `-var` substring check misses this, but Terraform parses the quoted flag as `-var`. The saved plan reports `variables.environment.value = "prod"`, and the script proceeds with destruction.

I reproduced both cases on `1a41c0e` with real Terraform 1.15.0, a local backend, and a local-only `terraform_data` marker. In each case the script exited successfully after destroying the marker. No AWS resources were touched.

Please consistently reject or normalize command-specific variable inputs across console, plan, and destroy, accounting for Terraform's argument quoting rather than checking only the raw substring. Add real-Terraform regressions for both cases.

JWThewes (changes requested):
Requesting changes for the remaining production-teardown guard bypass described inline. Both the unchecked console-specific variable input and quoted plan/destroy variable flags were reproduced on the current head.

The KMS runtime-authorization, initialization-order, key-lifecycle, and partial-teardown fixes appear to address the other findings.

Validation on 1a41c0e: all 56 relevant release-tooling and Terraform-hardening tests passed; Terraform formatting, validation, and shell syntax checks passed. Both remaining bypasses were reproduced using Terraform 1.15.0 and local-only resources. No live AWS deployment or KMS integration test was performed.

PR #465, opened 2026-09-14
fix(agentcore): grant dynamodb:ConditionCheckItem for workflow checkpoints

jeromevdl:
## Problem

Workflow checkpoint writes fail with:

```
User: ...assumed-role/collaborative-ai-dlc-agentcore-<env>/... is not authorized to
perform: dynamodb:ConditionCheckItem on resource: .../table/collaborative-ai-dlc-v2-executions-<env>
because no identity-based policy allows the dynamodb:ConditionCheckItem action
```

## Root cause

`putWorkflowCheckpoint` (`lambda/shared/v2-process-store.js`) issues a `TransactWriteCommand` whose first item is a `ConditionCheck` on the execution META row. DynamoDB maps that operation to the `dynamodb:ConditionCheckItem` action.

The agentcore role policy (`terraform/modules/compute/agentcore/main.tf`) granted `GetItem`/`PutItem`/`UpdateItem`/`Query`/`Scan` on the executions table but **not** `ConditionCheckItem`. The `Put` half of the transaction passed; the `ConditionCheck` half was denied, so the whole transaction aborted.

Introduced in #403 (2026-09-09), which added the checkpoint transaction but did not update the agentcore IAM policy. Latent since then — surfaces whenever a checkpoint write fires.

## Impact

Bounded. The transaction is atomic, so a denied `ConditionCheck` means nothing is written and the prior checkpoint stays intact. `createWorkflowCheckpoint` swallows the error (`{ ok: false, reason: 'checkpoint_failed' }`), so no run crash and no data loss — only workflow-export/rewind checkpoints stopped advancing. All other executions-table operations were unaffected, which is why it went unnoticed.

## Fix

- Add `dynamodb:ConditionCheckItem` to the agentcore role's executions-table statement.
- Add a regression test asserting the policy grants it (mirrors the existing managed-environments `tool_control` assertion).

Requires redeploying the agentcore role.


JWThewes (approved):
Looks good to me!

PR #466, opened 2026-09-14
fix: recover Kiro intents after quota exhaustion and lost sessions

### LINKED ISSUE #463: [Bug]: Recover existing intents after Kiro credit exhaustion and lost sessions
An existing Kiro-backed intent can remain stuck after the provider key runs out of credits and is replaced. New intents work, but the original intent can fail with `resume_no_session`, `unit_not_found`, and `orchestrator_error: gate_callback_conflict`.

The report concerns current main. Inspection baseline: `bec101f4942189654e7ef4b3a9ddbc2824d602d7`. Deployment records are unavailable, so the precise production sequence is not yet established.

Reproduction and confirmed gaps:

1. Answer an engine gate after it is opened/broadcast but before `setGateCallbackId` binds its callback. The pending-only conditional write fails and `awaitEngineGate` throws “already has a different callback owner”, even when no callback owns the gate. Stage-gate binding uses the same pending-only condition and needs the same race treatment.
2. Kiro session-ID capture and store persistence are best-effort. A stage can be parked with no usable persisted conversation. Its next answer takes the `resume_no_session` path instead of a controlled recovery from saved state.
3. Explicit credit exhaustion is not classified separately from generic CLI failure. Current per-invocation credential resolution already reads the intent's pinned source; rotating a different source can make new intents work without changing the existing intent.
4. `getUnitPlan(...).catch(() => null)` turns storage failures into `unit_not_found`, and scheduling reads are eventually consistent. Missing session-row reads are similarly conflated with absent state.

Expected behavior: preserve saved artifacts and answers, consume racing answers exactly once without stealing callbacks, reject retired/mismatched resumes, expose an actionable credit error, and let a correctly parked intent recover with a fresh CLI conversation if its original session is unavailable. Storage failures must remain distinct from an actually missing unit.

Implementation checklist:

- [x] Race-safe engine and stage gate binding, with retirement fencing and regression tests.
- [x] Guard Kiro parking on session durability; recover legacy missing-session waits only with matching persisted stage/gate ownership.
- [x] Explicit credit-exhaustion classification and guidance at the pinned credential scope.
- [x] Consistent unit-plan reads and distinct storage-read failures.
- [x] Same-intent key-rotation regression coverage and a recovery guide covering retry, repair limits and existing-intent export (#403).

Recovery available today: export a compatible started intent through #403 without a Kiro model call; rotate the originally bound credential; retry the failed stage after checking the approved unit plan. Parallel repair has narrower eligibility than a general recovery action. Export is a one-way handoff and does not bundle source repositories or every runtime log/memory.

Separate follow-ups: provider balance alerts, reviewed in-flight configuration refresh, versioned plan reconciliation, and complete diagnostic backups.

Implementation submitted in #466. Local regression checks pass; PR review, CI and deployment remain.

JWThewes:
An existing Kiro intent can remain stuck after a credit outage even when its key has been replaced: a parked stage may lack a saved session, and an answer arriving before callback binding can be mistaken for a conflicting callback owner.

This change:

- Handles answers that race engine/stage callback binding, while retaining callback ownership checks and fencing retired engine runs.
- Rejects Kiro parks without a captured session and successful configured store persistence. A legacy parked stage with matching gate ownership can start a fresh conversation using its saved context and answer; reset, sibling and expired waits remain protected.
- Classifies explicit credit exhaustion separately from invalid credentials and transient throttling, with guidance and a settings link for the intent's pinned credential scope. Regression coverage verifies that rotating that scope takes effect in the same intent.
- Uses consistent recovery and unit-plan reads, surfaces storage failures separately from absent state, and documents export, stage retry and the limits of parallel repair.

Validation: 1,984 backend tests across the agentcore, v2-orchestrator, shared and intents projects pass on Node 24, including durable-runner race/replay and real Git integration tests. All 56 intent-page tests pass, as do frontend typecheck, staged formatting/lint/secret checks and the intents/orchestrator bundles. The commit hooks also pass (959 affected tests; both production dependency audits report zero vulnerabilities). Rebased without changing this patch onto main's separate checkpoint IAM fix #465; its regression test passes.

The reported production sequence is not verified because deployment state was unavailable. A genuinely mismatched approved unit plan still requires operator reconciliation. Provider balance alerts, in-flight configuration refresh and complete diagnostic backups remain separate follow-ups.

Fixes #463.


lambda/agentcore/commands/run-stage.js
```diff
@@ -1437,9 +1496,13 @@ export const runStage = async (
       cli = null;
     }
     resumeAnswer = reviewFeedbackPrompt || formatResumeAnswer(resumeGate);
+    if (recoverMissingKiroSession) {
+      const recoveryFailure = await recoverLostConversation();
```
jeromevdl, line 1500: Calling `recoverLostConversation()` here sets `demotedResume = true`. Later, line 2004 restores the durable Kiro store only when `!demotedResume`, because demotion previously meant that the store was known to be lost.

  In this new path, however, we only know that the session ID is missing; the durable store may still exist. After a container restart, this can start with an empty local store and `persistKiroStore()` will replace the durable store, potentially deleting other Kiro sessions.

  Please distinguish “missing session ID” from “store confirmed lost” and restore the existing store before starting this fresh conversation.

lambda/agentcore/commands/run-stage.js
```diff
@@ -2344,6 +2407,28 @@ export const runStage = async (
     unitSlug,
     sectionIndex,
   });
+  const failUnresumableGate = async (reason, detail) => {
+    await store.supersedeHumanTask({
```
jeromevdl, line 2411: This helper can be reached after `updateStageState` has already failed. Both cleanup writes are awaited without error handling, so if the same DynamoDB outage continues, `runStage` throws before reaching `fail(...)`. `run-stage-start` then reports the generic `stage_job_crashed` reason instead of
  `stage_park_persist_failed`.

Please make these cleanup writes best-effort (catch and log each failure) and always return the original structured failure. A regression test should cover the park write and subsequent cleanup writes all rejecting.

lambda/agentcore/commands/run-stage.js
JWThewes: Confirmed; fixed in 291833a. Kiro store restoration is now tracked separately from conversation demotion. A missing session ID still starts a fresh conversation, but restores the durable store before spawning; resumes that already attempted restoration do not restore twice.

Added a restart regression with an empty local store and an existing sibling session in the durable store. It reproduced the sibling-session loss before the fix and now verifies that both the sibling and newly captured session survive persistence.

lambda/agentcore/commands/run-stage.js
JWThewes: Confirmed; fixed in 322e51c. Each cleanup write now catches and logs its own failure, so gate retirement failing does not prevent execution cleanup or the original structured failure from being returned.

The regression makes the park write, both cleanup writes, and the FAILED-stage write reject. It reproduced the exception before the fix and now verifies the original stage_park_persist_failed result and failure broadcast, with both cleanup errors logged.

Validation for both fixes: all 186 tests across run-stage, kiro-store, and run-stage-start pass on Node 22; formatting, lint, secret scanning, and the production dependency audit pass (zero vulnerabilities). Docker is unavailable here, so these unit suites ran with Docker service startup disabled; the container-backed integration suite was not run.

lambda/v2-orchestrator/section.js
```diff
@@ -157,17 +158,22 @@ export const awaitEngineGate = async (
   const { store, broadcast, ids, runId } = toolkit;
   const { executionId, intentId, projectId } = ids;
   const humanTaskId = `eg-${name}-${runId}`;
+  const ownsRun = async () => {
+    const meta = await store.getExecution(executionId, { consistentRead: true });
+    return Boolean(meta) && (!runId || !meta.orchestratorRunId || meta.orchestratorRunId === runId);
+  };
+  if (!(await ctxArg.step(`gate-owner-${name}`, ownsRun))) return { superseded: true };
```
jeromevdl: Adding `gate-owner-${name}` before the existing `gate-pre-${name}` changes the durable step order. Existing executions replay steps by position and verify their names. They will expect `gate-pre-*` but receive `gate-owner-*`, causing the durable runtime to terminate them as non-deterministic.

The new `gate-answered-early-*` step creates the same compatibility issue later.

Preserve the existing durable step names and order. Fold the ownership check into an existing compatible step, or keep existing executions on the old function version while routing only new executions to the new version. Add a replay test using checkpoints created by the previous implementation.

lambda/agentcore/commands/run-stage.js
```diff
@@ -1414,17 +1437,54 @@ export const runStage = async (
...
+        'resume_state_unavailable',
+        'The saved stage or answer could not be read. Retry when storage is available; the saved conversation has not been reset.',
+      );
+    }
     if (resumeFrom && !resumeGate) return fail(stageInstanceId, 'gate_not_found', resumeFrom);
+    if (resumeGate?.status === 'superseded') return { ok: false, reason: 'retired' };
     if (resumeFrom && resumeGate.status === 'pending')
       return fail(stageInstanceId, 'gate_not_answered', resumeFrom);
-    const row = await store.getStage(executionId, stageInstanceId).catch(() => null);
+    if (
```
jeromevdl, line 1516: This check rejects every batch “request changes” resume.

A batch gate is created without a `unitSlug`, so its value is null. However, each lane resumes with its own `unitSlug`. The comparison becomes `null !== "lane-slug"`, causing every lane to fail with `resume_state_conflict`.

Only compare unitSlug when the gate has one, as already done for `stageInstanceId`. Alternatively, store the correct lane identity on lane-specific gates. Add an integration test that sends a batch “request changes” answer through the real `runStage` resume validation.

lambda/v2-orchestrator/section.js
```diff
@@ -233,21 +243,35 @@ export const awaitEngineGate = async (
       `gate_callback_conflict: ${humanTaskId} already has a different callback owner`,
     );
   }
-  await callbackPromise;
+  const answeredEarly = await ctxArg.step(`gate-answered-early-${humanTaskId}`, () =>
+    store.getHumanTask(executionId, humanTaskId, { consistentRead: true }),
+  );
+  if (answeredEarly?.status === 'pending') await callbackPromise;
```
jeromevdl: The new early-answer check still leaves a race. It can read the gate as pending just before the answer is saved (lambda/intents/index.js:2475). Meanwhile, the answer endpoint may use its older gate snapshot, which has no callback ID, and therefore skip the callback.

This line then waits for a callback that will never arrive.

Use the updated row returned by `answerHumanTask` in the answer endpoint and resume `answered.callbackId`, not the older `gate.callbackId`. Please add a test covering this exact ordering.

lambda/v2-orchestrator/section.js
```diff
@@ -233,21 +243,35 @@ export const awaitEngineGate = async (
...
+  // no second read; the unpark CAS below still fences cancellation or rewind.
   const gate = await ctxArg.step(`gate-after-${name}`, () =>
-    store.getHumanTask(executionId, humanTaskId).catch(() => null),
+    answeredEarly && answeredEarly.status !== 'pending'
+      ? answeredEarly
+      : store.getHumanTask(executionId, humanTaskId, { consistentRead: true }),
   );
   if (!gate || gate.status === 'superseded') return { superseded: true };
-  await ctxArg.step(`gate-unpark-${name}`, async () => {
+  const unparked = await ctxArg.step(`gate-unpark-${name}`, async () => {
```
jeromevdl: This durable step may run more than once. The first attempt can successfully change META from `WAITING` to `RUNNING`, but crash before saving its checkpoint.

When retried, the same update fails because META is already `RUNNING`. The code interprets that failure as “this run was replaced” and incorrectly retires a healthy execution.

After a conditional failure, read META again. Treat it as success when it is already `RUNNING`, still belongs to the same `orchestratorRunId`, and has no pending gate. Only return superseded when ownership was actually lost. Add a replay test for a committed update followed by a lost checkpoint.

lambda/agentcore/commands/run-stage.js
```diff
@@ -2419,7 +2531,14 @@ export const runStage = async (
         cli,
         cliSessionId,
       })
-      .catch(() => {});
+      .then(() => true)
+      .catch(() => false);
+    if (!parkSaved) {
```
jeromevdl, line 2661: This now fails the stage for every CLI when the final park write fails.

For Claude, the CLI and session ID were already saved when the stage started, and the pending gate was already found. The conversation can still be resumed. Previously, this final write was best-effort, but now a temporary storage failure retires a valid question and fails the stage.

Retry the park write and then read the stage row to verify what was saved. Only retire the gate when the required resume information is truly missing. Add a Claude test proving that a failed final park write does not destroy an otherwise resumable stage.

lambda/agentcore/commands/run-stage.js
```diff
@@ -1414,17 +1437,54 @@ export const runStage = async (
...
     if (resumeFrom && resumeGate.status === 'pending')
       return fail(stageInstanceId, 'gate_not_answered', resumeFrom);
-    const row = await store.getStage(executionId, stageInstanceId).catch(() => null);
+    if (
+      resumeFrom &&
+      ((resumeGate.stageInstanceId != null && resumeGate.stageInstanceId !== stageInstanceId) ||
+        (resumeGate.unitSlug ?? null) !== unitSlug ||
+        (resumeGate.sectionIndex ?? null) !== sectionIndex ||
+        (row?.state != null && !['WAITING_FOR_HUMAN', 'SUCCEEDED'].includes(row.state)) ||
+        (row?.pendingHumanTaskId != null && row.pendingHumanTaskId !== resumeFrom))
```
jeromevdl: This guard rejects two valid resume situations:

- A retried durable dispatch can find the stage already RUNNING because its first attempt updated the row before the container stopped or the checkpoint was lost.
- A `SUCCEEDED` stage can retain an old pendingHumanTaskId because clearing that field is best-effort. The code allows `SUCCEEDED`, but then rejects it because of the stale pointer.
Both cases incorrectly return resume_state_conflict.

Allow `RUNNING` when the row still belongs to the same callback attempt. Only require `pendingHumanTaskId === resumeFrom` when resuming a `WAITING_FOR_HUMAN` stage; do not apply that requirement to a `SUCCEEDED` stage receiving later feedback. Add tests for both cases.

lambda/agentcore/commands/run-stage.js
```diff
@@ -2344,6 +2410,42 @@ export const runStage = async (
...
+        supersededBy: reason,
+      })
+      .catch((error) => {
+        console.error(
+          `[run-stage] failed to supersede unresumable gate ${parked.humanTaskId} stage=${stageInstanceId} reason=${reason}`,
+          error,
+        );
+      });
+    if (!unitSlug) {
+      await store.updateExecution({ executionId, pendingHumanTaskId: null }).catch((error) => {
```
jeromevdl: This update clears `META.pendingHumanTaskId` without checking which orchestrator run owns it.

A delayed container from an older run can finish after a rewind has started a new run. Because both runs use the same execution ID, the old container can clear the new run’s pending gate. Catching the error does not prevent this stale write.

Make the cleanup conditional on the expected `orchestratorRunId` and the expected `pendingHumanTaskId`. Also verify that the gate belongs to the same stage callback before superseding it. If ownership has changed, leave the new run untouched. Add a test simulating an old container completing after a rewind.

lambda/v2-orchestrator/section.js
```diff
@@ -216,11 +224,13 @@ export const awaitEngineGate = async (
     } catch {
       /* live fan-out is best-effort */
     }
+    return true;
   });
+  if (opened === false) return { superseded: true };
```
jeromevdl, line 240: Ownership can change after the initial ownership check.

The old run may create a pending human gate, then fail to update META because a new run has taken ownership. This line exits without retiring the gate that was just created.

The user can then see and answer a question that no execution is waiting for.

When the META ownership check fails, supersede the specific gate before returning. Preferably, create the gate and park META atomically in a DynamoDB transaction. Add a test where ownership changes between the initial check and the META update.

lambda/agentcore/commands/run-stage.js
```diff
@@ -1414,17 +1437,54 @@ export const runStage = async (
...
+    if (
+      resumeFrom &&
+      ((resumeGate.stageInstanceId != null && resumeGate.stageInstanceId !== stageInstanceId) ||
+        (resumeGate.unitSlug ?? null) !== unitSlug ||
+        (resumeGate.sectionIndex ?? null) !== sectionIndex ||
+        (row?.state != null && !['WAITING_FOR_HUMAN', 'SUCCEEDED'].includes(row.state)) ||
+        (row?.pendingHumanTaskId != null && row.pendingHumanTaskId !== resumeFrom))
+    ) {
+      return fail(
+        null,
```
jeromevdl: Passing null means the stage row is not marked `FAILED`. The execution becomes failed, but the stage can remain `WAITING_FOR_HUMAN`.

The UI only offers a targeted stage retry when it finds a `FAILED` stage row, so it falls back to restarting the complete workflow.

Ensure this failure leaves a retryable failed-stage record, using a conditional ownership check so a newer attempt cannot be overwritten. Add a test verifying that the UI offers “Retry stage” after `resume_state_unavailable` or `resume_state_conflict`.

lambda/agentcore/cli/credential-errors.js
```diff
@@ -13,4 +13,15 @@ export const isCredentialFailure = (output = '') => {
   return CREDENTIAL_FAILURE_PATTERNS.some((pattern) => pattern.test(text));
 };
 
+// A 429 alone is transient throttling. Only explicit credit/billing exhaustion
+// is actionable by replenishing or rotating the bound credential.
+export const isCreditExhaustion = (output = '') =>
```
jeromevdl: The new quota detector misses common provider messages such as:

Your credit balance is too low to access the Anthropic API
You exceeded your current quota usage limit reached

These errors fall through to `cli_nonzero_exit`, so the new “quota exhausted” recovery guidance is not shown.

Add patterns for low credit balances, exceeded quotas, and generic usage limits. Use real error messages captured from the supported CLIs as test fixtures.

`one-shot.js` also has inconsistent classification, but I would treat that separately rather than expanding this PR further.

jeromevdl (changes requested):
Overall, I think this PR needs another pass before merge.

The main blockers are the durable replay compatibility break, the deterministic batch “request changes” failure, and the remaining answer/unpark races.

The other recovery issues share the same root cause: ownership is checked differently across META, STAGE, HUMAN, and callback state. I would avoid adding more isolated guards. Please define one simple ownership invariant and make park, answer, unpark, and cleanup transitions conditional and idempotent:

  - a stale worker must do nothing;
  - retrying the same attempt must succeed;
  - a failed recovery must leave a retryable failed stage;
  - a gate must never be left pending without an owning execution.

The latest commits correctly resolve the Kiro store preservation and cleanup-exception findings. The quota classifier should also be tested with real provider messages.

Please add tests covering previous-version durable checkpoints, a batch revision through the real runStage validation, answer/bind interleavings, a committed write followed by a lost checkpoint, and an old worker completing after a rewind.

lambda/v2-orchestrator/section.js
JWThewes: Addressed in 98db835.

Fixed: the ownership check is inside the existing gate-pre step, and the early-answer observation is inside bind-callback. The deployed operation names/order are preserved. A frozen copy of the pre-PR implementation (590ace6) creates real durable checkpoints; the regression suspends that version and resumes with the current implementation, with exactly-once gate creation and binding.

lambda/agentcore/commands/run-stage.js
JWThewes: Addressed in 90e3b5b.

Fixed: unitSlug is compared only when the gate defines it. The regression creates a batch request-changes gate through awaitEngineGate and sends its answer through the real runStage resume validation for both auth and billing lanes, verifying that both receive the feedback and complete.

lambda/v2-orchestrator/section.js
JWThewes: Addressed in adcc7f7.

Fixed: the endpoint wakes answered.callbackId from the committed answer, rather than its initial gate snapshot. The real-handler regression explicitly orders the operations as: endpoint reads unbound gate; orchestrator binds; orchestrator reads pending; answer commits; endpoint delivers that callback. Coverage includes the transactional answer path introduced by the ownership fixes and an unbound sibling gate that must not be woken.

lambda/v2-orchestrator/section.js
JWThewes: Addressed in 4470288.

Fixed in one shared unpark helper used by both engine and stage gates. It conditions the write on the run and expected pending gate. On a conditional failure it reads META consistently and accepts already-RUNNING only for the same run with no pending gate. Real durable-runner regressions for both gate types commit the update, lose the step result, and retry successfully without retiring the run.

lambda/agentcore/commands/run-stage.js
JWThewes: Addressed in 534d4dc.

Fixed: the final park write is retried, then the stage is read consistently if both attempts fail. A correctly parked row with the same CLI, session and gate remains resumable; the question is preserved. An unavailable verification read preserves the saved question and returns the structured persistence failure. Claude regressions cover retry success and an already-saved conversation after lost acknowledgments.

lambda/agentcore/commands/run-stage.js
JWThewes: Addressed in 654359c.

Fixed: RUNNING is accepted only for the same stage callback attempt, and the pending-gate pointer is required only for WAITING_FOR_HUMAN. A SUCCEEDED stage can receive later feedback despite a stale pointer. Regressions cover both valid resumes and refusal of a different RUNNING callback; the ownership transaction also fences races after validation.

lambda/agentcore/commands/run-stage.js
JWThewes: Addressed in 40f7f35.

Fixed using a shared ownership invariant: META must name the dispatched orchestrator run and STAGE must name the worker callback. The dispatcher carries both identities into the worker/MCP scope. Stage mutations and cleanup check them transactionally; gate retirement also checks the gate's stage/callback, and clearing META checks the expected pending gate. Regressions let an old Kiro worker finish after a rewind and after a replacement callback in the same run: the new stage and gate are unchanged, and retired results cannot fail the replacement execution. The invariant and legacy/deployment limits are documented in docs/development/recovery-ownership.md.

lambda/v2-orchestrator/section.js
JWThewes: Addressed in e135e0e.

Fixed: current stores create the engine gate and park META in one DynamoDB transaction conditional on the run. The gate-open step also retires the specific legacy gate if ownership changes before the subsequent META check. A regression changes ownership between legacy gate creation and the META write and verifies retirement without changing the replacement gate. Owned gate answers are also conditional on the current run, preventing an abandoned question from waking a retired execution.

lambda/agentcore/commands/run-stage.js
JWThewes: Addressed in 2f0631a.

Fixed: the worker claims its callback with the run ownership check before reading recovery state, retaining the stage identity and saved conversation. For an owned attempt, resume_state_unavailable and resume_state_conflict now write FAILED conditionally rather than passing null. Backend regressions cover both read failures and conflicting state; IntentView tests verify that each error offers the targeted Retry Units Generation action and calls rewind instead of starting the full workflow. An unverified legacy dispatch still cannot overwrite a newer stage.

lambda/agentcore/cli/credential-errors.js
JWThewes: Addressed in 5900113.

Fixed: the classifier covers low credit balances, exceeded current quotas, generic usage limits and billing_hard_limit_reached. Fixtures include message excerpts from real Claude Code and OpenCode reports with source URLs. One important distinction from those reports: Gemini also says 'exceeded your current quota' during short-lived throttling, so an explicit retry-in-seconds/rate-limit hint stays transient rather than becoming billing guidance. Stage-level regressions verify the requested messages produce scoped recovery guidance. one-shot.js remains outside this change.

JWThewes:
The recovery goal is still the right one: the answer/bind race, missing Kiro conversation metadata, and distinction between storage failure and absent state are reproducible problems, and recovering the existing intent preserves work that a full restart discards. The previous patch was not ready to merge; the replay, batch revision and ownership blockers were valid.

The follow-up now uses one ownership rule for modern worker writes: META identifies the run, STAGE identifies the callback attempt, and both are checked with the mutation. Engine and agent gate creation are atomic with their owning wait; answer delivery uses the committed row; both unpark paths are idempotent after a lost checkpoint; retired workers cannot fail a replacement execution. Each of the ten new inline findings has its own published commit and reply. The two earlier fixes are retained.

I also merged current main and preserved its code-commit traceability and structured logging changes. Final validation on e36c3c8: all GitHub checks are green, including 3,064 backend tests on both Node 22 and Node 24, release-tooling tests, frontend build, lint/formatting and CodeQL. CI includes the Docker-backed DynamoDB Local/Neptune suites. Local validation also passed 58 IntentView tests, frontend typecheck, secret scanning, SDK alignment, real durable replay, real Git integration and bundle smoke tests. The dependency install audit reports zero vulnerabilities. The initial CI formatting failure was corrected in a documentation-only follow-up.

The remaining validation gap is a deployed, real-provider recovery of the reported production sequence. Existing legacy dispatches do not gain ownership tokens retroactively, so the orchestrator/runtime rollout needs to be coordinated; the contract is documented in docs/development/recovery-ownership.md. This PR does not reconcile a mismatched approved unit plan or refresh an intent's pinned configuration. I have left the review threads for you to confirm after re-review.

PR #467, opened 2026-09-15
fix: accept safe repository paths across Git providers

jeromevdl:
Space creation could reject repositories returned by the provider's own picker, including managed GitHub user names, GitLab subgroup paths, and paths exceeding GitHub-specific naming limits.

Validate safe repository paths across providers: accept two or more nonempty segments containing ASCII letters, digits, dots, underscores, and hyphens, without imposing provider-specific length or first-character rules. Reject non-string values, whitespace, shell metacharacters, and `.` or `..` path segments. Provider naming rules remain the provider's responsibility: the common validator accepts paths supported by any provider, without requiring each name to be valid on all three.

Apply the same validation when creating a space, adding a repository, or updating the primary repository. The create form's legacy `gitRepo` field and legacy repository migration also accept these safe paths so their older length guard does not reject a selected repository. Preserve support for previously accepted freeform legacy values.

Validation:

- Added 69 API regression tests covering all three write paths, persisted repository reads, provider-specific examples, long and nested paths, and unsafe or malformed inputs.
- All 190 project tests pass.
- Formatting and lint checks pass.

Review status: the requested rejection of `.` and `..` components is covered across all three API paths. The separate review item about manual native-export commands for leading-hyphen directory names remains open; this change does not update those commands.


JWThewes:
Reviewed commit `845a4141fe66b290b836c7814a5a02275df2f1f0`. The shared-validator change is small, follows the existing design, and preserves previously accepted repository names. I would address two cases before merging:

1. **[P2] Reject repository components equal to `.` or `..`.**  
   The new regex in `lambda/projects/index.js:293` accepts `owner/.` and `owner/..` through project creation, adding a repository, and updating the primary repository. For multi-repository checkout, `path.join(workspaceDir, 'owner/..')` resolves to the workspace root (`lambda/agentcore/workspace.js:207`), and failed-clone cleanup recursively removes that target (`workspace.js:188`). An isolated reproduction with the real checkout code, a temporary sibling checkout, and an injected credential failure deleted that sibling checkout. The normal launch flow has a separate repository-access check that limits reachability; this is not a demonstrated end-to-end production exploit.

   **Requested change:** Explicitly reject `.` and `..` as the repository component while retaining support for names such as `.github`. Add rejection tests for both values across all three API paths. Six additional checks returned 201/200 with this PR and correctly returned 400 with the original regex.

2. **[P2] Make native-export setup commands support leading-hyphen names.**  
   `owner/-repo` is now accepted, but `lambda/shared/native-repositories.js:52` assigns it the local directory `-repo`. The clone commands generated by `frontend/src/components/intent/NativeExportSetupDialog.tsx:191` and `:240` lack an option terminator. Git interprets `-repo` as options and exits with `unknown switch 'r'`; quoting the argument does not prevent option parsing. The following `cd '-repo'` command also needs a safe relative path.

   **Requested change:** Generate clone commands in the form `git clone --branch <branch> -- <url> './<directory>'`, preserving the existing shell quoting, and use `./<directory>` for relative directory changes in both manual setup variants. Add coverage for native export of `owner/-repo`.

Validation: all 144 project tests pass, and the changed files pass lint and formatting checks. The six additional dot-directory checks fail against this PR and pass with the original regex. The leading-hyphen Git argument-parsing failure was reproduced locally. No broader refactor is needed for either fix.

JWThewes:
Rechecked commit `166b3ec899ebf0aa74343b20364744881041a28b`. Both earlier findings are resolved: dot-directory names are rejected before filesystem changes, and native-export commands handle leading-hyphen directory names safely.

**[P2] Preserve the provider when a primary-repository update creates a repository record.**

The expanded validation in `lambda/projects/index.js:1650` now accepts GitLab subgroup paths and other provider-specific names, but `ensureLegacyRepoMigrated` still creates new records with `.property('provider', 'github')` at line 363.

Reproduction:

1. Create a space with `gitProvider: "gitlab"` and no linked repositories.
2. Send `PUT /projects/{id}` with `{"gitRepo":"team/subgroup/nested/repo"}`.
3. The update returns 200, but `GET /projects/{id}/repos` returns that repository with `provider: "github"`.

The same problem affects Bitbucket. It occurs when the target repository is not already linked; promoting an existing linked repository does not create a new record. Downstream source-control binding checks and intent checkout use the repository record's provider, so the successful update leaves the repository configured for the wrong service.

The hard-coded default predates this PR, but leaves the newly supported update paths incomplete.

**Requested change:**

- Pass the effective project provider into repository creation, accounting for `gitProvider` supplied in the same update as well as the stored project provider.
- Preserve the provider of an existing linked repository.
- Strengthen the new primary-repository update tests to assert `expect.objectContaining({ url, provider })` on the repository-list response. This assertion currently fails for all six GitLab/Bitbucket examples.

Validation: the unmodified PR passed 392 backend tests, 61 frontend tests, lint, and formatting. The strengthened provider assertion produced six failures and five passing GitHub cases. The shared validation and directory helper remain a simple, consistent design; this should only need a focused provider-propagation fix.

jeromevdl:
@JWThewes, Fixed (note that it was existing before this PR)

JWThewes:
Yeah it was - I checked that as well but nonetheless I think it's still good to fix it.

JWThewes (approved):

PR #468, opened 2026-09-15
feat: AgentCore Instances compute type (x86_64) for managed environments

matheus-1618:
## What

Adds opt-in support for the **AgentCore Instances compute type** to managed environments, including **x86_64** environments — the first non-arm64 path in the platform.

Admins creating a managed environment can now choose its compute:

- **Serverless microVMs (arm64)** — the existing default, unchanged
- **EC2 Instances (x86_64)** — runs the environment's runtime on EC2 managed instances in the deployment account, provisioned through an AgentCore capacity provider

Everything is gated behind a new terraform variable `enable_instances_compute` (default `false`). Deployments that don't opt in are unaffected — with the flag off there is no amd64 image build, no operator role, and the compute selector rejects `instances`.

## Why

- **x86_64 workloads**: microVMs are arm64-only; teams that need to build/run x86 artifacts had no path.
- **Persistent workspace**: the capacity provider's EBS volume replaces managed session storage at `/mnt/workspace`, lifting the fixed 1 GB session-storage ceiling and surviving session stops.
- **Account-level controls**: instances run in the deployment account (Savings Plans/ODCRs apply, data stays in-account), with sessions of up to 14 days.

## How

- `compute: { type, architecture }` field on the environment (API + UI selector on the create flow; immutable after creation, matching the AgentCore contract).
- **Capacity provider per architecture**, created lazily by the status lambda with a deterministic name on first use. Capacity providers are immutable (duplicate-to-change), so they are treated as create-once resources rather than terraform state.
- Runtime creation branches on the compute type: `capacityProviderConfiguration` + `capacityProviderVolume` (no `networkConfiguration` — Instances inherits the capacity provider's VPC).
- **One CodeBuild project serves both architectures** via per-build overrides (`environmentTypeOverride`/`imageOverride` + an `IMAGE_PLATFORM` variable in the buildspec).
- x86_64 environments build FROM an amd64 variant of the core image (opt-in terraform build). Catalog tools remain arm64-only for now and are rejected on x86_64 with a clear error.
- The core image Dockerfile is now architecture-aware: the base image is pinned to the **multi-arch index digest** (the previous pin referenced the arm64 manifest, which silently ignores `--platform`), and Kiro/OpenCode/bun/uv select artifact + pinned sha256 by `TARGETARCH`. arm64 builds resolve to the exact same artifacts as before.

## Fixes picked up along the way (apply to microVMs deployments too)

- `inspector2:ListCoverage`/`ListFindings` for the status lambda — in accounts with **ECR enhanced scanning**, `DescribeImageScanFindings` is served by Amazon Inspector and every image scan fails without these.
- `iam:CreateServiceLinkedRole` (scoped) — the first capacity provider in an account provisions the `AWSServiceRoleForBedrockAgentCoreRuntimeInstances` SLR.
- `bedrock-agentcore:PassCapacityProvider` — required by `CreateAgentRuntime` when associating a capacity provider.
- Transient first-invoke errors (instance provisioning) are retried instead of permanently failing the revision.
- Block-device workspace mounts come formatted — `mkfs.ext4` leaves `lost+found` at the volume root and `git clone` refuses the non-empty destination. The workspace init now clears the well-known filesystem entries before cloning (any other residue still fails loudly).

## Notes and limitations

- **Burstable instance types (t family) are rejected by CreateCapacityProvider** (`Instance type 't3.large' is not supported`); the default allowlist uses `m6i.large`. m5/m6i/c5/c6i were verified to be accepted.
- x86_64 environments must derive from the Standard environment (the base swap targets the amd64 core); tool-catalog x86 binaries are a natural follow-up.
- The environment verification guardrail now asserts the architecture from the recipe instead of hardcoding arm64.

## Validation

End-to-end on a fresh us-east-1 deployment with `enable_instances_compute = true`:

1. Environment created via the UI selector with `compute: { type: instances, architecture: x86_64 }`.
2. CodeBuild ran on the x86 fleet with `--platform linux/amd64`, image built FROM the amd64 core and passed the (now architecture-aware) verification script.
3. Capacity provider created lazily (`READY`), runtime created with `capacityProviderConfiguration` and the `workspace` EBS volume at `/mnt/workspace`.
4. Runtime validation invoked the runtime — an m6i instance was provisioned by AgentCore and the container reported `arch=x86_64`, non-root, writable workspace, all four agent CLIs installed.
5. Revision published; three further revisions repeated the pipeline cleanly with no intervention.
6. **A real intent ran end-to-end on the Instances runtime**: workspace cloned onto the EBS volume, and the agent CLI (Claude Code via Bedrock) executed stages on the EC2 managed instance.
7. `linux/amd64` and `linux/arm64` builds of the core image verified locally (all six CLI binaries at their pinned versions on both).

Tests: 122 backend (18 new covering the compute module and the Instances runtime path) and 569 frontend, all passing; `terraform validate` clean.


lambda/environments/compute.js
```diff
@@ -0,0 +1,199 @@
...
+          },
+          lifecycleConfiguration: {
+            maxLifetime: Number(process.env.MANAGED_INSTANCES_MAX_LIFETIME || 28800),
+          },
+          volumes: [
+            {
+              ebsConfiguration: {
+                name: WORKSPACE_VOLUME_NAME,
+                sizeGiB: Number(process.env.MANAGED_INSTANCES_WORKSPACE_GIB || 50),
+                volumeType: 'gp3',
```
JWThewes: **[P1] Add explicit cleanup for persistent Instances sessions**

These volumes persist after `StopRuntimeSession`, idle timeout, and maximum lifetime; they are deleted only by explicit session/provider deletion ([AWS lifecycle documentation](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/runtime-instances-data-management.html#runtime-instances-data-management-delete)). The existing validation cleanup in `status.js` only stops its randomly named session, and intent deletion in `lambda/shared/intent-deletion.js` also only stops sessions. Consequently, each validation session that provisions storage can leave a default 50-GiB volume behind, and deleting an intent leaves its workspace data and storage charges behind indefinitely.

Please retain the provider/session identifiers needed for cleanup, delete disposable validation sessions after a terminal result, and delete persistent sessions when their owning intent is permanently deleted. Add the necessary IAM permissions and lifecycle tests. Sessions temporarily stopped for park/resume should retain their volumes.

lambda/environments/status.js
```diff
@@ -538,7 +584,10 @@ const verifyRuntime = async ({
     });
     return { environment, revision: ready };
   } catch (error) {
-    if (RETRYABLE_CONTROL_ERRORS.has(error?.name)) {
+    if (
+      RETRYABLE_CONTROL_ERRORS.has(error?.name) ||
+      (environment.compute?.type === 'instances' && INSTANCES_TRANSIENT_ERRORS.has(error?.name))
+    ) {
```
JWThewes, line 666: **[P2] Retry provisioning with the same validation session**

Returning `pending` here does not retry the session that was provisioning: the next poll reaches `randomUUID()` at line 506 and invokes a new session, while the `finally` block stops the previous agent session. If the initial invocation consistently exceeds the provisioning window, every poll repeats a cold start and the revision can remain `VERIFYING` indefinitely. A local probe in which the first invocation of each session fails transiently and subsequent invocations can succeed produced three different sessions across three polls, with no revision progress.

Please persist or deterministically derive one validation session ID per revision, reuse it across transient retries, and defer terminal cleanup until validation succeeds or exhausts a bounded retry/deadline policy. Cover a cold start that requires more than one poll.

lambda/environments/compute.js
```diff
@@ -0,0 +1,199 @@
...
+    throw Object.assign(
+      new Error('x86_64 environments must derive from the Standard environment'),
+      { statusCode: 409, code: 'X86_64_BASE_MUST_BE_STANDARD' },
+    );
+  }
+  const amd64 = amd64CoreImage();
+  return {
+    ...recipe,
+    architecture: 'x86_64',
+    base: { ...recipe.base, imageUri: amd64.imageUri, imageDigest: amd64.imageDigest },
```
JWThewes: **[P2] Resolve the amd64 image from the selected published core revision**

`amd64CoreImage()` reads the latest deployment's environment variables, while this spread retains the selected Standard revision's `revisionId`. During a platform upgrade, `stageCoreRevision` deliberately leaves Standard's published pointer on the old revision until publication. Creating/editing/rebuilding an x86 environment during that interval therefore uses the new core image while claiming the old published base. A local probe kept the published base revision unchanged, changed only the configured amd64 digest, and obtained a different base image for that same revision.

Please store architecture-specific image references with the core revision and resolve the amd64 variant belonging to the selected published revision. Add an upgrade test where a newer core has been deployed but has not yet been published as Standard; x86 environments should continue using the published core's variant.

lambda/environments/compute.js
```diff
@@ -0,0 +1,199 @@
...
+  const name = capacityProviderName(architecture);
+  let token;
+  do {
+    const page = await controlClient.send(
+      new ListCapacityProvidersCommand({ maxResults: 20, nextToken: token }),
+    );
+    const match = (page.capacityProviders ?? []).find((item) => item.name === name);
+    if (match) {
+      if (match.status === 'READY') return { capacityProviderArn: match.capacityProviderArn };
+      if (match.status === 'CREATING') return { pending: true };
```
JWThewes, line 228: **[P2] Version capacity providers when their configuration changes**

The lookup uses only the project/environment/architecture name, so an existing `READY` provider is returned without checking its configuration. For example, changing the exposed Terraform allowlist from `m6i.large` to `m6i.xlarge` updates the Lambda environment but all subsequent runtimes still use the old provider. The same applies to replaced subnets/security groups and storage settings; a local probe changed the requested subnets and instance types and still received the original provider unchanged. A failed provider also permanently occupies the same lookup identity after configuration is corrected.

Please include a fingerprint of the immutable provider configuration in its identity, or implement an explicit replacement flow, so new revisions use the intended configuration while existing revisions retain their provider. Test both configuration changes and recovery after a failed configuration is corrected.

lambda/environments/index.js
```diff
@@ -355,20 +375,28 @@ export const createHandler = ({
...
           toolStore,
           data.recipe,
           baseEnvironmentId,
         );
+        const recipe = compute
+          ? applyComputeBase({ recipe: prepared.recipe, compute })
+          : prepared.recipe;
+        const flattenedRecipe = compute
+          ? applyComputeBase({ recipe: prepared.flattenedRecipe, compute })
+          : prepared.flattenedRecipe;
```
JWThewes, line 430: **[P2] Validate base architecture for default microVM environments too**

Published x86 environments are included in the base selector, but omitting `compute` normalizes to `null` and skips the compute/base checks here. A local handler probe created a default microVM environment derived from a published x86 parent and received HTTP 201 with an amd64 base digest and no recipe architecture. Its build is then scheduled as `linux/arm64`, so this accepted configuration cannot pass the image architecture checks. The edit path has the same validation gap.

Please validate the target architecture against the selected base revision for every creation/revision path, including the default compute type, and reject incompatible combinations before saving or queuing a build. Filter incompatible bases in the UI as well, and add a regression test for an arm64 child of an x86 parent.

JWThewes (changes requested):
This adds opt-in EC2-backed AgentCore environments with x86_64 images and persistent EBS workspaces, while retaining arm64 microVMs as the default. The focused compute module and reuse of the existing build pipeline are sensible.

Requesting changes for the five inline findings: persistent-session cleanup, cold-start retry identity, published-core image consistency, capacity-provider replacement after configuration changes, and architecture compatibility for inherited environments. These affect resource lifecycle, build reliability, and established revision semantics.

Validation on b5871c86678dec31db523ab6ec04903a36021fd5: all 122 environment tests and 11 workspace/runtime tests passed, and the environment Lambda bundles built successfully. Additional local probes reproduced the retry, image-selection, provider-configuration, and inheritance issues; persistent-volume retention was checked against the official AWS documentation. No AWS deployment or Docker image rebuild was performed during this review.

lambda/environments/status.js
matheus-1618: Addressed in de4beb2.

The validation session ID is now persisted on the revision and reused across polls, so a transient first-invoke error keeps the provisioning session alive and the next poll reattaches to it instead of repeating the cold start. The session is only stopped on a terminal outcome, and the retry budget is bounded (`MANAGED_INSTANCES_VALIDATION_MAX_POLLS`, default 30 one-minute polls — exhaustion fails the revision with a clear message). microVM validation keeps the original per-poll semantics. Covered by a multi-poll cold-start test that asserts the same session ID across polls plus the exhausted-budget path.

lambda/environments/compute.js
matheus-1618: Addressed in 5c06bb5.

- The capacity provider ARN is now persisted on the revision at runtime creation and carried into the intent's environment snapshot, so both cleanup paths have the identifiers they need.
- Disposable validation sessions are deleted (`DeleteCapacityProviderSession`) on every terminal outcome.
- Permanent intent deletion deletes the intent's main and lane sessions. An unexpected error aborts the cascade **before** the DynamoDB partition delete, so the intent still lists and the delete can simply be re-run (sessions that never existed are tolerated as misses). Park/resume is untouched — stopped sessions retain their volumes.
- IAM: `DeleteCapacityProviderSession` granted to the status, intents and projects lambdas; `@aws-sdk/*` bumped to ^3.1135.0 (the command only exists in the data-plane client from there) via the repo's `scripts/sync-aws-sdk.mjs`.
- Lifecycle tests added for both paths (`lambda/shared/test/intent-deletion-sessions.test.js` + `status-instances.test.js`).

lambda/environments/compute.js
matheus-1618: Addressed in 5724ba0.

The amd64 variant now travels with the core revision: `seedSystemEnvironments` and `stageCoreRevision` store `amd64Image` on the Standard revision, and when the deployed digest matches the already-published revision the variant is backfilled onto it (pre-existing deployments gain their variant exactly). `applyComputeBase` resolves the amd64 refs from the **selected base revision** only — the env-var reader is gone — and rejects with `AMD64_CORE_IMAGE_MISSING` when the revision has no variant. Includes the upgrade-window test: newer core deployed but not yet published as Standard → the x86 environment keeps using the published revision's variant.

lambda/environments/compute.js
matheus-1618: Addressed in 6ed7916.

The provider name now embeds a short hash of the create-time configuration (instance allowlist, VPC, storage, lifecycle, operator role), so a changed — or corrected-after-failure — configuration produces a fresh provider while runtimes created earlier keep the one they were built with. Superseded providers are intentionally left in place since they may still back existing runtimes (documented in the module header). Tests cover both a configuration change producing a new identity and recovery after a failed configuration is corrected.

lambda/environments/index.js
matheus-1618: Addressed in fbaa253.

`assertBaseArchitecture` now runs on every creation/revision path — create, edit, and rebuild-on-latest-base — including the default (compute omitted) case, rejecting an x86_64 base for an arm64 target with a 409 `BASE_ARCHITECTURE_MISMATCH` before anything is saved or queued. The UI base selector filters incompatible bases as well (x86_64 targets only offer Standard; arm64 targets exclude published x86_64 environments), and switching the compute selector to x86_64 resets the base to Standard. Regression test added for an arm64 child of a published x86_64 parent.

matheus-1618:
All five review findings are addressed, one commit per finding:

| Finding | Commit |
|---|---|
| [P1] Explicit cleanup for persistent Instances sessions | 5c06bb5 |
| [P2] Retry provisioning with the same validation session | de4beb2 |
| [P2] Resolve the amd64 image from the selected published core revision | 5724ba0 |
| [P2] Version capacity providers when their configuration changes | 6ed7916 |
| [P2] Validate base architecture for default microVM environments too | fbaa253 |

Each carries the regression tests requested inline (details in the per-thread replies). Note for reviewers: 5c06bb5 bumps `@aws-sdk/*` to ^3.1135.0 via `scripts/sync-aws-sdk.mjs` — `DeleteCapacityProviderSessionCommand` only exists in the data-plane client from that version.

Validation: 2871 backend + 569 frontend tests passing, `terraform validate` clean.

JWThewes (changes requested):
Re-reviewed `7b251f42086c10885ffd1d6f1f466a123a72f32d`. The retry-session identity, published-core image selection, capacity-provider configuration versioning, and base-architecture fixes look addressed. Requesting changes because the persistent-session cleanup fix still has two gaps that leave EBS workspaces and storage charges behind.

1. **[P1] Delete the actual persisted parallel-lane sessions.** In `lambda/shared/intent-deletion.js:265-272`, cleanup reads `stage.parallelSection`, but `buildStageRow` persists `sectionIndex`. With a real stage-row fixture, the cascade deletes only the main session and then deletes the intent partition; the lane session is never deleted. The new test misses this because it supplies `parallelSection` in its fixture.

   Please collect and deduplicate the persisted `records.units[].sessionId` values along with the main session, using `sectionIndex`/`unitSlug` as a fallback where necessary. Include historical/orphaned lanes rather than relying only on the current unit plan. Keep the intent records until session deletion succeeds or the session is confirmed absent. Add regression tests using `buildStageRow` and actual unit-row shapes, including a lane no longer in the current plan and a deletion failure that leaves the cascade retryable.

2. **[P2] Retain and retry failed validation-session cleanup.** In `lambda/environments/status.js:614-619`, `DeleteCapacityProviderSession` failures are logged and swallowed. The subsequent READY/FAILED transition clears `validationSessionId`, so the poller has no durable cleanup work to retry. A local probe returning `InternalServerException` from deletion after successful validation produced `READY`, `validationSessionId: null`, and no pending retry. The comment's eventual capacity-provider deletion is not implemented by this PR and is not a cleanup guarantee for a shared provider.

   Please persist the provider/session identity and pending cleanup state until deletion succeeds or the session is confirmed absent, and have the poller or a dedicated cleanup worker retry it even after validation reaches a terminal result. Test deletion failure after both successful and failed validation, then a later successful retry that uses the same session ID and clears the cleanup record only after success.

Validation: all 159 focused environment/session/workspace tests passed, the environment Lambda bundles built, and all PR CI checks are green. Both gaps above were reproduced separately against this head. No live AWS deployment was performed during this review.

JWThewes:
I just deployed that: if enable_instances_compute is false (default) we should hide that in the Environments settings. Now I can select EC2 but on creating the draft I get an error that enable_instances_compute is not true...

JWThewes:
We're also missing the whole documentation piece around this...

JWThewes:
Also I can't add any tool to the ec2 compute type as they're arm64 only. I would vote for adding support for this as otherwise we're missing out the main part of environments and tools...

JWThewes:
Screenshot 2026-09-22 at 14 32 25

I can't get an EC2 environment to work at all. I accepeted the security findings but afterwards nothing happens at all...

jeromevdl:
Wondering if we should not wait for https://aws.amazon.com/blogs/machine-learning/the-new-agentcore-runtime-elastic-optimized-and-consistently-fast-starts/ 
> x86 support. Run the agent, tool, or environment you already have with x86 microVMs. Teams whose code or dependencies target x86 can move an agent, a tool, or an execution environment to AgentCore as-is.

It would make it native, no custom code, reduced footprint on our current code. @JWThewes ?

matheus-1618:
Thanks for the re-review and the deployment feedback — all addressed as of eb72302.

**[P1] Parallel-lane session deletion — fixed in c5a1b1c.** You were right: the cleanup read `stage.parallelSection` while `buildStageRow` persists `sectionIndex`, so the lane set was always empty. The cascade now rebuilds the session set from the persisted records via `collectIntentSessionIds`: the deduplicated `records.units[].sessionId` values (UNIT# rows survive plan rewinds, so historical/orphaned lanes are covered without depending on the current unit plan), a `sectionIndex`/`unitSlug` fallback for lanes that never stamped a session, the lane STAGE# rows as a second net, and the main session. `deleteRuntimeSessions` still throws before the intent records are deleted, so a failed delete leaves the whole cascade retryable. The regression tests now build every fixture with `buildStageRow`/`buildUnitRow` (a store field rename now breaks the test instead of silently missing sessions) and cover a lane that is no longer in the current plan plus a deletion failure that preserves the records.

**[P2] Validation-session cleanup retention — fixed in 1dfce9a.** A failed `DeleteCapacityProviderSession` now persists a `SESSION_CLEANUP` record (provider ARN + session id + attempt count) the moment the delete fails with anything other than a confirmed-absent error, before the READY/FAILED transition clears `validationSessionId`. The status poller retries the delete on every tick — including after the revision reached a terminal state — and clears the record only after a successful delete or a confirmed-absent response. Tests cover deletion failure after both successful (READY) and failed validation, and a later retry that reuses the same session id and clears the record only on success. Your `InternalServerException` probe now produces READY + a pending cleanup record instead of a leak.

**EC2 selectable with `enable_instances_compute = false` — fixed in 9737fe7.** Added `GET /environments/capabilities` (reports `instancesCompute`/`amd64CoreImage` from the deployment configuration); the environment builder only renders the compute selector when the Instances type is actually configured, so the 409 at draft creation can no longer be reached from the UI.

**Documentation — added in eb72302.** New "The Instances (EC2) compute type" section in `docs/using-the-platform/managed-environments.md`: enablement, the Standard-base and no-catalog-tools restrictions, cold-start validation semantics, and the persistent EBS workspace lifecycle (when sessions are deleted and how failed deletions are retried).

**"Accepted the security findings and nothing happens" — hardened in 1dfce9a.** The strongest silent-stall vector I found: the status poll ran all passes in one unguarded sequence, so any single revision whose handling threw unexpectedly aborted the entire poll — including the acknowledged→runtime-creation pass — on every tick. That presents exactly as "accepted the findings, nothing happens", with the real error only visible as a generic `status handling failed` log. Each poll item is now isolated and failures are logged with the pass name and environment/revision ids. If your environment still stalls after deploying this head, the `environment-status` CloudWatch logs will now say exactly which item is failing and why — please share the `Managed environment poll item failed` entries and I'll chase it. Also note the EC2 path is legitimately slower than microVMs: capacity-provider creation plus the first-instance cold start can keep a revision in VERIFYING for several minutes.

**arm64-only tools on x86_64.** Agreed this is the main gap for Instances environments. It's a bigger change than it looks: a tool version's identity is currently bound to a single-architecture artifact (aarch64 download URL + publisher checksum) and the build/verification pipeline runs `docker build --platform linux/arm64` on the arm64 core base — x86_64 support means per-architecture sources, checksums, builds, OCI digests, and verification runs for every tool version, plus catalog/recipe changes to resolve the right variant. I'd rather not fold that into this PR (it's already large); I propose a follow-up issue for multi-architecture tool builds, with the interim documented workaround being apt packages / build commands under Advanced settings for x86_64 environments. Happy to open the issue with a design sketch if that works for you.

Validation on this head: 852 lambda tests across the environments and shared projects (including the new P1/P2/poll-isolation regressions), frontend typecheck + tests, and the four environment Lambda bundles built cleanly.

JWThewes:
> Wondering if we should not wait for https://aws.amazon.com/blogs/machine-learning/the-new-agentcore-runtime-elastic-optimized-and-consistently-fast-starts/
> 
> > x86 support. Run the agent, tool, or environment you already have with x86 microVMs. Teams whose code or dependencies target x86 can move an agent, a tool, or an execution environment to AgentCore as-is.
> 
> It would make it native, no custom code, reduced footprint on our current code. @JWThewes ?

Yeah but having the base to actually build x86 tools and environments still makes sense, right?
We could then let users decide between arm64 and x86 on AgentCore Runtime vs AgentCore Instances?
WDYT @jeromevdl

PR #469, opened 2026-09-15
Update deployment workflow

eipasteur:
Adjust deployment configuration handling.


.github/workflows/deploy-demo.yml
```diff
@@ -112,6 +121,28 @@ jobs:
...
+            node --input-type=module <<'NODE'
+            const aliases = JSON.parse(process.env.APP_DOMAIN_ALIASES);
+            if (!Array.isArray(aliases) || aliases.some(alias => typeof alias !== 'string')) {
+              throw new Error('APP_DOMAIN_ALIASES must be a JSON array of hostnames.');
+            }
+            for (const alias of aliases) {
+              console.log(`::add-mask::${alias}`);
+              console.log(`::add-mask::https://${alias}`);
+            }
+            NODE
```
JWThewes: [P1] Fix the indentation of both heredoc terminators

After YAML removes the block indentation, this closing `NODE` still has two leading spaces. Bash does not recognize it as the terminator for `<<'NODE'`, so the first deployment step fails with `syntax error: unexpected end of file`, even when `APP_DOMAIN_ALIASES` is empty. The second new heredoc in `Validate deployment configuration` has the same problem at line 277. I reproduced both errors by parsing the workflow YAML and running `bash -n` on the extracted scripts; the base workflow passes. Align both closing markers with the script block's baseline indentation (the surrounding `if`), so they appear at column 1 in the generated shell script.

.github/workflows/deploy-demo.yml
```diff
@@ -218,6 +249,33 @@ jobs:
...
+          if [[ -n "$APP_DOMAIN" ]]; then
+            node --input-type=module <<'NODE'
+            import { lookup } from 'node:dns/promises';
+
+            const domains = [
+              process.env.APP_DOMAIN,
+              ...JSON.parse(process.env.APP_DOMAIN_ALIASES),
+            ];
+            for (const domain of domains) {
+              await lookup(domain);
```
JWThewes: Scope clarification (non-blocking): does this workflow support initial provisioning with a custom domain, or only updates to an existing demo?

For a brand-new distribution using external DNS, the generated CloudFront hostname is not available until Terraform creates the distribution. Requiring the application hostname to resolve before Terraform runs blocks that setup sequence; after correcting the heredoc, I confirmed an unresolved name stops this step with `ENOTFOUND`. For an existing distribution, its target is already known and DNS can be configured beforehand, so this can be a reasonable prerequisite. If initial provisioning is in scope, defer DNS readiness checking until the distribution exists and external DNS can be configured. Otherwise, documenting preconfigured DNS as a prerequisite would make the intended scope clear.

.github/workflows/deploy-demo.yml
eipasteur: I replaced both heredocs with multiline node -e invocations and validated every YAML-extracted run block with bash -n

.github/workflows/deploy-demo.yml
eipasteur: This targets an existing deployment. The preflight check is intentional and the initial provisioning is out of scope here I would say.

JWThewes (approved):

PR #472, opened 2026-09-17
chore(deps): add dependency-cruiser for backend module analysis

jeromevdl:
## Why

This application is growing quickly, and much of the code is AI-generated. That combination makes it easy for structure to erode silently: circular dependencies, cross-module coupling, and "transversal" concerns that get threaded by hand through every layer instead of living behind one abstraction. None of that shows up in unit tests or linting — it only surfaces later as churn, where a small feature touches dozens of files.

This PR adds an automated guardrail for exactly that: it makes the backend's dependency structure visible and enforces a few architectural invariants, so we can keep coupling low and catch regressions at PR time rather than at review time (or never). It is the JS equivalent of a JArchitect-style coupling/cycle analysis.

## What

[dependency-cruiser](https://github.com/sverweij/dependency-cruiser) over `lambda/`, with rules in `.dependency-cruiser.cjs` enforced by the pre-commit hook and CI:

- **no-circular** — forbid circular dependencies.
- **shared-is-foundation** — `lambda/shared/` must not import a Lambda workspace; it stays a leaf.
- **no-cross-lambda** — a workspace must not import another workspace's internals; shared code goes through `lambda/shared/`.
- **not-to-test** — production code must not import test files.
- **high-fan-in** (warning) — flags any module imported by more than 10 other production modules. High afferent coupling means a wide blast radius on change, and often points to a transversal concern that would be better behind one owning abstraction. Advisory only: it does not fail the pre-commit hook or CI, it is a prompt to investigate with `npm run dep:reaches`. The threshold is tunable in `.dependency-cruiser.cjs`.
- plus `no-orphans`, `not-to-dev-dep`, `no-deprecated-core`, `no-duplicate-dep-types`.

Tests are kept in the graph so `not-to-test` is enforceable; the report/metrics/graph scripts pass `--exclude` on the CLI to keep those outputs source-only.

## Scripts (documented in the README under "Testing and code quality")

| Script | Purpose |
|---|---|
| `npm run dep:check` | Enforce the rules (fails on any violation). Runs in pre-commit + CI. |
| `npm run dep:report` | HTML dependency matrix. |
| `npm run dep:metrics` | Coupling / instability table (afferent/efferent coupling, instability). |
| `npm run dep:graph` | Mermaid module graph. |
| `npm run dep:all` | report + metrics + graph. |
| `npm run dep:focus -- "<pattern>"` | Focused HTML matrix for one subsystem. |
| `npm run dep:reaches -- "<file>"` | Blast radius: everything that transitively depends on a module. |

Reports are written to `reports/dependency/` (git-ignored).

## Current state

The backend passes cleanly today: **no circular dependencies**, and no cross-workspace or prod→test imports. So this ships as a green baseline — the intent is to keep it that way as the codebase grows.

## Reading the matrix

In the HTML matrix a filled cell means the **row** module imports the **column** module. A dense column is a widely-depended-on module (high fan-in); a dense row depends on many others (high fan-out). A concern with no owning abstraction shows up as a column whose filled cells are spread across many unrelated workspace blocks — a useful lens for spotting refactor candidates.


jeromevdl:
`npm run dep:check` currently gives :

```
  warn high-fan-in: lambda/shared/response.js
  warn high-fan-in: lambda/shared/legacy-authz.js
  warn high-fan-in: lambda/shared/agent-credentials.js
  warn high-fan-in: lambda/agentcore/mcp/graph-writer.js

x 4 dependency violations (0 errors, 4 warnings). 365 modules, 729 dependencies cruised.
```

Confirming the highspot on agent-credentials (https://github.com/aws-samples/sample-collaborative-ai-dlc/pull/452#issuecomment-5698564355)

.dependency-cruiser.cjs
```diff
@@ -0,0 +1,207 @@
...
+    doNotFollow: {
+      path: ['node_modules'],
+    },
+    // NB: test files are intentionally NOT excluded here so the `not-to-test`
+    // rule can see (and forbid) any prod -> test edge. The report/metrics/graph
+    // npm scripts pass `--exclude` on the CLI to keep those outputs source-only.
+    exclude: {
+      path: ['node_modules', '\\.build/', '(^|/)coverage/', '(^|/)\\.stryker-tmp/'],
+    },
+    includeOnly: {
```
JWThewes: [P2] Preserve imported targets for rule validation

`includeOnly: '^lambda'` removes dependency targets outside `lambda/`, and the `node_modules` exclusion above removes package dependencies before validation. In a temporary checkout, adding both `import 'vitest';` and `import '../../test/setup.js';` to `lambda/agents/index.js` still let `npm run dep:check` exit 0 with exactly the same graph and warnings as the baseline. This makes `not-to-dev-dep` ineffective and allows production imports of the repository's root test helpers to bypass `not-to-test`. Please retain imported targets in the validation graph and use `doNotFollow` to limit traversal, with rule scopes/exceptions that reflect the existing workspace dependency layout.

package.json
```diff
@@ -51,6 +51,13 @@
     "audit:prod": "npm audit --omit=dev --audit-level=high",
     "audit:prod:frontend": "npm --prefix frontend audit --omit=dev --audit-level=high",
     "audit:prod:all": "npm run audit:prod && npm run audit:prod:frontend",
+    "dep:check": "depcruise lambda",
+    "dep:report": "depcruise lambda --exclude \"node_modules|\\.build/|(^|/)coverage/|(^|/)\\.stryker-tmp/|(^|/)test/|\\.(test|spec)\\.(js|mjs|cjs|ts|tsx)$\" --output-type html --output-to reports/dependency/report.html",
```
JWThewes: [P2] Create the report output directory before writing

On a fresh checkout, `reports/dependency/` does not exist and dependency-cruiser does not create it. I reproduced an `ENOENT` failure from `npm run dep:all` at its first report step, and the same failure from `npm run dep:focus -- 'shared/agent-credential'`. After manually creating the directory, all reports generated successfully. Please make each reporting entry point create the output directory automatically, directly or through a shared helper, so the documented commands work after installation.

JWThewes (changes requested):
The overall approach fits the repository's existing npm/Husky/CI workflow, and I found no direct application regression or changes to existing production dependencies. Requesting changes for the two reproduced issues noted inline: graph filters hide forbidden dependency targets, and report commands fail on a fresh checkout.

Validation: installed the locked dependencies; ran the new commands, lint, formatting checks, SDK alignment, and targeted violation probes. Cycles, cross-workspace imports, shared-to-workspace imports, and imports of tests inside lambda/ were correctly detected. GitHub checks were green; the full application test suite was not rerun locally.

jeromevdl:
Thanks @JWThewes, all fixed in 1c55c81

.dependency-cruiser.cjs
```diff
@@ -0,0 +1,217 @@
...
+// The root manifest contains both tooling and dependencies shared by Lambda
+// workspaces. Treat only the tooling subset as forbidden in runtime modules.
+const rootDevDependencies = Object.keys(require('./package.json').devDependencies);
+const runtimeRootDependency =
+  /^(?:@aws-sdk\/|@smithy\/|gremlin$|gremlin-aws-sigv4$|neptune-lambda-client$)/;
+const escapeRegExp = (value) => value.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+const devOnlyDependencyPath = `^node_modules/(?:${rootDevDependencies
+  .filter((dependency) => !runtimeRootDependency.test(dependency))
+  .map(escapeRegExp)
+  .join('|')})(?:/|$)`;
```
JWThewes: [P2] Include workspace devDependencies in runtime import validation

`devOnlyDependencyPath` is built exclusively from the root manifest's `devDependencies`, so tooling declared only in a Lambda workspace is omitted from `not-to-dev-dep`. The repository already has an affected dependency: `lambda/v2-orchestrator/package.json` declares `@aws/durable-execution-sdk-js-testing` as a devDependency.

Reproduced on `1c55c81` with the locked dependencies installed:

1. Add `import '@aws/durable-execution-sdk-js-testing';` to `lambda/v2-orchestrator/index.js`.
2. Run `npm run dep:check`.

The command exits **0** with only the four baseline warnings. The JSON graph resolves the import to `node_modules/@aws/durable-execution-sdk-js-testing/dist-cjs/index.js`, classifies it as `npm-dev`, and marks it `valid: true`. This leaves production imports of workspace-only test tooling undetected by both the hook and CI.

Please cover workspace devDependencies as well as root tooling, for example by using the `npm-dev` classification with explicit exceptions for the intended shared runtime dependencies, or by deriving the rule from the relevant workspace manifests. Verify that this production import fails with `not-to-dev-dep`, while existing test imports of the SDK and legitimate shared runtime dependencies remain allowed. The original `vitest` and root test-helper probes should continue to fail.

JWThewes (changes requested):
The report-directory issue is fixed, and the original `vitest` and root test-helper probes now fail correctly. Requesting changes for the remaining workspace devDependency gap described inline.

Validation on `1c55c81`: all reporting entry points succeeded with the output directory initially absent; baseline dependency checks, lint, formatting, and SDK alignment passed. Targeted probes confirmed cross-workspace, shared-to-workspace, and circular imports are rejected. All seven GitHub checks are green. The full application test suite was not rerun locally.

JWThewes (approved):

